PR #99, opened 2018-04-25
Service testing

### LINKED ISSUE #28: Service testing/development
Have a way through the CLI to execute a local service in development

antho1404:
Add the test command for the service.

This command now run a server, start the service in parameter then listen for events (filter them if necessary) and also send task with data

closes #28

### DIFF
--- a/api/event/emit.go
+++ b/api/event/emit.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"log"
 
+	"github.com/logrusorgru/aurora"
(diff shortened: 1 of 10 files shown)


antho1404:
@NicolasMahe why add a random port ? how the service can communicate if it stays alive and we change the port for every tests...
If it's just to avoid the conflict with the daemon I don't agree, maybe we should provide a `--port` flag to fix this but definitely not a random port.

Can you explain why you added this please

NicolasMahe:
good point. commit reverted.

PR #100, opened 2018-04-25
Result of tasks

### LINKED ISSUE #93: Result of task
Implement the api to send the result of a task and listen the result of tasks.

In order to do that we need to track the task with an ID that will be send back by the service

antho1404:
Add the apis for task results

- Result.Submit
- Result.Listen

fixes #93


NicolasMahe:
Instead of `Result.Listen`, can we send the result back to `Task.Execute`? It will be way easier for application to get the result of a task.

antho1404:
Agreed for the id to track which execution I'll add it today, for the result directly from the `Task.Execute` I'm afraid that this will be too complicated for users with an input stream and an output stream. Also how do we call the `Result.Listen` function in this case, we cannot say `Task.Listen` because this is when we have to execute tasks, so if we keep the `Result.Listen` it kind of makes sense to have the `Result.Submit` at the same place 
What do you think ?

NicolasMahe:
In the system you did, I prefer `Result.Listen`.

---

My suggestion doesn't change the `Task.Execute` api. It's still a simple input, and a simple output. But instead of publishing the task job on the channel and return right after, the core could wait for the result of the task job (with a timeout), and send the result directly to the caller. It is still a 1 to 1 call (1 job -> 1 result). No stream needed.
Because anyway, what application will listen for every `Result.Listen`? It only want to listen to the result of the job it ask for.

Eg:
Application: execute query `Task.Execute` -> Core: publish in `Task.Listen` -> Service: Receive job. Execute job. Post result in `Result.Submit` -> Core: publish result in response of the first query `Task.Execute`.

What do you think?

antho1404:
`Task.Execute` cannot wait for the answer. What happen if it takes the service 10min to process the task, we will have timeouts, we might just open a stream and then when the result is coming from the service publish on the stream and close the stream but then what's the difference with a stream that stays open to listen any kind of results

antho1404:
@NicolasMahe i would love some feedbacks about this one, I'm writing the documentation and I would like to finish this one to have something full functional

NicolasMahe:
I think it is more complicated to have to listen for the result of an task than wait for the response on the same query.
A lot of task with useful result will be fast to execute (eg: get data, generate a new ethereum wallet), but yes, some might take a long time.
The concept of timeout is a really vague on gRPC: https://grpc.io/docs/guides/concepts.html#deadlinestimeouts
It seems it's the client that decide.

After reflexion, we could use a stream with `Result.Listen` (as you did) and in our app lib, automaticaly listen for the result and return it in the same "promise" as the call. So we have the most generic way in the lowest level, and if we want, do a simpler version in our app lib.
What do you think?

antho1404:
100% agree, for me this is the low level API but sure we can use a lib to reduce the complexity but on the "client side" but the API will still be the same just for exemple in js something like that
```node
const taskX = (inputs, onResult) => doSomeProcessing().then(onResult)
```
where onResult is a function that will be created already with the id of the task and everything and where you just need to call it with the results and that's it but this is just some help on the client side

antho1404:
What would be the best solution to generate a unique id to identify the task execution ?
At first I thought a random like an UUID but I really don't want to rely on a random, I would like to have a deterministic way to generate this `executionID`, I think this can be really helpful when we will have the network like that we can detect two similar executions.

Any thoughts about that ?

I was thinking a hash of 
- request execution time
- task name
- task data
- signature of the core ? this one maybe too much for now

antho1404:
I added a new package to manage the state of the executions basically this package have 3 list :

- pending executions
- in progress executions
- processed executions

With `Create`, `Execute` and `Complete` for every steps with all the time the right notifications to the right pub/sub.

@NicolasMahe ready for review

NicolasMahe:
Nice job on the execution package 👍 

I like your idea of a deterministic id for the execution. I'm ok with your list but i will also add the service name.

Also, the gRPC messages contain the parameters `task` and `data` with `executionID`. It's useless in this case, no?
I suggest for `api_result.proto`:
```proto
syntax = "proto3";

package types;

import "service.proto";

service Result {
  // Service
  rpc Submit (SubmitResultRequest) returns (SubmitResultReply) {}

  // Application
  rpc Listen (ListenResultRequest) returns (stream ListenResultReply) {}
}

message SubmitResultRequest {
  string executionID = 1;
  string output = 2;
}

message SubmitResultReply {
  string error = 1;
}

message ListenResultRequest {
  ProtoService service = 1;
}

message ListenResultReply {
  string executionID = 1;
  string output = 2;
}
```
and for `api_task.proto`:
```proto
syntax = "proto3";

package types;

import "service.proto";

service Task {
  // Service
  rpc Listen (ListenTaskRequest) returns (stream ListenTaskReply) {}
  
  // Application
  rpc Execute (ExecuteTaskRequest) returns (ExecuteTaskReply) {}
}

message ListenTaskRequest {
  ProtoService service = 1;
}

message ListenTaskReply {
  string executionID = 1;
  string task = 2;
  string data = 3;
}

message ExecuteTaskRequest {
  ProtoService service = 1;
  string task = 2;
  string data = 3;
}

message ExecuteTaskReply {
  string executionID = 1;
  string error = 2;
}
```

Otherwise, I really doesn't like the organization of the API we are doing. Each gRPC service has 1 endpoint for the application and another for the service. It's super confuse and we have to create so many different client to access only 1 endpoint per client:
```go
cli := types.NewTaskClient(connection)
reply, err := cli.Execute(context.Background(), &args)
...
resultCli := types.NewResultClient(connection)
stream, err := resultCli.Listen(context.Background(), &args)
```
I was thinking why we just do one big service that group the task and result services?
Something like:
```proto
service Task {
  // Service
  rpc Listen (ListenTaskRequest) returns (stream ListenTaskReply) {}
  rpc Submit (SubmitTaskRequest) returns (SubmitTaskReply) {}
  
  // Application
  rpc Execute (ExecuteTaskRequest) returns (ExecuteTaskReply) {}
  rpc Result (ResultTaskRequest) returns (stream ResultTaskReply) {}
}
```
We could also organize into 1 gRPC service for the Services, and 1 gRPC service for the Application:
```proto
service Task {
  // Service
  rpc Listen (ListenTaskRequest) returns (stream ListenTaskReply) {}
  rpc Submit (SubmitTaskRequest) returns (SubmitTaskReply) {}
}
service Execution {
  // Application
  rpc Execute (ExecuteExecutionRequest) returns (ExecuteExecutionReply) {}
  rpc Listen (ListenExecutionRequest) returns (stream ListenExecutionReply) {}
}
```

antho1404:
I extracted your comment in the issue https://github.com/mesg-foundation/core/issues/112 like that we can finalize this PR.

I already added the name of the service and I just added the inputs in json too so now this feature should be finish

PR #110, opened 2018-05-01
add MESG endpoint to the service container

### LINKED ISSUE #105: Add core endpoint as an env variable in the services
In order to contact the core the service needs to know which endpoint to check so we should provide this endpoint directly to the service through an environmental variable `MESG_ENDPOINT`

antho1404:
close #105

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,15 +7,7 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker
-      - restore_cache:
-          key: v1-mesg-core-test
-      - run: go get -v -u -t ./...
-      - save_cache:
-          key: v1-mesg-core-test
-          paths:
-            - /go/bin
-            - /go/pkg
-            - /go/src
+      - run: go get -v -t ./...
       - run: go test -coverprofile=coverage.txt -v ./...
       - run: bash <(curl -s https://codecov.io/bash)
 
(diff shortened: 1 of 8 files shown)


antho1404:
this one is quite easy but unfortunately the communication between the core and the service doesn't work.

The container cannot have access to the host.

from what I looked it's not supported by swarm 
https://github.com/moby/moby/issues/25873
https://github.com/docker/swarmkit/issues/989

I see two solutions for this :
- We remove swarm and the docker services and we just create normal containers that all have access to the same network 👎 
- We might use a file with an IPC / socket connection like that it's a file to share between the host and the docker service and this is definitely possible we just have to see if we can use grpc with an unix socket or something like that

Any ideas ?

antho1404:
for information, if I use a system like ngrok to bind your localhost to outside, my service can reach the api and submit events with no problems, it's really just the localhost

NicolasMahe:
That's not cool...
I'm pretty sure gRPC works fine over socket. It uses default net lib that are also compatible with socket.

antho1404:
Created a socket that is stored in the `~/.mesg/server.sock` file, everything is fine if we use it directly but not inside docker, the socket is accessible in the container but we cannot dial anything

Maybe a permission problem, a folder sharing problem or something like that

Any idea ?

antho1404:
problem with the rebase so new pull request https://github.com/mesg-foundation/core/pull/117

PR #114, opened 2018-05-02
Simplify api

antho1404:


antho1404:
I added the new API with the client and service api 
https://github.com/mesg-foundation/core/blob/2e73981067b6d45b7d66e427fd711fb6211b5281/api/client/api.proto

https://github.com/mesg-foundation/core/blob/2e73981067b6d45b7d66e427fd711fb6211b5281/api/service/api.proto

Also added an event package to handle all events

and added a `build/proto.sh` script to build all proto files

antho1404:
don't mind about the code coverage, this is because of the generated files from protobuffer, we need to remove them from the codecoverage but for now it's fine

antho1404:
bug on test tracked here https://github.com/mesg-foundation/core/issues/116

Issue #116, opened 2018-05-03
Fix TestStartService test


antho1404:
When testing the api the CI fails to execute this test because the swarm on docker is not ready yet. We have to find a way to fix this test and make sure that everytime we ask for swarm to be created to wait that it's really created to continue.

Maybe do a polling on docker api until the node is swarm

PR #117, opened 2018-05-03
add MESG endpoint to the service container

### LINKED ISSUE #105: Add core endpoint as an env variable in the services

antho1404:
close #105


antho1404:
Created a socket that is stored in the ~/.mesg/server.sock file, everything is fine if we use it directly but not inside docker, the socket is accessible in the container but we cannot dial anything

Maybe a permission problem, a folder sharing problem or something like that

Any idea ?

antho1404:
found this https://github.com/docker/for-mac/issues/483
it can be related

NicolasMahe:
Did you see this article: https://www.jujens.eu/posts/en/2017/Feb/15/docker-unix-socket/
It says that you need to give permission to the user of the container.

Also with the socket, there will be only one socket shared for all the services? It may cause problem in the futur. I don't really like to use socket. It may also cause problem on Windows system.

antho1404:
Ready for review as quick as possible, the test fails because of #116 

How I made it works...

First the socket doesn't works between unix container and anything else (at least mac) so cannot use the socket to communicate from the core to the service directly.

## Solution

I dockerize the core and then run it in docker with the `mesg-daemon` script. Like that the server starts in an unix container and so the socket can be shared with any other containers and the communication can be done.

The socket is created in `/mesg/server.sock` in the core container then shared in the `~/.mesg` in the host and the shared from the `~/.mesg` to the `/mesg/server.sock` for the service. So I give access to the socket to the containers through the host.

Then the host still cannot read/write on the socket osx/unix so I created a second server TCP to handle the connections from the CLI.

So now the CLI connects with the TCP to the daemon that runs in docker and then the services communicates with the daemon containers through a shared socket

daemon_service_communication

NicolasMahe:
I'm ok for this solution. We need something that works.. We can find a nicer solution in the future.
But, why use a socket instead of a TCP for the core to services communication? Docker allows container links.

PR #118, opened 2018-05-05
Volumes for dependencies

antho1404:
It is now possible to add volumes from dependencies. This is useful when you need to preserve a database or anything that need to persist on disk (like for example the blockchain and configurations with ethereum)

```mesg.yml
name: "Test"
...
dependencies:
  ethereum:
    image: ethereum/client-go
    volumes:
      - /root/.ethereum
```

In this case the shared folder will be stored in `~/.mesg/services/MESG-Test/ethereum/root/.ethereum`

This pull request requires is on top of #117


antho1404:
Also now we can use the `volumefrom` to have a volume from another dependency

example:

```yml
dependencies:
  app:
    image: mesg/test
    volumesfrom:
      - ethereum
  ethereum:
    image: ethereum/client-go
    volumes:
      - /root/.ethereum
```

app will have the folder `/root/.ethereum` from the ethereum container

NicolasMahe:
What is the advantage of `volumesfrom`? I don't see the point.

antho1404:
in this case for example with ethereum, geth creates a unix socket in the `/root/.ethereum` folder so to communicate with the geth api I use the socket that is from the shared volume of ethereum. I could use the ws connection that geth provides but it means extra configurations and why using the web when we have a file

Issue #120, opened 2018-05-05
Update CD for core dockerization


NicolasMahe:
- [x] The CD should build the docker image and publish on docker hub.
- [x] rename the tag for bin release
- [x] improve bin name

Link to #117

PR #131, opened 2018-05-08
Manage the daemon from the CLI

### LINKED ISSUE #121: Merge mesg-daemon script file to CLI
The mesg-daemon script should be integrated as the "default" command of the CLI. 
The go core docker startup could be move to a specific package.

We can also provide the `detach` option of Docker run.

Link to #117 

@antho1404 what do you think?

antho1404:
We have now the following commands

- mesg-core daemon start
- mesg-core daemon stop
- mesg-core daemon status

fix #121


NicolasMahe:
When executing the start command, it runs indefinitely because I don't have the `mesg-daemon` image.
I can see where the error is coming from:
image

Can we check for tasks success?

antho1404:
yes the error is normal for now we have to build the docker image but this will be fixed by https://github.com/mesg-foundation/core/issues/120

PR #137, opened 2018-05-10
Create install curl script

### LINKED ISSUE #107: Create install curl script

NicolasMahe:
Close #107

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+# Copyright 2018 mesg.tech
+
+ARCH=$(uname -m)
+if [[ "$OSTYPE" == "linux-gnu" && "$ARCH" == "x86_64" ]]; then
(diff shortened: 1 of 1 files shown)


NicolasMahe:
@antho1404 how do we publish the file on https://get.mesg.tech ?

antho1404:
just after installing it we cannot run the `mesg-core` command, If we open a new terminal it's fine I think we need to find a way to add the `source` in the end with the right file to source and reset the terminal

antho1404:
I think we can just make a redirection from the subdomain `get` directly to the location of the file on github
Something like `get CNAME github.com/mesg-foundation/core/.../install.sh`

NicolasMahe:
Put the install command in the doc and remove the `install.sh` file

NicolasMahe:
I update the documentation with the install for mac.

PR #139, opened 2018-05-10
[WIP] Improve `mesg-core daemon start`

### LINKED ISSUE #138: Improve `mesg-core daemon start`
It will be super nice to display a loader during the `mesg/daemon` image download.
As the image is big, the user can feel confuse by waiting without progress bar.

Maybe we can download the image before starting the service in order to have a progress.

@antho1404 what do you think?

NicolasMahe:
close #138

### DIFF
--- a/cmd/daemon/start.go
+++ b/cmd/daemon/start.go
@@ -1,7 +1,9 @@
 package daemon
 
 import (
+	"bytes"
 	"fmt"
+	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


NicolasMahe:
The stream from pull doesn't display nice info and it quite buggy...

antho1404:
I added a https://github.com/mesg-foundation/core/issues/138#issuecomment-388246570 in the issue. This work has no reason to be done and especially that there is way more important than this right now especially that we just discussed about that

PR #143, opened 2018-05-16
Example folder

antho1404:
A proposition to have access to all the services created on MESG, either the one that we develop or the one that other people develop like that we have a common place to find all the different services that developers can use

In order to add a new service you need to add it with the command
```
git submodule add GIT_REPO examples/service/xxxx
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,9 @@
+[submodule "examples/services/js-function"]
+	path = examples/services/js-function
+	url = https://github.com/mesg-foundation/service-js-function.git
+[submodule "examples/services/slack"]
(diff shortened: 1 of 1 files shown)


NicolasMahe:
I really don't like to have this example folder in the core repository and even less with submodule.

We could create a new repository called "awesome-mesg" with a list of links of github's applications and services.
That way, anyone will be able to do simple PR to add their own links and we separate core repo from app and services.
@antho1404 what do you think?

antho1404:
Agree with some real services these are just examples but maybe we can just in the readme links the different repo for those examples, that might be enough

PR #144, opened 2018-05-17
Make applications more developer friendly

antho1404:
## Proposal

This is a proposal to make application code way easier. right now there is too much boilerplate example:
https://github.com/mesg-foundation/application-devcon-update-on-slack/blob/master/main.go

We should be able to reduce it to only 
```When(service, "event").Then(service, "task", calculateInputs)```

This is one way to do that.

## Example application with those functions

The exact same application would be as follow
```go
package main

import (
	"os"

	"github.com/mesg-foundation/core/api/core"
)

func main() {
	devcon, _ := service.ImportFromPath("../../services/devcon-update")
	slack, _ := service.ImportFromPath("../../services/slack")

	core.
		When(devcon, "update").
		Then(slack, "notify", func(data *core.EventData) interface{} {
			return map[string]string{
				"channel": "conferences",
				"title":   "Update on https://devcon.ethereum.org",
				"token":   os.Getenv("SLACK_TOKEN"),
			}
		})

	core.StartWorkflow()
}
```

## Notes 

This code is just a proof of concept and needs to be re-written and tested, it's actually just a test

### DIFF
--- a/api/core/helpers.go
+++ b/api/core/helpers.go
@@ -1,0 +1,112 @@
+package core
+
+import (
+	"context"
+	"encoding/json"
+	"log"
+	"os"
+	"os/signal"
+	"strings"
+	"sync"
+	"syscall"
+
+	"github.com/mesg-foundation/core/service"
+	"google.golang.org/grpc"
+)
+
+type workflow struct {
+	service   *service.Service
+	event     string
+	listeners []*task
+}
+
+type task struct {
+	service *service.Service
+	task    string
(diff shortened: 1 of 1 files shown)


antho1404:
totally agree to move this in a different repo and have a separated library, this was just a proposal anyway but we can reuse to create an external lib. I will just keep this one open as a reminder

antho1404:
see issue https://github.com/mesg-foundation/core/issues/153

PR #146, opened 2018-05-20
Add Database package

NicolasMahe:
This PR contains two new package

## `database` package
- an interface that driver must implement
- a `Db` singleton that reference the driver to use

## `json` package
A basic database driver that read and write the data on the disk with a JSON representation.
I implement this driver first to be able to debug the database super easily.

I recommend to switch to https://github.com/syndtr/goleveldb relatively soon for better performance.


NicolasMahe:
@antho1404 i would like to have you comment about the tests.
As there are 2 packages so I wrote 2 test files, but they are almost identical.
When we will add more driver, I think we have tests per driver and also one test for the `database` package that is just an interface so it will use function already tested by drivers' tests.

NicolasMahe:
After a lot of test and headache because of go typing system, it's way better to use a binary database like https://github.com/syndtr/goleveldb.
I think we should absolutely not use the JSON database. It is not well written, not completely thread safe and no cache system.

LevelDB is super easy to use and seems super fast. It's again a Google creation and will be maintained for years.
The only difficulty with LevelDB was to encode and decode the struct in binary. The `encoding/json` is the simplest encode/decoder to use even if I use the bin and the string output version.
But them, I saw that LevelDB recommend to use ProtoBuf to encode and decode in binary. So I give a try in the `database/leveldbprotobuf`

NicolasMahe:
- [ ] re-organize the package to have a model by actions (like service). Can use leveldb and protobuf

NicolasMahe:
Good work 👍 I make micro changes (useless filter on iterator and remove verbose on CI test)

PR #154, opened 2018-05-25
Add library for applications

antho1404:
Here is an example of devcon update to slack

```golang
package main

import (
	"os"
	"os/signal"
	"syscall"

	"github.com/mesg-foundation/core/api/client"
)

func main() {

	workflow := client.When(&client.Event{
		Service: os.Getenv("SERVICE_DEVCON_ID"),
		Name:    "update",
	}).
		Then(&client.Task{
			Service: os.Getenv("SERVICE_SLACK_ID"),
			Name:    "notify",
			Inputs: func(interface{}) interface{} {
				return map[string]string{
					"channel": "conferences",
					"title":   "Update on https://devcon.ethereum.org",
					"token":   os.Getenv("SLACK_TOKEN"),
				}
			},
		})

	go workflow.Start()
	defer workflow.Stop()

	abort := make(chan os.Signal, 1)
	signal.Notify(abort, syscall.SIGINT, syscall.SIGTERM)
	<-abort
}
```


antho1404:
you can find an example of application using this api here https://github.com/mesg-foundation/application-devcon-update-on-slack/tree/master

NicolasMahe:
I added my reviews on specific stuff on the code.

I'm a bit surprised by the chaining of `Then` function.
If your write:
```go
workflow := client.When(&client.Event{
  ...
}).
Then(&client.Task{
  ....
}).
Then(&client.Task{
  ....
})
```

the 2 tasks will be executed with the same event and thus the second `Then` is a then on the event and not on the first then (unlike JS Promise chaining for example).

I would love to have either a real chaining of Then (like JS Promise). It will feel like a data workflow when every step can transform the input and returned it as output.

antho1404:
I think we should discuss about what we want this library to do. If we want to have multiple steps and everything it's should be quite similar to the workflow file in the end but is it something that we want for that ?

If we want to do all the chaining this is the role of the workflow file and this would add so much complexity on the lib I think. Maybe we can just remove the chaining and like that keep this simple

antho1404:
Here is a new version with a different interface but that make it really simple.

```go
package main

import (
  "os"
  "os/signal"
  "syscall"

  "github.com/mesg-foundation/core/api/client"
)

func main() {

  workflow := client.Workflow{
    OnEvent: &client.Event{Service: os.Getenv("SERVICE_DEVCON_ID"), Name: "update"},
    Execute: &client.Task{Service: os.Getenv("SERVICE_SLACK_ID"), Name: "notify",
      Inputs: func(interface{}) interface{} {
        return map[string]string{
          "channel": "conferences",
          "title":   "Update on https://devcon.ethereum.org",
          "token":   os.Getenv("SLACK_TOKEN"),
        }
      },
    },
  }

  go workflow.Start()
  defer workflow.Stop()

  abort := make(chan os.Signal, 1)
  signal.Notify(abort, syscall.SIGINT, syscall.SIGTERM)
  <-abort
}
```

I think the goal of this library is not to have something really complex, just some helpers and the complexity will come from the application if needed.

We could just add also a workflow with a `OnResult` as input event

antho1404:
You can find a fully functional example that uses events, results and tasks from 3 different services in 2 different languages here https://github.com/mesg-foundation/application-devcon-update-on-slack/tree/new-api

antho1404:
@NicolasMahe if everything is ok you can merge it

PR #169, opened 2018-05-29
Move docker related function to a dedicated package

### LINKED ISSUE #91: Refactor service package
In the services we have all the functions related to the services but also all the parts related to docker, this should be a separated package

NicolasMahe:
This PR add a new package container that handle all the docker related stuff

Close #91


service/start.go
```diff
@@ -18,9 +19,17 @@ func (service *Service) Start() (dockerServices []*swarm.Service, err error) {
 	}
 	// If there is one but not all services running stop to restart all
 	if service.IsPartiallyRunning() {
-		service.Stop()
+		for name, dependency := range service.GetDependencies() {
```
NicolasMahe: I had to change a little bit the logic here in order for the test `TestPartiallyRunningService` to pass. Otherwise the network was destroyed and not re-created.

container/container_test.go
```diff
@@ -0,0 +1,124 @@
...
+	startTestService(namespace)
+	<-WaitContainerStatus(namespace, RUNNING, time.Minute)
+
+	StopService(namespace)
+	err := <-WaitContainerStatus(namespace, STOPPED, time.Minute)
+	assert.Nil(t, err)
+}
+
+// WaitContainerStatus wait for the container to have the provided status until it reach the timeout
+func WaitContainerStatus(namespace []string, status StatusType, timeout time.Duration) (wait chan error) {
```
NicolasMahe: This function is here for the test to wait enough time to pass. It should be useless in the future (when the stop and start wait for container).

container/container.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"context"
+	"errors"
+
+	godocker "github.com/fsouza/go-dockerclient"
+)
+
+// FindContainerStrict returns a docker container if exist. Return error if not found.
+func FindContainerStrict(namespace []string) (container *godocker.APIContainers, err error) {
```
NicolasMahe: What do you think about this function Strict? the logic is to return an error if the container is nil.

container/network.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	client, err := Client()
+	if err != nil {
+		return
+	}
+	return client.RemoveNetwork(network.ID)
+}
+
+// FindNetworkStrict finds a Docker Network by a namespace. If no network if found, an error is returned.
+func FindNetworkStrict(namespace []string) (network *godocker.Network, err error) {
```
NicolasMahe: What do you think about this function Strict? the logic is to return an error if the container is nil.

antho1404: Same for this one I would use only the `FindNetwork` that returns an error if the network doesn't exists and we have only one function again 

Also this one is even worst because the `FindNetwork` uses the `FindNetworkStrict` and unset the error... that's a weird behavior for a function

service/dependency.go
```diff
@@ -58,40 +57,13 @@ func extractVolumes(service *Service, dependency *Dependency, details dependency
...
-	return
-}
-
 func dependencyStatus(namespace string, dependencyName string) (status StatusType) {
-	ctx := context.Background()
-	cli, err := dockerCli()
+	dockerStatus, err := container.ServiceStatus([]string{namespace, dependencyName})
 	if err != nil {
-		return
+		panic(err)
```
antho1404: this panic shouldn't exists here, I guess we should return an error for this function or have a default behavior like before where it returns `STOPPED` status is an `int` so the initial value is 0 so `STOPPED` so if it returns it will return stop.

Maybe we should instead return the error but anyway have a panic here is not a viable solution I think

service/start.go
antho1404: Yes just tried with your modification and without and the tests fails in both cases. The problem doesn't come from here and the `service.Stop()` should be more than enough in that case

antho1404: for some obscure reasons I restarted my docker and now the tests are ok with or without your modification. The docker api was "creating" the network with the same ID but when creating the service with this networkID I got an error that the network was not found and I had the network in my list of network in docker with `docker network list` but even a `docker rm ID` returned me an error that the network didn't exists even it was still on the list.

I guess there is a bug in docker about that but so this modification is not necessary and add complexity so I remove it

NicolasMahe: This fix was only for the CI tests. Locally, both cases also work.

container/container.go
NicolasMahe: Ok good, I also prefer to use the strict way.

container/network.go
NicolasMahe: This difference of behavior (compare to FindContainer and FindService) is because of go-dockerclient or maybe directly of docker api. All our Find functions should have the same behavior.

.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       - setup_remote_docker
       - run: go get -t ./...
       - run: docker swarm init
-      - run: env DAEMON.IMAGE=mesg/daemon:local go test -timeout 30s -p 1 -coverprofile=coverage.txt ./...
+      - run: env DAEMON.IMAGE=mesg/daemon:local go test -timeout 60s -p 1 -coverprofile=coverage.txt ./...
```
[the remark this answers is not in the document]
NicolasMahe: Because 30s is very close to some package test. I prefer to increase the timeout than having error because of timeout. Again it's depend on the CircleCI remote docker speed.

cmd/daemon/constants.go
```diff
@@ -1,6 +1,6 @@
 package daemon
 
 const (
-	name       = "mesg-daemon"
+	name       = "daemon"
```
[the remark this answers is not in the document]
NicolasMahe: The result is the same because the service use a Namespace function that add `MESG` too all services, container and networks that we create.

container/network.go
antho1404: So every functions like that should have the same behavior which is to be strict. If we ask to find for something it means it should be here otherwise we should have a search or something like that

service/dependency.go
NicolasMahe: I'm completely ok with you, I would prefer to return the error. I didn't to preserve the API and make the modification in a other PR.

PR #173, opened 2018-05-30
Use official docker sdk

NicolasMahe:
- [x] have to make sure that the change of behavior of all Find functions (throw error when no found) doesn't break anything


container/network.go
```diff
@@ -1,62 +1,67 @@
...
+	"github.com/docker/docker/api/types"
+	docker "github.com/docker/docker/client"
 )
 
 // CreateNetwork creates a Docker Network with a namespace
-func CreateNetwork(namespace []string) (network *godocker.Network, err error) {
-	network, err = FindNetwork(namespace)
-	if network != nil || err != nil {
+func CreateNetwork(namespace []string) (networkID string, err error) {
+	network, err := FindNetwork(namespace)
```
antho1404: I know you're still working on it but just as a reminder this should return a network object and wait for it if needed

container/shared_network.go
```diff
@@ -1,35 +1,38 @@
...
-				Driver:         "overlay",
-				Labels: map[string]string{
-					"com.docker.stack.namespace": namespace,
-				},
-			})
-		default:
-		}
+		return
+	}
+	if networkID.ID != "" {
```
antho1404: seems weird this

container/network.go
NicolasMahe: As the function from docker sdk `client.NetworkCreate` only return the networkID, I change the API.
I will revert the change of API and make this modification in another PR. Return the ID is actually enough, we never do anything with the full returned network or service.

container/shared_network.go
NicolasMahe: Ok yes, it should be `network.ID`.

container/network.go
NicolasMahe: Actually, I would like to leave it like this. We really never use anything else than the ID of the create thing. Tell me if you prefer having this in an other PR.

NicolasMahe:
@antho1404 ready to merge

container/client.go
```diff
@@ -5,64 +5,63 @@ import (
...
 }
 
 func createClient() (client *docker.Client, err error) {
-	client, err = docker.NewClientFromEnv()
+	client, err = docker.NewEnvClient()
 	if err != nil {
 		return
 	}
-	info, err := client.Info()
+	client.NegotiateAPIVersion(context.Background())
```
NicolasMahe: This set automatically the right API version with docker

cmd/service/utils.go
```diff
@@ -46,42 +41,15 @@ func loadService(path string) (importedService *service.Service) {
...
-			imageHash,
-			string(buf[:n]),
-		}, "")
-		if len(imageHash) > 0 {
-			break
-		}
-		time.Sleep(500 * time.Millisecond)
+	tag, err := container.Build(path, []string{name})
+	s.Stop()
+	if err != nil {
```
antho1404: why not returning the `err` like before ? is there any reason ?

antho1404: hum my bad it uses the `handleError` that exit so it's fine 👍

antho1404:
The build have some problems...
- [ ] errors when we have capital case on the name of the service like the devcon update
- [ ] the name of the image generated is mesg/service-xxx where in the mesg.yml we have mesg/xxx we will need to have a way to inject this automatically but for now at least we shouln't break this

Now I cannot run any services neither deploy some

NicolasMahe:
- The tag is now lowercased
- The tag prefix is now back to `mesg/`

@antho1404 please review

PR #175, opened 2018-05-31
Add ServiceOptions to describe docker service

NicolasMahe:
This PR should be merged after https://github.com/mesg-foundation/core/pull/173

This PR remove the dependencies from other package to docker by adding container.ServiceOptions.
It's only make the declaration of Docker service way easier and normalize them by adding labels in the right place.

### DIFF
--- a/container/service.go
+++ b/container/service.go
@@ -36,12 +36,13 @@ func FindService(namespace []string) (service swarm.Service, err error) {
 }
 
 // StartService starts a docker service
-func StartService(spec swarm.ServiceSpec) (serviceID string, err error) {
+func StartService(options ServiceOptions) (serviceID string, err error) {
 	client, err := Client()
 	if err != nil {
 		return
 	}
-	response, err := client.ServiceCreate(context.Background(), spec, types.ServiceCreateOptions{})
+	options.merge()
+	response, err := client.ServiceCreate(context.Background(), options.ServiceSpec, types.ServiceCreateOptions{})
 	if err != nil {
 		return
 	}
--- a/container/service.go
+++ b/container/service.go
@@ -41,8 +41,8 @@ func StartService(options ServiceOptions) (serviceID string, err error) {
 	if err != nil {
 		return
 	}
-	options.merge()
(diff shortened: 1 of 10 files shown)


container/service.go
```diff
@@ -1,137 +1,106 @@
...
 
 // StartService starts a docker service
-func StartService(options godocker.CreateServiceOptions) (dockerService *swarm.Service, err error) {
+func StartService(options ServiceOptions) (serviceID string, err error) {
 	client, err := Client()
 	if err != nil {
 		return
 	}
-	return client.CreateService(options)
+	options.merge()
```
antho1404: I really don't like this for many reasons
- first you don't know what you are merging here `option.merge()` but you give no parameters so it looks like you merge nothing so this function looks useless.
- also this kind of function should return something, merge means there is a modification in something so you need to get the update (like the function `append` for example)
- finally it's just not clear what this is doing and there is no reason to call this method if you don't know that you have to call it...

My proposition:

Rename this function with a `option.toDockerService()` that returns the object needed for docker like that the function name is clear and we can give the result directly in parameter of the `ServiceCreate` function and it's easier to test etc...

container/service_options.go
```diff
@@ -0,0 +1,131 @@
...
+	Published uint32
+}
+
+// Mount is a simplify version of mount.Mount
+type Mount struct {
+	Source string
+	Target string
+}
+
+func (options *ServiceOptions) merge() {
```
antho1404: I think if we want to have a proxy for these data it's fine but in that case we shouldn't add the `ServiceSpec` and like that this function is just a way to build the `ServiceSpec` and there is no "merge" which in most of cases are not even merge but more override so for me all this file doesn't really make sense. It's like trying to do something but half way, either we keep the original object (the docker one) or we create our own but not something in the middle especially because this add a lot more unnecessary complexity and we have to deal with these "merge" I really don't like that

NicolasMahe:
I removed `ServiceSpec` from the struct `ServiceOptions`.
Rename `merge` to `toSwarmServiceSpec`. `toSwarmServiceSpec` return a new `swarm.ServiceSpec`.

PR #179, opened 2018-05-31
API filter on events and results

### LINKED ISSUE #23: Filter messages from services
## Goal

- [x] Be able to be notified when we listen for an event only if it's a specific event, by default we should listen every events
- [x] Be able to listen a specific output from a specific task, by default we listen all the outputs of all the tasks

antho1404:
This allow to have more granularity on the events we listen from the core.

- [x] subscribes to events by key
- [x] subscribes to results by key
- [ ] add filter on data from events
- [ ] add filter on data from result

Also this request update the test command to be able to test an already deployed service and also keep a service alive

fix #23


antho1404:
I think this pull request should be done, the part to filter the data is not critical for now, we can do this in another pull request and also a bit tricky because we have to manage all the different types so let's keep this one simple and focus on more important developments.

For the subscribe you are right now we don't unsubscribe every time a stream is disconnected this should be added

NicolasMahe:
- [ ] should update documentation of service test command once merged.

cmd/service/test.go
```diff
@@ -83,41 +84,53 @@ func executeTask(serviceID string, task string, dataPath string) (execution *cor
...
 
 func init() {
-	Test.Flags().StringP("event", "e", "*", "Only log a specific event")
-	Test.Flags().StringP("task", "t", "", "Run a specific task")
+	Test.Flags().StringP("event-filter", "e", "*", "Only log the data of the given event")
+	Test.Flags().StringP("task", "t", "", "Run the given task")
 	Test.Flags().StringP("data", "d", "", "Path to the file containing the data required to run the task")
+	Test.Flags().StringP("task-filter", "r", "", "Only log the result of the given task")
+	Test.Flags().StringP("output-filter", "o", "", "Only log the data of the given output of a task result. If set, you also need to set the task in --result")
+	Test.Flags().StringP("serviceID", "s", "", "ID of a previously deployed service")
```
NicolasMahe: Flags `event`, `result`, `output` and `service` doesn't match with the API.
I renamed them with the API names.

NicolasMahe:
I create this https://github.com/mesg-foundation/core/issues/195 to track the improved version of the filter

PR #181, opened 2018-06-01
Make start and stop functions blocking

NicolasMahe:
- [x] Add `WaitForContainerStatus` func
It's using a `time.Sleep` to block the thread. Should I start a go routine that will sleep and on the main block with a waitgroup? I think the `time.Sleep` is enough

- [x] Update Daemon start

- [x] Update Daemon stop

- [x] Update daemon commands

- [x] Update Service start
It's using a `WaitGroup` to wait for all dependencies to start.
It's also using a `mutex` to edit `serviceIDs` array that is shared between the goroutines.

- [x] Update Service stop
It's using a `WaitGroup` to wait for all dependencies to stop.

- [x] Update service commands


container/container.go
```diff
@@ -48,3 +49,25 @@ func ContainerStatus(namespace []string) (status StatusType, err error) {
...
 	return
 }
+
+// WaitForContainerStatus wait for the container to have the provided status until it reach the timeout
+func WaitForContainerStatus(namespace []string, status StatusType, timeout time.Duration) (err error) {
+	start := time.Now()
+	for {
+		currentStatus, err := ContainerStatus(namespace)
+		if err != nil {
+			return err
```
antho1404: I would use some `break` here in order to make sure that we always exit the for loop, it's a bit easier to test

container/service_options.go
```diff
@@ -71,7 +74,12 @@ func (options *ServiceOptions) swarmPorts() (ports []swarm.PortConfig) {
 	return
 }
 
-func (options *ServiceOptions) swarmMounts() (mounts []mount.Mount) {
+func (options *ServiceOptions) swarmMounts(force bool) (mounts []mount.Mount) {
+	// hack for preventing mount when in CircleCI
```
antho1404: A bit hacky, I would add a comment TOFIX just to make sure that we can find this again when the CI allow to mount directories

container/type.go
```diff
@@ -8,3 +12,13 @@ const (
 	STOPPED StatusType = 1
 	RUNNING StatusType = 2
 )
+
+// TimeoutError represents an error of timeout
```
antho1404: I think this is not a really good way to have errors I like the fact that we have the details in the error message but i'm sure we can do it better, I would just use something like that directly
```go
errors.New(Timeout reached after " + e.duration.String() + " for ressource " + e.name)
```

or maybe have a function
```go
func timeoutError(duration, resource) error {
  return errors.New(Timeout reached after " + e.duration.String() + " for ressource " + e.name)
}
```

or just a constant with generic error message but like that we can compare this error in the tests for example.
```go
const TimeoutError = errors.New("Timeout error on container action")
```

This is a bit too much and confusing because you don't even return an error

service/start.go
```diff
@@ -26,18 +28,29 @@ func (service *Service) Start() (serviceIDs []string, err error) {
...
 			serviceName:    service.Name,
-		}, networkID)
-		i++
-		if err != nil {
-			break
 		}
+		wg.Add(1)
+		go func(service *Service, d dependencyDetails, name string, i int) {
+			serviceID, errStart := dependency.Start(service, d, networkID)
+			mutex.Lock()
```
antho1404: I'm not sure a mutex is needed here because in this code we have different goroutine that access different index so there is no reason to have conflicts but still is not something bad to do

service/start.go
```diff
@@ -78,4 +92,9 @@ func (dependency *Dependency) Start(service *Service, details dependencyDetails,
 		Ports:      extractPorts(dependency),
 		NetworksID: []string{networkID, sharedNetworkID},
 	})
+	if err != nil {
+		return
+	}
+	err = container.WaitForContainerStatus(namespace, container.RUNNING, time.Minute) //TODO: be careful with timeout
```
antho1404: a minute is really long. If it was in the cli it might be ok but for an internal function I think it's definitely too long

service/stop.go
```diff
@@ -22,20 +25,34 @@ func (service *Service) Stop() (err error) {
...
-			return
-		}
+		wg.Add(1)
+		go func(d *Dependency, name string, dependencyName string) {
+			errStop := d.Stop(name, dependencyName)
+			mutex.Lock()
+			if errStop != nil && err == nil {
+				err = errStop
+			}
+			mutex.Unlock()
```
antho1404: same remark here about using `defer`

antho1404: also here i don't see why there is a mutex, in the map it makes sense because it's not thread safe if two threads want to write on the same index but with the error I don't think there is any problems

service/stop.go
```diff
@@ -22,20 +25,34 @@ func (service *Service) Stop() (err error) {
...
+	if !dependency.IsRunning(name, dependencyName) {
+		return
+	}
+	namespace := []string{name, dependencyName} //TODO: refacto namespace
+	err = container.StopService(namespace)
+	if err != nil {
 		return
 	}
-	err = container.StopService([]string{namespace, dependencyName})
+	err = container.WaitForContainerStatus(namespace, container.STOPPED, time.Minute) //TODO: be careful with timeout
```
antho1404: again probably too long one minute

container/container.go
NicolasMahe: Because of the new context of the `for` and the creation of `currentStatus`, `err` is also created. So if I use `break`, I'm pretty sure the err will not be assign to the return `err`.
If I write, the following code, go says `err is shadowed during return` on the `return`.
```go
for {
    currentStatus, err := ContainerStatus(namespace)
    if err != nil {
	return
```

So there are 2 solutions.
The first one is to use `return err` as it's already the case.
The second is to created `currentStatus` and then assign the value:
```go
var currentStatus StatusType
currentStatus, err = ContainerStatus(namespace)
if err != nil {
	break
}
```

@antho1404 What solution do you think it's the best?

container/type.go
NicolasMahe: It's the classic way to create new errors in go and be able to compare. Eg with type casting:
```go
_, ok := err.(*TimeoutError)
```
https://blog.golang.org/error-handling-and-go

On the contrary, I think we should start creating custom error than using the default one with just a custom description. It will be way easier to debug.

container/service_options.go
NicolasMahe: Comment updated

service/stop.go
NicolasMahe: Updated to use defer.

NicolasMahe: The reason for the mutex is the `err` comparison in the `if`:
```go
if .... err == nil {
```

service/start.go
NicolasMahe: What timeout you think is ok?

Otherwise, we could pass the timeout duration as a parameter of the `Start` and `Stop` function.

NicolasMahe:
Changelog updated.
Please review my comments of your comments.

container/container.go
antho1404: I personally use the second solution all the time, we should definitely avoid shadowed variable it's the kind of things that can create bugs easily.

container/type.go
antho1404: it just feels like a lot of code just for a simple error that could be done in one constant but ok, let's follow the guidelines

antho1404:
still have some errors from codeclimate

NicolasMahe:
I'm merging even with the codeclimate issues. I don't know how to fix it in a simple way..

PR #183, opened 2018-06-01
Integrate service image automatically

### LINKED ISSUE #180: Generate better tag for service build
## Proposal
When building service image with the cmd, we should generate a unique tag that include a hash of the service, the `mesg/service-` prefix and the name of the service.
Then this tag should be either injected in the mesg.yml or at least in the struct of the service before starting it. Maybe this tag can be inject in a `.mesg` file and in the `mesg.yml` add a reference.
We should also manage the version of the tag.

## Solution

Inject image tag in a `.mesg` file.
Use a special variable/keywork to reference the tag in the `mesg.yml`
Automatic 'replacement' when reading the `mesg.yml` with the value in the `.mesg`.

`mesg.yml`:
```yml
name: function
description: execute a function
tasks:
....
dependencies:
  function:
    image: MESG_IMAGE
    volumeFrom: ethereum
  ethereum:
    image: geth/geth
```

`.mesg`
```yml
MESG_IMAGE: mesg/service-function-5c7f199c-6547-11e8-adc0-fa7ae01bbebc
```

Service struct after reading the `mesg.yml` and `.mesg`:
```go
service.Dependencies = map[string]*s.Dependency{
	"function": &s.Dependency{
		Image: "mesg/service-function-5c7f199c-6547-11e8-adc0-fa7ae01bbebc",
		VolumeFrom: "ethereum",
	"ethereum": &s.Dependency{
		Image: "geth/geth",
	},
}
```

antho1404:
This will simplify the development and deployment of a service, users will not have to generate dependencies for their docker, this will be automatically generated but they can always configure their container or add additional dependencies to their services

fix #180


antho1404:
ready for review :)

cmd/service/utils.go
```diff
@@ -41,18 +41,36 @@ func loadService(path string) (importedService *service.Service) {
...
 	if err != nil {
 		handleError(err)
 		return
 	}
-	fmt.Println(aurora.Green("Image built with success. Tagged: " + tag))
+	fmt.Println(aurora.Green("Image built with success. Tagged: " + imageHash))
 	return
 }
 
+func tagService(s *service.Service, tag string) {
```
NicolasMahe: This should not be named `tagService` but something with `hash`

PR #186, opened 2018-06-02
fix codeclimate warnings

antho1404:


container/container.go
```diff
@@ -32,8 +32,8 @@ func FindContainer(namespace []string) (container types.ContainerJSON, err error
 	return
 }
 
-// ContainerStatus returns the status of a docker container
-func ContainerStatus(namespace []string) (status StatusType, err error) {
+// Status returns the status of a docker container
+func Status(namespace []string) (status StatusType, err error) {
```
NicolasMahe: @antho1404 Why this modification? Ok it was throwing a warning, but only naming it `Status` remove all the context. Is it a Service status? A network? A container?

PR #187, opened 2018-06-03
Stop services when the daemon stops

### LINKED ISSUE #158: Stop services when stopping daemon

antho1404:
In this PR I added a new label for docker services `mesg.hash`

With this label, we can now from a docker service retrieve the service in the database and do whatever we want. In this case just before stoping the daemon we can stop all the running services

fix #158


cmd/daemon/stop.go
```diff
@@ -28,3 +38,33 @@ func stopHandler(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		return
+	}
+	hashes, err := service.List()
+	if err != nil {
+		return
+	}
+	for _, hash := range hashes {
+		// TODO: this function should be execute in a go routine. StopService is blocking
+		_, err := cli.StopService(context.Background(), &core.StopServiceRequest{
```
NicolasMahe: This should be called in a async fashion to stop all services at the same time.

NicolasMahe:
I updated `daemon.stopServices` function. Now it's calling `cli.StopService` in an async way in order to stop all services at the same time.

PR #190, opened 2018-06-05
Possibility to fetch deploy a service from git

### LINKED ISSUE #157: Publish a service from a repo
We should accept github repo on the `service publish` command like that for tutorials and example people can just type the command line and it will automatically download all the sources and then publish it

antho1404:
Now it is possible to automatically fetch the sources from any git repository from the cli.

The following commands are now working and behave like expected (fetch the repository and then build and deploy or test the service)

- `mesg-core service test https://github.com/mesg-foundation/service-ethereum.git`
- `mesg-core service deploy https://github.com/mesg-foundation/service-ethereum.git`

For now it's not testable because example are already updated according to the pull request #183, we need to merge this one first and test again that everything works fine

fix #157


cmd/service/utils.go
```diff
@@ -31,8 +35,32 @@ func handleError(err error) {
 	}
 }
 
+func gitClone(url string) (path string, err error) {
+	tmpFile := "/tmp/mesg-templates"
```
NicolasMahe: This is very ugly to put the path in a root folder like this.
It's not even deleted at the end.

cmd/service/utils.go
```diff
@@ -31,8 +35,32 @@ func handleError(err error) {
...
+	err = os.MkdirAll(tmpFile, os.ModePerm)
+	if err != nil {
+		return
+	}
+	path, err = ioutil.TempDir(tmpFile, string(time.Now().UnixNano()))
```
NicolasMahe: On my computer `path` is like: `/tmp/mesg-templates/�094045028`
There is a encoding error. You should `strconv.FormatInt(time.Now().UnixNano(), 10))` instead of `string(....)`

cmd/service/utils.go
```diff
@@ -31,8 +35,32 @@ func handleError(err error) {
...
+		URL: url,
+	})
+	return
+}
+
 func loadService(path string) (importedService *service.Service) {
-	importedService, err := service.ImportFromPath(path)
+	var err error
+	if _, err = url.ParseRequestURI(path); err == nil {
+		path, err = gitClone(path)
```
NicolasMahe: The new `path` is not forward to the function `buildDockerImage` thus it cannot build it..

cmd/service/utils.go
antho1404: even the `TempDir` accepts empty string for defaults and manages everything so I used that and it takes the defaults from that function better than trying to create a specific folder that might have some conflicts

PR #206, opened 2018-06-07
Remove daemon wording

### LINKED ISSUE #204: Remove Daemon wording
The software that people download and run locally should be `mesg-cli`.
The docker container should be named `mesg/core`.

Daemon is not a name but only a way to start a software.
> In multitasking computer operating systems, a daemon is a computer program that runs as a background process, rather than being under the direct control of an interactive user.

We should rename the daemon package and commands accordingly.

NicolasMahe:
close #204 

New docker hub created: https://hub.docker.com/r/mesg/core/

I didn't change the behavior of the `mesg-core start` command. It launch the core in docker.
@antho1404 do you want me to add the `--daemon` flag now and change the behavior when the flag is not set? (launch the core not in docker)


daemon/start.go
```diff
@@ -33,11 +33,11 @@ func serviceSpec() (spec container.ServiceOptions, err error) {
 	}
 	spec = container.ServiceOptions{
 		Namespace: Namespace(),
-		Image:     viper.GetString(config.DaemonImage),
+		Image:     viper.GetString(config.CoreImage),
 		Env: []string{
 			"MESG.PATH=/mesg",
-			"API.SERVICE.SOCKETPATH=" + filepath.Join(viper.GetString(config.MESGPath), "server.sock"),
-			"SERVICE.PATH.HOST=" + filepath.Join(viper.GetString(config.MESGPath), "services"),
+			"API.SERVICE.SOCKETPATH=" + filepath.Join(viper.GetString(config.MESGPath), "server.sock"), // TODO: this one seems useless because MESGPath is already set.
```
antho1404: not useless because here the path will be from the `MESGPath` while executing this command so from the CLI so with the `~/.mesg` of the user, if not set, inside the docker it will bind to the `MESG.PATH` that is `/mesg` so the link will not be good because on the host user will not have `/mesg`

NicolasMahe:
I renamed `core` to `MESG Core` in logs

PR #207, opened 2018-06-07
[WIP] Update README

w1l1ams:


README.md
```diff
@@ -17,7 +17,7 @@
 
 MESG is a platform for the creation of efficient and easy-to-maintain applications that connect any and all technologies.
 
-# Issues
+# ISSUES
```
NicolasMahe: Why uppercase?

antho1404:
I think we need to put a menu to have a quick access to the different blocks 

Also I really think the build from source shouldn't be directly in the readme, the readme should be oriented for users, I create a wiki, we can just in the contribution part give a link to the wiki [https://github.com/mesg-foundation/core/wiki](https://github.com/mesg-foundation/core/wiki) that should be enough otherwise it might confuse users

Also for the part about issues and contribution this is the same, an issue is already a contribution so we should only have one block contribution and we should use the process that everyone knows about, if there is an issue just check if it's not already in github and otherwise post an issue with the right flags but let's not redirect them to discord for this, it's not really effective, discord is a chat and what we will tell them anyway is to go to create an issue on github... I agree on the part to remind them that they have to check the documentation make sure it's not from them and otherwise the directly create an issue and even if it's from their understanding but still something doesn't work it means we need to have better documentations or examples to be clearer so an issue will be valid

NicolasMahe:
@w1l1ams I just commit my modification. Could you read and fix any english mistakes please?

NicolasMahe:
@antho1404 The js lib is `mesg-js` and not `mesg`?
https://www.npmjs.com/package/mesg-js

It was `mesg` in the example. Just to be sure I don't include mistake.

PR #212, opened 2018-06-12
Use and inject hash to identify service

antho1404:
With this pull request, the API for services is now updated and doesn't accept a service anymore but a hash that will need to send back to identify itself.

I also created the pull request https://github.com/mesg-foundation/mesg-js/pull/3 to update the javascript library


antho1404:
when this is merged we also need to update the documentation

api/service/api.proto
```diff
@@ -3,22 +3,20 @@ syntax = "proto3";
...
-
 service Service {
   rpc EmitEvent (EmitEventRequest) returns (EmitEventReply) {}
   rpc ListenTask (ListenTaskRequest) returns (stream TaskData) {}
   rpc SubmitResult (SubmitResultRequest) returns (SubmitResultReply) {}
 }
 
 message EmitEventRequest {
-  service.Service service = 1;
+  string serviceHash = 1;
```
NicolasMahe: Why naming it `ServiceHash`? We already have `ServiceID` in the core api.

service/start.go
```diff
@@ -84,6 +84,7 @@ func (dependency *Dependency) Start(service *Service, details dependencyDetails,
 		Image: dependency.Image,
 		Args:  strings.Fields(dependency.Command),
 		Env: []string{
+			"MESG_SERVICE_HASH=" + service.Hash(),
```
NicolasMahe: I think we should rename it `MESG_SERVICE_ID`

NicolasMahe: We could also rename the function `hash` to `id`.
Even if technically it's a hash of the service, we are using it as a ID. So it easier to understand if it named `id`

api/service/api.proto
antho1404: the idea is that this hash can be something else than the id of the service, can be the public address or stuff like that, I was trying to avoid `serviceID` to not be confusing because how the developer can have the id when the id is generated when you deploy the service.

For now we use the service ID as identifier but we might change that with a random string created when you start the service or a uuid or anything like that but yes maybe it should be more generic like "SECRET" or "AUTHORIZATION" or "TOKEN" stuff like that what do you think @NicolasMahe

antho1404:
I used `token` everywhere, it's more generic and i updated the js lib too

NicolasMahe:
doc updated

Issue #200, opened 2018-06-05
Improve service and dependencies namespace


NicolasMahe:
We should create a function for the dependency namespace.
In a lot of place in the code, we have stuff like:
```go
service.namespace(), name
```
We should have one function that return the full namespace of the service + dependency

PR #223, opened 2018-06-14
Improve service and dependencies namespace

NicolasMahe:
- Add struct `DependencyFromService` to associate a Dependency, a Service and the dependency name.
- Use it to solve the complexity of the namespace

Close https://github.com/mesg-foundation/core/issues/200


antho1404:
I reduced the complexity of the `DependencyFromService.Start` by moving the `WaitForContainerStatus` function directly in the `container.Start` function and to have this consistent I did the same for the `container.Stop` function and so `WaitForContainerStatus` can be private in this package now

Issue #215, opened 2018-06-13
Add service logs command


NicolasMahe:
- [x] We should add a command to return the logs of the services` containers instead of relaying on docker cli or portainer. Done in https://github.com/mesg-foundation/core/pull/222
- [x] We should copy how docker compose do it: by default, every dependency in the same logs output, filtering by flags. Done in https://github.com/mesg-foundation/core/pull/222
- [ ] Add a prefix in the log for every dependency: should be the name or/and different color.

PR #228, opened 2018-06-15
Improve function status

### LINKED ISSUE #199: Improve service status
The file `service/status.go` and `service/dependency.go` should be improved:

- [x] func `dependencyStatus` has a panic. it should return the error.
- [x] Are the func helper `IsRunning`, `IsPartiallyRunning` and `IsStopped` really useful? They don't returned errors and are barely used. It will be as easy to use directly `serviceStatus` and `dependencyStatus`.
- [x] func `serviceStatus` should also return the error from `dependencyStatus`
- [ ] The function `List` should move to its own file.

NicolasMahe:
- Add error to status function
- Remove helper isRunning, isStopped

Dependent on #200 and #215 
Close #199


service/status.go
```diff
@@ -9,16 +9,22 @@ type StatusType uint
 
 // status for services
 const (
-	STOPPED StatusType = 0
-	RUNNING StatusType = 1
-	PARTIAL StatusType = 2
+	STOPPED StatusType = 1
```
antho1404: Why changing this ?
`0` is good like that if it's not initialized then by default the status is `STOPPED`

NicolasMahe: I change it exactly to avoid to have a not initialized version at `STOPPED`.
But now, I'm not sure if it better.

antho1404: I think it's better to revert this because what happen if one moment we don't initialize a status at least it will be marked as stopped and not as nothing that will be really hard to track down

PR #233, opened 2018-06-17
Add logs to test command

antho1404:
Now the test command also have the logs from the service and the logs from all dependencies with the `--logs-all` flag


cmd/service/logs.go
```diff
@@ -27,20 +25,18 @@ func init() {
...
 	readers, err := service.Logs(dependency)
-	if err != nil {
-		fmt.Println(aurora.Red(err))
-		return
-	}
+	utils.HandleError(err)
 	for _, reader := range readers {
 		defer reader.Close()
-		go stdcopy.StdCopy(os.Stdout, os.Stderr, reader)
+		stdcopy.StdCopy(os.Stdout, os.Stderr, reader)
```
NicolasMahe: Remove the goroutine prevents to see the logs from many dependencies..

NicolasMahe: Adding the goroutine, without having the `<-utils.WaitForCancel()` in the same function make the function executes the `defer reader.Close()`.
We could return a function `close` that will call the `reader.Close()`. What do you think?

cmd/service/test.go
```diff
@@ -38,6 +38,7 @@ func init() {
 	Test.Flags().StringP("event-filter", "e", "*", "Only log the data of the given event")
 	Test.Flags().StringP("task-filter", "r", "", "Only log the result of the given task")
 	Test.Flags().StringP("output-filter", "o", "", "Only log the data of the given output of a task result. If set, you also need to set the task in --task-filter")
+	Test.Flags().BoolP("full-logs", "f", false, "Display logs from service and its dependencies")
```
NicolasMahe: The flag `full-logs` is not explicit enough.
I suggest the name `logs-dependencies` and as explanation: "Also displays the logs of all service's dependencies".
Also please add this flag in the Example of this command.

antho1404: `logs-dependencies` is even more confusing I think because we don't know if we will have only the dependencies or all

NicolasMahe: `logs-all` ?

PR #235, opened 2018-06-17
Add core api

antho1404:
Add the following APIs
- `rpc ListServices (ListServicesRequest) returns (ListServicesReply) {}`
- `rpc GetService (ServiceRequest) returns (ServiceReply) {}`


api/core/api.proto
```diff
@@ -13,6 +13,8 @@ service Core {
   rpc StopService (StopServiceRequest) returns (StopServiceReply) {}
   rpc DeployService (DeployServiceRequest) returns (DeployServiceReply) {}
   rpc DeleteService (DeleteServiceRequest) returns (DeleteServiceReply) {}
+  rpc ListServices (ListServicesRequest) returns (ListServicesReply) {}
+  rpc Service (ServiceRequest) returns (ServiceReply) {}
```
NicolasMahe: I suggest to name `Service` by `GetService`. It will be better for the uniformization and explicitly of the API

antho1404: done

NicolasMahe: Don't forget to rename the files `api/core/service.go` and `api/core/service_test.go`.

PR #246, opened 2018-06-21
Add .mesgignore file to exclude files for the Docker build

NicolasMahe:


CHANGELOG.md
```diff
@@ -23,12 +23,14 @@
 - (#190) Possibility to `test` or `deploy` a service from a git or GitHub url
 - (#233) Add logs in the `service test` command with service logs by default and all dependencies logs with the `--full-logs` flag
 - (#235) Add `ListServices` and `GetService` APIs
+- (#) Add .mesgignore to excluding file from the Docker build
```
antho1404: This should not be places in this part of the release but in the unreleased part because the v1.0.0-beta has been released already

container/build.go
```diff
@@ -19,7 +20,15 @@ type BuildResponse struct {
 
 // Build a docker image
 func Build(path string) (tag string, err error) {
-	buildContext, err := archive.Tar(path, archive.Gzip)
+	excludeFilesBytes, err := ioutil.ReadFile(filepath.Join(path, ".mesgignore"))
```
antho1404: This is not valid because this assume that the `.mesgignore` file is required which is not true

PR #247, opened 2018-06-22
Update service init command

### LINKED ISSUE #244: Command service init generate a not valid mesg.yml
The command export `configuration: null` to the `mesg.yml` file that is not valid.

Use go template instead of marshal.

- Remove `visibility` and `publish`
- remove all empty variable
- add example like:
```yml
name: ERC20
description: ""
tasks:
  foo:
    name:
    outputs:
events:
  foo:
    data:
      bar:
        type: String
```

antho1404:
Now the `service init` use a template to generate the `mesg.yml` file and correct a bug because of the dependency attribute

fix #244

### DIFF
--- a/cmd/service/init.go
+++ b/cmd/service/init.go
@@ -1,18 +1,42 @@
 package service
 
 import (
+	"bytes"
 	"fmt"
+	"html/template"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 
 	"github.com/logrusorgru/aurora"
+	"github.com/mesg-foundation/core/cmd/utils"
 	"github.com/mesg-foundation/core/service"
 	"github.com/spf13/cobra"
 	"gopkg.in/AlecAivazis/survey.v1"
-	"gopkg.in/yaml.v2"
 )
+
+const templateText = `name: "{{.Name}}"
+description: "{{.Description}}"
+tasks:
+  foo:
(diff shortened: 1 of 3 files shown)


cmd/service/init.go
```diff
@@ -32,100 +57,76 @@ func initHandler(cmd *cobra.Command, args []string) {
...
-		string(service.VisibilityNone),
-	})
-	res.Publish = askOpts("Publish (ALL):", cmd.Flag("publish").Value.String(), []string{
-		string(service.PublishAll),
-		string(service.PublishSource),
-		string(service.PublishContainer),
-		string(service.PublishNone),
-	})
+func buildService(cmd *cobra.Command) (res *service.Service) {
+	res = &service.Service{}
```
NicolasMahe: I had to init the `res` variable otherwise memory error.

cmd/service/init.go
```diff
@@ -32,100 +57,76 @@ func initHandler(cmd *cobra.Command, args []string) {
...
+		return
 	}
-	return value
+	err = tmpl.Execute(&doc, service)
+	res = doc.Bytes()
+	return
 }
 
-func ask(label string, value string) string {
+func ask(label string, value string, validator survey.Validator) string {
```
NicolasMahe: Name is now mandatory

PR #267, opened 2018-06-29
Task to generate a default documentation for service

### LINKED ISSUE #258: Export service definition in README.md
It is really annoying and long to create the service definition/doc in the `README.md`.

We should create a command `service gen-doc` that export the service definition from the `mesg.yml` to the `README.md`.

If the `README.md` already exist, the command should append the definition to it.

antho1404:
We can now use the command `mesg-core service gen-doc PATH_TO_SERVICE` to generate a readme based on the mesg.yml file

# WARNING:
Need to update the documentation to add the new attribute repository on the service file

Close #258


antho1404:
## WARNING:

Need to update the documentation to add the new attribute `repository` on the service file

Issue #279, opened 2018-07-13
Allow branch download in the command service deploy


NicolasMahe:
## Problem

We cannot download a specific branch when using the command service deploy.

## Solution

Be compatible with git urls that contain branch info. Eg:

```
mesg-core service deploy https://github.com/mesg-foundation/service-ethereum-erc20#mesg-token-ropsten
```

> Make sure that other command that download service also become compatible with this feature

Related to https://github.com/mesg-foundation/service-ethereum-erc20/issues/4

NicolasMahe:
Let's discuss more if it really useful.

ilgooz:
@NicolasMahe I think it's useful to deploy different branches to different development or production environments.

pts-solutions:
just a thought, since we are using go-git, should we also aim to support more protocols than just HTTP(s)? In particular ssh:// which is quite frequently used and according to the [compatibility chart](https://github.com/src-d/go-git/blob/master/COMPATIBILITY.md) it should work out of the box

PR #282, opened 2018-07-14
cmd/service: deploy: support custom branches and urls without a schema

ilgooz:
closes https://github.com/mesg-foundation/core/issues/279

### DIFF
--- a/cmd/service/utils.go
+++ b/cmd/service/utils.go
@@ -73,16 +73,14 @@ func gitClone(repoURL string, path string, message string) (err error) {
 	if u.Scheme == "" {
 		u.Scheme = "https"
 	}
-	ref := plumbing.Master
+	options := &git.CloneOptions{}
 	if u.Fragment != "" {
-		ref = plumbing.ReferenceName("refs/heads/" + u.Fragment)
+		options.ReferenceName = plumbing.ReferenceName("refs/heads/" + u.Fragment)
 		u.Fragment = ""
 	}
+	options.URL = u.String()
 	utils.ShowSpinnerForFunc(utils.SpinnerOptions{Text: message}, func() {
-		_, err = git.PlainClone(path, false, &git.CloneOptions{
-			URL:           u.String(),
-			ReferenceName: ref,
-		})
+		_, err = git.PlainClone(path, false, options)
 	})
 	return
 }
--- a/cmd/service/utils.go
+++ b/cmd/service/utils.go
@@ -3,7 +3,10 @@ package service
 import (
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
+
+	"gopkg.in/src-d/go-git.v4/plumbing"
 
 	"github.com/asaskevich/govalidator"
 	"github.com/logrusorgru/aurora"
@@ -50,7 +53,6 @@ func prepareService(path string) (importedService *service.Service) {
 }
 
 func downloadServiceIfNeeded(path string) (newPath string, didDownload bool, err error) {
-	didDownload = false
 	newPath = path
 	if govalidator.IsURL(path) {
 		newPath, err = createTempFolder()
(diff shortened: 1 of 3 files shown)


pts-solutions:
hi ilgooz, i am afraid that your solution will not quite work. the govalidator.IsURL in downloadServiceIfNeeded will evaluate to false on schema-less URLs and therefore never call gitClone

ilgooz:
Hi @pts-solutions,
Thank you for your second check on this. It seems govalidator does support urls without a schema. Please see here: https://github.com/asaskevich/govalidator/blob/master/validator_test.go#L708

pts-solutions:
Yeap sorry my bad, for some reason i tested with localhost/test#test which evaluates to false, but as soon as it contains a (dot) or a (colon) it works

cmd/service/utils.go
```diff
@@ -50,23 +53,35 @@ func prepareService(path string) (importedService *service.Service) {
...
-func gitClone(url string, path string, message string) (err error) {
+func gitClone(repoURL string, path string, message string) (err error) {
+	u, err := url.Parse(repoURL)
+	if err != nil {
+		return err
+	}
+	if u.Scheme == "" {
+		u.Scheme = "https"
+	}
+	ref := plumbing.Master
```
antho1404: what about when the default branch is different than `master`. In case of the core for example we use the `dev` as a default branch.
Maybe keep the ref empty would be a better solution like that we really keep the default behavior to take de default branch and not force the `master`

cmd/service/utils.go
```diff
@@ -50,23 +53,35 @@ func prepareService(path string) (importedService *service.Service) {
...
-	didDownload = false
 	newPath = path
 	if govalidator.IsURL(path) {
 		newPath, err = createTempFolder()
 		if err != nil {
 			return
 		}
 		err = gitClone(path, newPath, "Downloading service...")
-		didDownload = true
+		didDownload = err == nil
```
antho1404: nice catch on this one :)

antho1404:
That's perfect, if you can just add a line with the details of this feature on the `changelog.md` file with the id of this pull request and a small description of in the `unreleased` part :)

ilgooz:
So, I gave up on supporting https urls without a schema in favor of supporting git file:// protocol. Otherwise there is no way to understand the difference in between. Which also means we now support all git compatible urls that supported by git-go (https:// ssh:// git:// file://).

There are also two new flags for test and deploy commands for ssh. `--private-key-path`, `--passphrase`
You can also use basic auth with https thanks to go-git.

I removed two tests called `TestDownloadServiceIfNeededAbsolutePath` and `TestDownloadServiceIfNeededRelativePath`because go-git thinks them as file protocol compatible urls and our tests won't pass because the urls are fake. I don't know these tests are for.

* Any suggestions on codeclimate issue?
* Any suggestions on deleted tests, flags or the new feature?

antho1404:
I'm not a big fan of the latest developments, I think it adds too much complexity for the user and I think not so many user (maybe even none) will use this feature. I actually don't remember even one time where I needed to use this kind of access.
I would recommend KISS (Keep It Simple Stupid). We can add this feature if users really need it but for now it's too much I think.
@NicolasMahe any feedbacks ?

NicolasMahe:
@ilgooz The issue is only about specifying the branch, not be compatible with private keys.
I suggest to revert your last commit.

ilgooz:
@antho1404 @NicolasMahe thank you for both of you guys for taking time and checking out. 
I strongly believe that we should have a natural support of private git repositories. I gave a second look on go-git and it seems they have a full implementation of git all in Go, so we needed to use it's auth option to have access on private repositories. Also sometimes providing a path for private keys in command-line level helps with the CI.

@NicolasMahe I re-checked the code after the panic you have running the code but I couldn't reproduce it. Actually we needed to add more tests about the changes but couldn't since it requires us to have special private repositories for this purpose.

```shell
Ilkers-MacBook-Pro:cli ilgooz$ go run main.go service test git@github.com:mesg-foundation/service-ethereum-erc20
ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
``` 

```shell
Ilkers-MacBook-Pro:cli ilgooz$ go run main.go service test git@github.com:mesg-foundation/service-ethereum-erc20 --private-key-path /Users/ilgooz/.ssh/id_rsa
Service downloaded with success
Temp folder: /var/folders/bn/wg5vt5ls2ml_77gd4rw09q800000gn/T/mesg-086168266
Image built with success
Image hash: sha256:51788939cdc33a39fa91e5df92a1128a6f2d8bca87e3ec6cc3b0c78b0fd7605b
Service deployed with success
Service ID: v1_60c329bb45f66aee367f7caa864acdc1
Service startedice... 
Listening for results from the service...
Listening for events from the service...
new block 5981311
```

So I'm reverting the last commit in favor of our branch support. We maybe can do a further discussion about this feature in another issue, if you think it's needed.

NicolasMahe:
@ilgooz PR merged. Thank you 👍

Issue #281, opened 2018-07-13
change every "golang.org/x/net/context" to new "context"


ilgooz:

antho1404:
Go added the `context` as a standard library

more details here https://golang.org/doc/go1.7#context

PR #285, opened 2018-07-16
Feature/context

ilgooz:
tests seems passing with the new version of base images
closes https://github.com/mesg-foundation/core/issues/281


antho1404:
could you run the `gofmt` command, it seems that you have some issues just because the standard libraries needs to be included before external libraries. Nothing big but it's good to have this kind of checks to have a consistency it the codebase.
This should fix the codeclimate issues

PR #295, opened 2018-07-20
Add validation of all the different payload in the api

### LINKED ISSUE #194: Should return error when service emit an invalid event or result of task
The core should throw an error when it receives non valid event or result of task from services.
It will prevent developer to send event and result of task that are not defined in the `mesg.yml` file.

It should also check for variable types, and optional/required as defined in the `mesg.yml`

- [x] validate input data
- [x] validate event data
- [x] validate result data
- [x] add strict verification when the payload doesn't have enough data (compare to the mesg.yml)
- [ ] add strict verification when the payload have too much data (compare to the mesg.yml)
- [x] only accept correct payload `map[string]interface{}`

antho1404:
The API didn't had any validation and this pull request will validate the following data sent to the API:
- inputs when `executeTask`
- results when `submitResult`
- event when `emitEvent`

The validate makes sure that all required parameters are presents (based on the `optional` tag). It also check the type of the data, either `String`, `Number` (float or int), `Boolean`, `Object` (either json object or array)

closes #194


antho1404:
The issue from the build comes from here https://github.com/asaskevich/govalidator/issues/295

antho1404:
modifications done

Issue #273, opened 2018-07-06
Building on windows


tantanchen:
Building from source on windows wasn't too difficult but it looks like GitBash won't allow `service init` to select an option. It just terminates the program
![image](https://user-images.githubusercontent.com/14248948/42393177-84eae940-8123-11e8-8668-36d8944c0262.png)

antho1404:
Thanks a lot for this issue and for taking the time to build on windows :)
I think the issue might be related to a path that is a different system between unix system and window system but it definitely needs some investigations

pts-solutions:
i have a great interest in getting to the bottom of this as well, basically what i tried is to use windows containers, meaning not going via mingw. I think i found something suspicious worth investigating with the was sockets are used, via URL unix:///tmp/my_socket. the bottom line is that when you try to use mesg-core outside a unix-like environment you basically hit a brick wall, but i will keep digging :)

NicolasMahe:
Yes, MESG Core's container try to access Docker via socket.. We should update this to be compatible in not UNIX environment.
Also, a lot of path are hardcoded with `/`. We should also update them by the path functions that are compatible with non UNIX environment.

If you want, I can dig the codebase and make a list of things that we should update for the MESG Core to be compatible with Windows. Tell me if you want this list ;)

pts-solutions:
yes that would be great to have such a list, as i am a little limited on spare time right now.

maybe a few more insights that might help. there are actually 2 ways you can work with docker containers under windows. either you create linux containers or windows containers. for simplicity sake i think we should focus on linux containers for now, which means that most likely whatever is running inside the container (core and/or daemon) will not need much changes, except where there is mounts of local windows directories. but what i think needs changes is the cli, which would be run in either windows cmd or ps (powershell).

what makes this a little slow for me is that i don't know everything yet about the codebase, meaning what belongs to cli and what belongs to the image running in the linux container. however it is actually not a bad thing, since it forces me to go through the entire codebase carefully.

hope the above makes kinda sense :)

tantanchen:
I was able to get further using windows powershell
![image](https://user-images.githubusercontent.com/14248948/42851819-cc4a713a-89fb-11e8-8723-2aa18972c6b2.png)

NicolasMahe:
Yes, I also think we should only works with Linux container.
It will be very complicated to force all services to run perfectly on both windows and linux container. Let's just go with Linux container as standard.

But definitely, the CLI should run on Windows.

[...]

PR #296, opened 2018-07-22
fix-windows-compatibility

pts-solutions:
this PR resolves issues #273

the changes are still a little rough around the edges, but i think it is a good basis to start some discussion.

the changes are rather aggressive, as i basically removed all code working with unix pipes and switched to TCP. i consider this a good move, as we wanted to be able to run the server on a different machine or even go completely server-less, and to achieve this i believe that TCP is the way to go. actually windows also has something like unix pipes, it is called npipes, but they are totally incompatible.

i have also removed some directories, environment variables, mounts etc. that didn't make sense to me and that seemed unused. after all the changes done i tested everything with a small service

feel free to leave your comments


pts-solutions:
sorry that was too quick, created a pull request to the beginning of time :)

PR #297, opened 2018-07-22
fixed windows compatibility

pts-solutions:
this PR will address issue #273 

it is still a little rough around the edges, but i think good enough to start a little discussion on the changes made.

i have quite aggressively removed anything that is using unix pipes, except for docker.sock that needs to stay in place so the container can access the host's docker daemon. windows does actually also have something like unix pipes, called npipes, but they are not compatible. i also believe that since we have plan, at least that is what i understood, to have mesg-core run on a different machine or even go server-less, going all TCP seemed the right approach here,  

i also removed mounts, env variables and directories that seemed redundant and/or not used, even in sample services/apps, including fixing variable <-> package name clashes, as well as redundant variable declarations.

i have tested everything with a small service and all seems to be working well.

please feel free to comment

image

image

### DIFF
--- a/api/core/execute.go
+++ b/api/core/execute.go
@@ -2,10 +2,9 @@ package core
 
 import (
 	"encoding/json"
+	"context"
 
 	"github.com/mesg-foundation/core/database/services"
-
-	"context"
 	"github.com/mesg-foundation/core/execution"
 )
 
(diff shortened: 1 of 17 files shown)


pts-solutions:
i will close this PR first in favor of splitting it up into smaller more manageable pieces.

PR #312, opened 2018-07-31
Clean up code

krhubert:


api/service/submit_result.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/mesg-foundation/core/execution"
 )
 
-// SubmitResult of an execution
+// SubmitResult subbmits results of an execution.
```
ilgooz: // SubmitResult submits results of an execution.

pubsub/pubsub_test.go
```diff
@@ -6,11 +6,9 @@ import (
 	"github.com/stvp/assert"
 )
 
-type MessageStructTest struct{}
-
 func TestPublish(t *testing.T) {
 	key := "TestPublish"
-	data := MessageStructTest{}
```
ilgooz: @antho1404 @krhubert I think both version of this line is error prone. We should use a struct type with a field and set a value to it to do a proper assertion on struct equality.

_"Struct values are comparable if all their fields are comparable. Two struct values are equal if their corresponding non-blank fields are equal."_

pubsub/pubsub_test.go
```diff
@@ -20,7 +18,7 @@ func TestPublish(t *testing.T) {
 
 func TestPublishMultipleListeners(t *testing.T) {
 	key := "TestPublishMultipleListeners"
-	data := MessageStructTest{}
```
ilgooz: please also fix this

service/start.go
```diff
@@ -14,67 +15,66 @@ import (
...
 		go func(dep *DependencyFromService, i int) {
-			defer wg.Done()
 			serviceID, errStart := dep.Start(networkID)
 			mutex.Lock()
-			defer mutex.Unlock()
 			serviceIDs[i] = serviceID
 			if errStart != nil && err == nil {
 				err = errStart
 			}
+			mutex.Unlock()
```
antho1404: I'm curious on this one, why removing the `defer` the end result will be the same because anyway we are doing it in the end of the function. Is it for optimization or is their any other reasons ?

pubsub/pubsub_test.go
NicolasMahe: 👍

NicolasMahe: Done in https://github.com/mesg-foundation/core/pull/321/files

pubsub/pubsub_test.go
NicolasMahe: Done in https://github.com/mesg-foundation/core/pull/321/files

database/services/save.go
```diff
@@ -9,14 +9,12 @@ import (
...
 	db, err := open()
 	defer close()
 	if err != nil {
-		return
+		return "", err
 	}
-	hash = service.Hash()
-	err = db.Put([]byte(hash), bytes, nil)
-	return
+	return service.Hash(), db.Put([]byte(hash), bytes, nil)
```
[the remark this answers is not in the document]
NicolasMahe: Fixed in https://github.com/mesg-foundation/core/pull/324

NicolasMahe:
I didn't merge this PR but as I merge other branch that were partially based on this PR I think GitHub considered this PR merged...

PR #317, opened 2018-08-03
Return an error when trying to execute a task on a non running service

### LINKED ISSUE #307: Throw error when execute task on stopped service
## Problem

When using the command `service execute` on a service that is stopped, nothing happen / show up.

## Solution

The core should throw an error when trying to execute a task on a service that is not running

antho1404:
When calling the `executeTask` API, if the service was not running the system was adding the task in the queue but this task would have no chances to be executed. 
Now the API will return an error if the service is not running

fixes #307


api/core/error.go
```diff
@@ -0,0 +1,10 @@
+package core
+
+// NotRunningServiceError is an error when a service is not running
```
krhubert: Small change, missing dot at the end.

api/core/execute.go
```diff
@@ -6,25 +6,50 @@ import (
...
+	err := json.Unmarshal([]byte(request.InputData), &inputs)
+	return inputs, err
+}
+
+func execute(srv *service.Service, key string, inputs map[string]interface{}) (executionID string, err error) {
+	exc, err := execution.Create(srv, key, inputs)
+	if err != nil {
+		return "", err
+	}
+	err = exc.Execute()
```
ilgooz: please add this line directly to return

ilgooz:
lgtm

api/core/execute.go
```diff
@@ -6,25 +6,50 @@ import (
...
-
-	var inputs map[string]interface{}
-	if err := json.Unmarshal([]byte(request.InputData), &inputs); err != nil {
+	inputs, err := getData(request)
+	if err != nil {
 		return nil, err
 	}
-	execution, err := execution.Create(&service, request.TaskKey, inputs)
-	if err != nil {
+	if err := checkService(&srv); err != nil {
```
ilgooz: checkService() before getData()

ilgooz:
on CI:
```
--- FAIL: TestExecuteWithInvalidJSON (0.02s)
	assert.go:20: /go/src/github.com/mesg-foundation/core/api/core/execute_test.go:57
	assert.go:29: 💩  "Service v1_f7a73ba457db8de2f187635fe8bbdb47 is not running" != "unexpected end of JSON input"
```

antho1404:
Tests are ok now, I moved the `checkService` and renamed it into `checkServiceStatus` to be more precise. This check can be done either before or after the `getData`, I chose to put it after just for optimizations, `checkServiceStatus` make an api call to docker so if there is an error with the data at least we save one call

ilgooz:
@antho1404 makes sense, we should always validate data before an external call

PR #320, opened 2018-08-03
Krhubert/service

NicolasMahe:


NicolasMahe:
I approve this PR

NicolasMahe:
The test fails for this PR: https://circleci.com/gh/mesg-foundation/core/1772?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

service/importer/import.go
```diff
@@ -2,30 +2,30 @@ package importer
...
-		return
+		return nil, &ValidationError{}
 	}
 	data, err := readServiceFile(path)
 	if err != nil {
-		return
+		return nil, err
+	}
+	var importedService service.Service
+	if err := yaml.UnmarshalStrict(data, importedService); err != nil {
```
antho1404: can compact and also i'm not sure this works, in the first version we ad a pointer, in this case we don't and the unmarshal might have some problem to write data

```
return &importedService, yaml.UnmarshalStrict(data, &importedService)
```

service/start.go
```diff
@@ -13,68 +14,67 @@ import (
...
 		go func(dep *DependencyFromService, i int) {
-			defer wg.Done()
 			serviceID, errStart := dep.Start(networkID)
 			mutex.Lock()
-			defer mutex.Unlock()
 			serviceIDs[i] = serviceID
 			if errStart != nil && err == nil {
 				err = errStart
 			}
+			mutex.Unlock()
```
antho1404: I added a comment on the original PR. Is there a reason not to use the defer ?
Except there is a good reason I would like to keep the unlock and done as soon as possible because we might have some deadlocks if one day we add some returns in between and forget about that

service/status.go
```diff
@@ -33,27 +32,29 @@ func (service *Service) Status() (status StatusType, err error) {
...
+	return container.ServiceStatus(dependency.namespace())
 }
 
-// ListRunning all the running services
+// ListRunning returns all the running services.2
 // TODO: should move to another file
-func ListRunning() (res []string, err error) {
+func ListRunning() ([]string, error) {
 	services, err := container.ListServices("mesg.hash")
+	if err != nil {
```
antho1404: 👍

service/importer/import.go
antho1404: yes issue with not passing the pointer, fixed it

service/start.go
krhubert: My motivation was to get rid of defer call in simple function with zero/one return statement and I find it more readable, feel free to change it.

NicolasMahe: I'm putting back the defers

NicolasMahe:
@mesg-foundation/core please review.

service/start.go
```diff
@@ -13,27 +14,28 @@ import (
 	"github.com/spf13/viper"
 )
 
-// Start a service
-func (service *Service) Start() (serviceIDs []string, err error) {
+// Start starts the service.
+func (service *Service) Start() ([]string, error) {
```
ilgooz: it's a good practice using named returns on Public methods for **basic** types. Otherwise returned values doesn't make sense to dev and there is no documentation about it.

service/start.go
```diff
@@ -13,27 +14,28 @@ import (
...
 	if err != nil || status == RUNNING {
-		return //TODO: if the service is already running, serviceIDs should be returned.
+		return nil, err //TODO: if the service is already running, serviceIDs should be returned.
 	}
 	// If there is one but not all services running stop to restart all
 	if status == PARTIAL {
-		err = service.StopDependencies()
-		if err != nil {
-			return
+		if err = service.StopDependencies(); err != nil {
```
ilgooz: use `err :=` syntax instead

service/start.go
NicolasMahe: Done

service/start.go
NicolasMahe: Done

service/start.go
```diff
@@ -52,29 +54,27 @@ func (service *Service) Start() (serviceIDs []string, err error) {
...
 		service.Stop()
 	}
-	return
+	return serviceIDs, nil
 }
 
-// Start will start a dependency container
-func (dependency *DependencyFromService) Start(networkID string) (containerServiceID string, err error) {
+// Start starts a dependency container.
+func (dependency *DependencyFromService) Start(networkID string) (string, error) {
```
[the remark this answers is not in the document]
NicolasMahe: @mesg-foundation/core can anyone do this modifications today please?
So tomorrow morning I can merge

ilgooz:
@NicolasMahe can you please check the failed CI test before a merge?

NicolasMahe:
I restart the test and it pass. But it's the first time I see this error. I will investigate more.

NicolasMahe:
@antho1404 can you check and merge please?

PR #324, opened 2018-08-03
Krhubert/database

NicolasMahe:


NicolasMahe:
I approve this PR

database/services/save.go
```diff
@@ -5,18 +5,17 @@ import (
...
-		return
+		return "", err
 	}
 	db, err := open()
 	defer close()
 	if err != nil {
-		return
+		return "", err
 	}
 	hash = service.Hash()
```
antho1404: to stay consistent with the result of the refactoring, I would not use the named output and assign a new variable here

MarinX: ...but the whole package consists of named outputs.
Should we refactor service package to remove named outputs?

antho1404: hum no not really, let's do things step by step, i'm ok with that anyway

PR #326, opened 2018-08-03
Krhubert/container

NicolasMahe:


container/network.go
```diff
@@ -7,23 +7,19 @@ import (
 	docker "github.com/docker/docker/client"
 )
 
-// CreateNetwork creates a Docker Network with a namespace
-func CreateNetwork(namespace []string) (networkID string, err error) {
+// CreateNetwork creates a Docker Network with a namespace.
+func CreateNetwork(namespace []string) (string, error) {
```
ilgooz: return with networkID name for docs

container/build.go
```diff
@@ -18,68 +19,64 @@ type BuildResponse struct {
 	Error  string `json:"error"`
 }
 
-// Build a docker image
-func Build(path string) (tag string, err error) {
+// Build builds a docker image.
+func Build(path string) (string, error) {
```
ilgooz: use named returns for docs

container/network.go
```diff
@@ -33,36 +29,29 @@ func CreateNetwork(namespace []string) (networkID string, err error) {
...
+// DeleteNetwork deletes a Docker Network associated with a namespace.
+func DeleteNetwork(namespace []string) error {
 	network, err := FindNetwork(namespace)
-	if docker.IsErrNotFound(err) {
-		err = nil
-		return
-	}
-	if err != nil {
-		return
+	if err != nil && !docker.IsErrNotFound(err) {
```
antho1404: this is not the same behavior than before
before: if the network is not found we exit with no error
now: if the network is not found we continue and probably have a crash on the `network.ID`

container/service.go
```diff
@@ -10,88 +10,78 @@ import (
...
-func ServiceStatus(namespace []string) (status StatusType, err error) {
-	status = STOPPED
-	_, err = FindService(namespace)
-	if docker.IsErrNotFound(err) {
-		err = nil
-		return
+// ServiceStatus returns the status of the Docker Swarm Servicer.
+func ServiceStatus(namespace []string) (StatusType, error) {
+	if _, err := FindService(namespace); !docker.IsErrNotFound(err) {
+		return STOPPED, err
```
antho1404: not the same behavior too. Now if the service is not found we return that it's running which doesn't make sense

container/shared_network.go
```diff
@@ -27,25 +24,23 @@ func createSharedNetworkIfNeeded(client *docker.Client) (err error) {
...
 }
 
-// SharedNetworkID returns the id of the shared network
-func SharedNetworkID() (networkID string, err error) {
+// SharedNetworkID returns the ID of the shared network.
+func SharedNetworkID() (string, error) {
 	client, err := Client()
 	if err != nil {
-		return
+		return "", nil
```
antho1404: we need to return the error

container/shared_network.go
```diff
@@ -27,25 +24,23 @@ func createSharedNetworkIfNeeded(client *docker.Client) (err error) {
...
 	}
 	network, err := sharedNetwork(client)
 	if err != nil {
-		return
+		return "", nil
```
antho1404: here too

antho1404:
@ilgooz this one was not ready to merge, there is few regressions in the code

NicolasMahe:
I confirm this PR has been reverted. Let's continue the discussion on https://github.com/mesg-foundation/core/pull/336

container/network.go
NicolasMahe: No, the behavior is the same as before:
The tests pass on this function

container/shared_network.go
NicolasMahe: Fixed

container/service.go
NicolasMahe: Fixed

container/shared_network.go
NicolasMahe: Fixed

container/network.go
NicolasMahe: Fixed 👍

PR #329, opened 2018-08-03
Krhubert/cli

NicolasMahe:


cli/main.go
```diff
@@ -5,12 +5,12 @@ import (
 	"github.com/mesg-foundation/core/cmd/utils"
 )
 
-// version of this release. Will be replaced automatically when compiling in CI
-var version = "vX.X.X"
+// Version of this release. Will be replaced automatically when compiling in CI
+var Version = "vX.X.X"
```
antho1404: why make this variable accessible ?

ilgooz: I think there is no rules about public variables when using ldflag. @krhubert might converted version to public maybe to document that Version will be accessed and set from outside, in this case by ldflag.

NicolasMahe: Let's not forget to edit the build script:
https://github.com/mesg-foundation/core/blob/dev/scripts/build-cli.sh#L7

NicolasMahe:
@mesg-foundation/core a modification should be test or done on: (see comment of this PR)
https://github.com/mesg-foundation/core/blob/dev/scripts/build-cli.sh#L7

NicolasMahe:
Closing this PR in favor of https://github.com/mesg-foundation/core/pull/343/files

PR #336, opened 2018-08-03
Krhubert/container

ilgooz:
is a part of https://github.com/mesg-foundation/core/pull/326


NicolasMahe:
@antho1404 please review

container/service.go
```diff
@@ -10,88 +10,82 @@ import (
...
-	_, err = FindService(namespace)
+// ServiceStatus returns the status of the Docker Swarm Servicer.
+func ServiceStatus(namespace []string) (StatusType, error) {
+	_, err := FindService(namespace)
 	if docker.IsErrNotFound(err) {
-		err = nil
-		return
+		return STOPPED, nil
+	}
+	if err != nil {
```
antho1404: good, seems like we had a bug on this one, in case of error, we would say that the service is RUNNING, not a big problem because we will check always the error but good to have this catch

Issue #283, opened 2018-07-16
convert to unit tests


ilgooz:
We are heavily depending on integration tests in docker client related tests, gRPC calls and tests that run [gitClone](https://github.com/mesg-foundation/core/blob/dev/cmd/service/utils.go#L66). 
* Docker client supports custom dialers [here](https://godoc.org/github.com/docker/docker/client#WithDialer). We should use an in memory net.Conn as a return value of dialer and validate our business logic by investigating what kinda api calls been produced. Or simply use [WithHTTPClient](https://godoc.org/github.com/docker/docker/client#WithHTTPClient) like this: https://github.com/moby/moby/blob/master/client/client_mock_test.go#L20
* gRPC calls can be easily mocked by satisfying it's interfaces. e.g.: https://github.com/ilgooz/mesg-go/blob/master/service_test.go#L183-L219
* we can add pre downloaded repositories to our test folder to test gitClone behavior.

ilgooz:
includes https://github.com/mesg-foundation/core/issues/274

ilgooz:
Update:
use [buffconn](https://godoc.org/google.golang.org/grpc/test/bufconn) to create a mock server for gRPC

ilgooz:
docker client doesn't provide a mocking server nor interface at this moment. this possibly leaves us directly checking generated api calls like described in the first message of this thread.

see https://github.com/moby/moby/issues/30454

ilgooz:
I split git and gRPC related tests to other issues. See https://github.com/mesg-foundation/core/issues/339 and https://github.com/mesg-foundation/core/issues/310.

Issue #337, opened 2018-08-04
Proposal/refactor: Drop package level funcs in favor of Newable Go types and Improve package APIs


ilgooz:
Current implementation of core code should be refactored with more idiomatic Go code. Changes will make unit testing easier as a side effect which is the second purpose of this proposal.

We need to drop package level functions and create Go types for most of the core packages to attach them as their methods. Converting package level functions to type methods gives a nice isolation, grouped APIs and overall look. It also helps with testing to pass mocking interface/socket in the initialization time with `New(options ...Option)`.

_This requires package modifications on `container`, `service` and `daemon`. 
This requires code changes on `importer`, `core proto`, `api/core`, `api/service` and `cmd/service`.
(we can separate this as 3 PRs -or more- like `feature/container`, `feature/service`, `feature/daemon`)_

_There are some other packages that needs be refactored too like: `database/services`, `execution` etc. But I'll leave them for now and target  these most important three: `container`, `service` and `daemon`._

# Container
There is no major changes on container package except all the package level functions is now a method of `Container` type.

Please check this on going changes on container: https://github.com/mesg-foundation/core/tree/feature/container

# Service
### Issues
* 1- Consumers of `service` package directly uses `database/services` package to get a `Service`. This functionality should be provided by service package itself. database/services interactions should be made internally within service package. This is also good for separation of logic in the code and unit testing. _(See new Manager API related to this)_

* 2- Manually craft Service, Event, Task, Output, Parameter, Dependency types. I really don't like extending types generated by protocol buffers. It's not possible to add private fields to Service because of that. I need to extend Service type with some private fields to set the container client interface and other information needed by the new version of service API.

### Package level functions
Some of the public/private APIs of service package:
* `service.ParseYAML(data []byte) (Definition, error)` > error will be filled validation errors if exists
* `service.ParseDockerfile(data []byte) (Dockerfile, error)` > error will be filled validation errors if exists
* `service.NewManager(options ...Options) (*Manager, error)`
* `service.LevelDBOption(path string) Option`
* `service.ContainerDockerClientOption(client docker.CommonAPIClient) Option` > used to mock container package's Docker Client.

### Introduce Manager type
Some of the public/private APIs of Manager:
 * `Manager.GetService(id string) (*Service, error)`
 * `Manager.CreateService(definition Definition) (*Service, error)`
 * `Manager.CreateServiceFromYAML(data []byte) (*Service, error)`
 * `Manager.DeleteService(id string) (*Service, error)`
[...]

Issue #268, opened 2018-06-29
Investigate test with address already in use


antho1404:
Sometime when running the tests we have the error 

`Address already in use, failed to allocate gateway (10.0.1.1): Address already in use`

This shouldn't happen and need to be investigated and fixed to always have reliable builds

ilgooz:
This can be related with a bug in docker or docker client:
* https://github.com/moby/moby/issues/35204
* https://github.com/moby/moby/issues/35834

I investigated this issue for an hour yesterday but I didn't come up with a solid fix yet.  Putting a 3 second sleep above the line of [here](https://github.com/mesg-foundation/core/blob/dev/container/network.go#L56) or [here](https://github.com/mesg-foundation/core/blob/dev/container/network.go#L38) solves this problem somehow. Maybe there is something async process going on when creating a network on docker's side or there is some name conflicts in network-service relationship.

But converting to [unit tests](https://github.com/mesg-foundation/core/issues/283) should definitely solve this if we don't have a business logic mistake in our side.

NicolasMahe:
This issue didn't occured for a long time. It should have been fixed with the fix on listening docker's event for network deletion.

PR #340, opened 2018-08-05
Feature/container: new Container type & unit tests

ilgooz:
closes https://github.com/mesg-foundation/core/issues/283.
related with https://github.com/mesg-foundation/core/issues/337, https://github.com/mesg-foundation/core/issues/268.

* It would be much simpler to review this PR by reviewing each commit individually.
* Run tests with `integration` tag to enable integration tests: `go test -tags=integration`
* new `dockertest` package provided within this PR is used while unit testing of container package and will be used by its consumer packages to mock Docker API.
* Container type is introduced.
* Minor improvements done on the container package.
* All contexts replaced with WithTimeout contexts.


NicolasMahe:
I will review first thing tomorrow morning ;)

@ilgooz If you want to start another dev from this branch before it's merged, you can start a new branch from your last commit ;)

container/dockertest/client.go
```diff
@@ -0,0 +1,268 @@
...
+	imageBuild            chan imageBuildResponse
+	networkInspect        chan networkInspectResponse
+	networkCreate         chan networkCreateResponse
+	taskList              chan taskListResponse
+	serviceCreate         chan serviceCreateResponse
+	serviceList           chan serviceListResponse
+	serviceInspectWithRaw chan serviceInspectWithRawResponse
+	serviceRemove         chan serviceRemoveResponse
+	serviceLogs           chan serviceLogsResponse
+	contanerInspect       chan containerInspectResponse
```
NicolasMahe: typo on `contanerInspect`. Should be `containerInspect`.
This typo is also at 3 other places.

container/client_test.go
```diff
@@ -1,44 +1,5 @@
...
-	assert.NotEqual(t, client, client2)
-}
-
-func TestCreateSwarmIfNeeded(t *testing.T) {
-	client, _ := createClient()
-	err := createSwarmIfNeeded(client)
-	assert.Nil(t, err)
-}
-
 // TODO: this tests break other tests on my machine
```
NicolasMahe: Let's remove completely this file. I don't think we really need integration test on the `createSwarm` functionality. 
@antho1404 what do you think?

container/container.go
```diff
@@ -2,19 +2,91 @@ package container
...
 	if err != nil {
-		return types.ContainerJSON{}, err
+		return err
+	}
+	if info.Swarm.NodeID != "" {
+		return nil
 	}
-	containers, err := client.ContainerList(context.Background(), types.ContainerListOptions{
+	ctx, cancel = context.WithTimeout(context.Background(), c.callTimeout)
+	defer cancel()
```
NicolasMahe: is it ok to assign `cancel` multiple time and call it using `defer`?
Is `defer` will really execute the 2 cancel or the last cancel 2 times?

container/client_test.go
antho1404: Why would you want to remove it ? if the test is ok let's keep it, after I agree swarm tests are not  absolutely necessary but I think it's still really good to keep these tests except if there is a good reason to remove them

NicolasMahe: https://github.com/mesg-foundation/core/blob/e0e344e6449caabd857f0eebaa7eb81e99236309/container/client_test.go#L3-L25

cmd/service/init.go
```diff
@@ -43,6 +48,13 @@ func init() {
 	Init.Flags().StringP("description", "d", "", "Description")
 	Init.Flags().BoolP("current", "c", false, "Create the service in the current path")
 	Init.Flags().StringP("template", "t", "", "Specify the template URL to use")
+
+	// TODO(ilgooz): remove this after service package made Newable.
+	c, err := container.New()
+	if err != nil {
+		log.Fatal(err)
+	}
+	defaultContainer = c
```
antho1404: I think this is to initialize a `defaultContainer` in the CLI but this file is the file for the command `mesg-core service init` so that might not make sense to put it here but maybe more in the `utils` where it's used

Also if it's only used one time in the `service/utils#buildDockerImage` why not initialize it in this function directly

container/client_test.go
antho1404: my bad you're right

container/dockertest/testing.go
```diff
@@ -0,0 +1,172 @@
...
+	t.client.responses.serviceCreate <- serviceCreateResponse{response, err}
+}
+
+// ProvideServiceRemove sets fake return values for the next call to *Client.ServiceRemove.
+func (t *Testing) ProvideServiceRemove(err error) {
+	t.client.responses.serviceRemove <- serviceRemoveResponse{err}
+}
+
+// LastContainerList returns a channel that receives call arguments of last *Client.ContainerList call.
+func (t *Testing) LastContainerList() <-chan ContainerListRequest {
```
antho1404: I don't like much the naming with `Last` but i'm definitely not against it

daemon/init.go
```diff
@@ -0,0 +1,18 @@
+// TODO(ilgooz): remove this file after daemon package made Newable.
+package daemon
+
+import (
+	"log"
+
+	"github.com/mesg-foundation/core/container"
+)
+
+var defaultContainer *container.Container
```
antho1404: Isn't it better to create an instance every time we need it instead of creating it an `init` function ?
It would be good to avoid `init` functions that save a state.
This container is totally stateless so we can re-create it all the time and it will not change anything except some memory/cpu optimization but that are negligible for these features

service/init.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"log"
+
+	"github.com/mesg-foundation/core/container"
+)
+
+var defaultContainer *container.Container
+
+func init() {
+	c, err := container.New()
```
antho1404: same remark here, do we really need a state and an init function ?

container/container.go
ilgooz: yes it is. please see: https://play.golang.org/p/UeNDuSXbZz7

container/dockertest/client.go
ilgooz: done

container/client_test.go
ilgooz: I'm removing `core/container/client_test.go` file.

cmd/service/init.go
ilgooz: We are going to remove these lines after service package refactored. `*Container` will be initialized once and stored in the `*Service` struct. I don't mind where to put this temporary code, want me to move to utils for now?

daemon/init.go
ilgooz: After daemon refactored, `*Container` will be initialized only once and stored in the `*Daemon` struct. Until the refactor we have to initialize a Container type and store it in the package level to be able to use it. I won't recommend _-also don't like-_ initializing Container every time it's needed because it makes additional Docker API calls in the initialization time. Please see [here](https://github.com/mesg-foundation/core/blob/feature/container/container/container.go#L40-L44).

service/init.go
ilgooz: Yes, until we refactor service. This is just a temporary solution.

cmd/service/init.go
antho1404: yes that would be good, I'm afraid to forget it here because it's not really related to this command

daemon/init.go
antho1404: ok let's keep it like that for now, and for the initialization of the container every time, you're right with the docker api calls even if we could add a singleton on that or some stuff like that but we will think about this kind of optimization later
So this is ok now :)

ilgooz: I directly changed the corresponding commit from the history as you liked. This way we'll not see an irrelevant code in the init.go file.

Issue #280, opened 2018-07-13
need vendoring


ilgooz:
could be useful:
* docker client sample:
https://github.com/moby/moby/blob/master/VENDORING.md
https://github.com/moby/moby/blob/master/vendor.conf

NicolasMahe:
Fixing the version of the dependencies of this project is a good idea but which solutions to choose? There is no standard way in Go to do it.

https://blog.golang.org/versioning-proposal
https://github.com/golang/go/issues/24301

ilgooz:
* Docker has it's own vendor definition file called [vendor.conf ](https://github.com/moby/moby/blob/master/vendor.conf). Please check [here](https://github.com/moby/moby/blob/76e43532f60355220619b5ff808c0d2cd4bd624b/hack/README.md#vendor-vendorsh). So using [vndr](https://github.com/LK4D4/vndr) is an option.
* Go offical has an experimental [dep](https://github.com/golang/dep) tool. It both has a conf yml and lock yml.

There are other popular dep tools like [govendor](https://github.com/kardianos/govendor) but I think we should stick with the official one or maybe vndr. Without the matter of which tool we use we are going to have a [standard](https://golang.org/cmd/go/#hdr-Vendor_Directories) vendor/ dir in our root.

Not directly related but also worth to see https://research.swtch.com/vgo-tour

antho1404:
https://glide.sh/ can be a solution to explore

krhubert:
Golang has dep as it's (semi) official tool. I've used the tool as it would be merged with go standard tools

ilgooz:
lets consider using an automation tools to update vendor/ folder. it'd be nice to automatically update vendor/ folder depending on the changes in the vendor.conf file. this way we don't forgot about updating vendor/ folder and prevent corrupted vendor files to merge. please see:  https://dependabot.com/

krhubert:
As @ilgooz suggested there is a  [platform](https://dependabot.com/) for automated dependency updates. It uses dep under the hood and creates PR with dependencies update.

krhubert:
https://github.com/mesg-foundation/core/pull/342

ilgooz:
* We can add a CI task to create automated PR's to update vendor/ for the latest version of dep conf file. This task should run in daily basis or after a new code merged to the dev branch.
[...]

PR #342, opened 2018-08-06
Vendor

krhubert:


NicolasMahe:
@krhubert can you please reference the linked issue

antho1404:
do we really need to save all the packages if we have the `Gopkg.lock`, we have all the versions so there is certainly a way to fetch all the packages with these current versions/hashes

ilgooz:
@antho1404 you are absolutely right! I like it. I don't like having a huge vendor/ in the middle of the repo too. But one problem is having vendor/ within the repo has an advantage that we don't get effected by deleted dependencies from Github or other git hosts.

ilgooz:
after changes made we can merge this PR and only have Gopkg.toml, Gopkg.lock in the project for now. https://github.com/mesg-foundation/core/issues/280 can be closed after CI task is created

NicolasMahe:
@ilgooz can you create a new issue with what your were saying about the CI

PR #343, opened 2018-08-06
Output version in log when the daemon is starting

krhubert:


NicolasMahe:
Very cool! I didn't know we could inject a value in a variable of another package than main:
```
```

NicolasMahe:
@krhubert please, reference the issue in the description of the pull request.
You can write:
```
close #issue's-number
```
and it will close the issue when the PR is merged ;)

.circleci/config.yml
```diff
@@ -56,7 +56,7 @@ jobs:
     steps:
       - checkout
       - run: docker pull mesg/core:latest
-      - run: docker build -t mesg/core:$CIRCLE_TAG -t mesg/core:latest .
+      - run: docker build -t mesg/core:$CIRCLE_TAG -t mesg/core:latest --build-arg version=`echo $CIRCLE_SHA1 | cut -c1-7` .
```
NicolasMahe: In this step (`publish_docker_prod`) we want to use the version of the tag (`$CIRCLE_TAG`)

NicolasMahe: But you forget to add the version based on the sha1 in the step `publish_docker_dev`:
https://github.com/mesg-foundation/core/blob/a880f34c273793004dab254061b888cdeb41ff42/.circleci/config.yml#L45-L52

You have to add the version based on the sha1 with a nice prefix like:
https://github.com/mesg-foundation/core/blob/a880f34c273793004dab254061b888cdeb41ff42/.circleci/config.yml#L68

krhubert: I didn't put version in docker_dev because I didn't know what version should we use there,

About docker prod, this is version for core binary and I took it from cli build. Do we want diffrent version for core & cli?

NicolasMahe: Look at the step to build the cli.
For dev build, we use as version: `dev build `echo $CIRCLE_SHA1 | cut -c1-7``
For prod build, we use: `$CIRCLE_TAG`

krhubert: ok

.circleci/config.yml
```diff
@@ -56,7 +56,7 @@ jobs:
     steps:
       - checkout
       - run: docker pull mesg/core:latest
-      - run: docker build -t mesg/core:$CIRCLE_TAG -t mesg/core:latest .
+      - run: docker build -t mesg/core:$CIRCLE_TAG -t mesg/core:latest --build-arg version=$CIRCLE_TAG
```
NicolasMahe: I think you forgot the `.` at the end of the command

Issue #256, opened 2018-06-25
Add data to the service execute with no file


NicolasMahe:
We should have a way to pass a data to the command service execute.

Right now, we have to put the JSON in a file and pass the path with the flag `json`.

Related to https://github.com/mesg-foundation/core/issues/255

antho1404:
**Warning**: This issue depends on https://github.com/mesg-foundation/core/issues/288, make sure that the https://github.com/mesg-foundation/core/issues/288 issue is fixed before starting this one

antho1404:
I updated this issue and talked about data. I really think JSON is not a good way to pass data through the command line we will have to put something like that
```
mesg-core service execute --task=taskX --data="{\"foo\"=\"bar\"}"
```

This is terrible and anything other than that will be invalid so json is definitely not good for UX if it's not in a file.

We can find of some ways to do that that would be nicer for the developer.

Maybe something like

```
mesg-core service execute --task=taskX --foo=bar
```

But this has a problem if one of the input is the same name than a flag of the command. We could fix this by prefixing with `data` like:
```
mesg-core service execute --task=taskX --data-foo=bar
```

also with that everything will be a string and it will add a lot of complexity on the cli to typecast to the right type which is annoying because the api is already doing this check

I think it's good to be able to add another way than a json file for the data but I would actually go maybe with proposing a yaml file for example and then use env variables if we really want to have this on the console like

```
DATA_FOO=bar mesg-core service execute --task=taskX --yaml=test.yml
```

with the following yaml:
```yml
foo: ${DATA_FOO}
# even prefixing the type if needed
foo: !!str ${DATA_FOO}
```

NicolasMahe:
I'm of for the yaml file.

For passing data as flag, I don't think it's a good idea. You make a good point.

What about asking questions for every input of the tasks? The CLI could read the service definition and create those questions.

antho1404:
The question is really what we want to provide for the user, if it's a quick and easy way to set the data I really don't think that asking in the console will be quick nor easy.

If this is the real thing that we try to fix maybe we should think about generating these files for the user or something like that.

Otherwise there is another solution similar to curl it will be to add something like that
```
mesg-core service execute --task taskX --data param1=hello --data param2=bar
```

this is the way you can give data for the curl command

NicolasMahe:
I really like it!

Maybe even?
```
mesg-core service execute --task taskX --data param1=hello param2=bar
```

antho1404:
[...]

PR #344, opened 2018-08-06
Add data to the service execute with no file

krhubert:
related to https://github.com/mesg-foundation/core/issues/256


NicolasMahe:
@krhubert can you please reference the linked issue in the description at least

cmd/service/execute.go
```diff
@@ -96,14 +101,28 @@ func getTaskKey(cmd *cobra.Command, serviceID string) string {
...
-		}
+func getData(cmd *cobra.Command) (string, error) {
+	data := cmd.Flag("data").Value.String()
+	jsonStr := cmd.Flag("json").Value.String()
+	jsonFile := cmd.Flag("json-file").Value.String()
+
+	if (data != "" && jsonStr != "") ||
+		(data != "" && jsonFile != "") ||
+		(jsonStr != "" && jsonFile != "") {
+		return "", errors.New("data, json and json-file flags are  mutually exclusive")
```
NicolasMahe: Can you write something more user friendly like: "You can use only of one of the following flags: 'data', 'json' and 'json-file'."

antho1404:
This one will definitely need some update on the documentation and changelog and also I'm not really sure about it first we already have a `json` that reads a file and not it does something different. Also I don't see the point of having a json directly in the command (not in the file) and having a data.

cmd/service/execute.go
krhubert: Sure, done.

As xpflag is eXtension to [spf13/pflag](https://github.com/spf13/pflag) that contains custom flags (implements [pflag.Value interface](https://godoc.org/github.com/spf13/pflag#Value)) and I put it in different package.

krhubert:
@antho1404 For sure, I've kept 3 options for further discussion (which of them should/shouldn't stay).
As for JSON in command, I find it frustrating when I wanted to execute some task I had to create a file with simple JSON, then edit the file in order to execute with different task data, etc...

It's nice to have json in cmd:
```
$ mesg-core service test --task execute --data '{"key":"1"}'
$ mesg-core service test --task execute --data '{"key":"2"}'
$ # do more 
```

antho1404:
I was actually against json as parameter because if using double quotes it looks like 
`"{\"foo\":\"bar\"}"` which is terrible to write but with single quotes it's way better. I'm just afraid for a user perspective they might do this mistake and spend hours to understand why their data are not good. Users might be junior developers so probably not that much comfortable with the terminal.

I like the data part but this will be interpreted as string and if the data needs to be boolean for example the core will reject the task so this is something to manage too

cmd/service/execute.go
```diff
@@ -11,11 +13,13 @@ import (
 	"github.com/mesg-foundation/core/api/core"
 	"github.com/mesg-foundation/core/cmd/utils"
 	"github.com/mesg-foundation/core/service"
+	"github.com/mesg-foundation/core/utils/xpflag"
 	"github.com/spf13/cobra"
 	survey "gopkg.in/AlecAivazis/survey.v1"
 )
 
 var executions []*core.ResultData
```
NicolasMahe: This is not used anymore.

NicolasMahe:
@mesg-foundation/core What about asking the value of the task's inputs by questions if no data has been passed by flags? It's not the most efficient way to pass data, but for a quick test of a task it's super convenient! No need to know the inputs before executing the command.

cmd/service/execute.go
krhubert: don't get it, what do you mean? `executions` wasn't touched

NicolasMahe:
To sum up the conversion on discord:

- [ ] Revert `--json` flag to accept a JSON file
- [ ] Keep flag `--data`. But be compatible with: `--data key1=value1 --data key2=value2`
- [ ] Cast every value passed with the `data` flag using the type defined in the definition of the service (https://docs.mesg.com/guide/service/service-file.html#service-file-2, https://github.com/mesg-foundation/service-ethereum/blob/master/mesg.yml)

antho1404:
I would add also to update the documentation
https://github.com/mesg-foundation/core/blob/fc75e9b86e16947a351f0202ea72a16c6216941f/docs/guide/service/test.md

service/cast.go
```diff
@@ -0,0 +1,49 @@
...
+	task, ok := s.Tasks[taskKey]
+	if !ok {
+		return nil, fmt.Errorf("Task %q not found", taskKey)
+	}
+
+	m := make(map[string]interface{}, len(taskData))
+	for key, value := range taskData {
+		inputType, ok := task.Inputs[key]
+		if !ok {
+			return nil, fmt.Errorf("Task %q has no input named %q", taskKey, key)
```
NicolasMahe: Please create a new type for this error in `service/error.go`

service/cast.go
```diff
@@ -0,0 +1,49 @@
...
+				m[key] = i
+			}
+		case "Boolean":
+			b, err := strconv.ParseBool(value)
+			if err != nil {
+				return nil, fmt.Errorf("Task %q - input %q is not a Boolean type", taskKey, key)
+			}
+			m[key] = b
+		case "Object":
+			return nil, fmt.Errorf("Task %q - input %q - Object type is not supported", taskKey, key)
```
NicolasMahe: Can we really not pass any Object with the CLI? Even a JSON stringified?

service/cast.go
krhubert: done

service/cast.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"fmt"
+	"strconv"
+)
+
+// Cast converts map[string]string to map[string]interface{} based on defined types in the service tasks map.
+func (s *Service) Cast(taskKey string, taskData map[string]string) (map[string]interface{}, error) {
+	task, ok := s.Tasks[taskKey]
+	if !ok {
+		return nil, fmt.Errorf("Task %q not found", taskKey)
```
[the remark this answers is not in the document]
krhubert: done

service/cast.go
krhubert: Ok, so I'll continue here instead of return an error

service/cast.go
```diff
@@ -0,0 +1,75 @@
...
+func castBoolean(value string) (interface{}, error) {
+	b, err := strconv.ParseBool(value)
+	if err != nil {
+		return nil, fmt.Errorf("input %q is not a Boolean type", value)
+	}
+	return b, nil
+}
+func castObject(value string) (interface{}, error) {
+	// do not cast object type
+	return nil, nil
```
antho1404: This is not returning any value so any service with object will fail.

We can add a basic support by at least returning the value or even better trying to parse it from string

PR #356, opened 2018-08-09
Change testing pkg

krhubert:
What do you think about replacing `github.com/stvp/assert` with `github.com/stretchr/testify/require`

It has the same API but with way more features:

- Zero asserts that i is the zero value for its type.
- test if dir/file exists
- HTTP testing (in the future)
- test if objects are equal
- test regexp
- test time duration
- test JSON as a string
- more ...

PR is done and only needs to merge.

@mesg-foundation/core What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Changed
 - (#282) Branch support added. You can now specify your branches with a `#branch` fragment at the end of your git url. E.g.: https://github.com/mesg-foundation/service-ethereum-erc20#websocket
 - (#299) Add more user friendly errors when failing to connect to the Core or Docker
+- (#356) Use github.com/stretchr/testify package
 
 #### Added
 - (#242) Add more details in command `mesg-core service validate`
--- a/api/client/client_test.go
+++ b/api/client/client_test.go
@@ -3,18 +3,18 @@ package client
 import (
 	"testing"
 
-	"github.com/stvp/assert"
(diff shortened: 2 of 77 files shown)


ilgooz:
I like the simplicity of assert package and its features looks enough for now while unit testing core. And I don't like to switching to a testing package that has tons of features because MESG is an open source project and we'll have lots of contributions from the community and I don't want devs to use so many different testing functions. It could be not easy to understand what each function does for everyone in the first glance. 

We should switch to something like this only when we feel that there is an absolute need for it. 
Also, I don't want to add extra complexity for http testing, using standard httptest package is enough to me.

krhubert:
> I like the simplicity of assert package and its features looks enough for now

Yes for now. This is my motivation - we have to think about feature use.

>  And I don't like to switching to a testing package that has tons of features because MESG is an open source project and we'll have lots of contributions from the community and I don't want devs to use so many different testing functions. 

This is why testify is great option. Based on stvp/assert starts nobody (besides us) is using this package. So you force developer to used new api and testify (>5k stars) is known in community. Based on that arguemnt we should stick with testify or std testing pkg.

> We should switch to something like this only when we feel that there is an absolute need for it.

Then it could be too late and you'll need even more code refactor.

> Also, I don't want to add extra complexity for http testing, using standard httptest package is enough to me.

If you look into source code, testify has time saving wrapers around httptest (you can still use httptest)

krhubert:
@NicolasMahe done

PR #358, opened 2018-08-09
fix goroutine leaks on stream handlers of api package

ilgooz:
There are goroutine leaks in the api handlers where gRPC streams used. When client disconnects or cancels, api handlers will keep running forever. This PR fixes this.

Btw, I don't understand why we compare chans with nil [here](https://github.com/mesg-foundation/core/blob/403153e5f7cbb3d8911814d064be9b5ced277a95/pubsub/pubsub.go#L19)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - (#303) Deprecate command `service test` in favor of `service dev` and `service execute`
 
 #### Fixed
+- (#358) Fix gorutinue leaks on api package handlers where gRPC streams are used. Handlers were blocking forever because of not respecting to context cancellation and stream.Send() errors.
 
 ## [v1.0.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 - (#303) Deprecate command `service test` in favor of `service dev` and `service execute`
 
 #### Fixed
-- (#358) Fix gorutinue leaks on api package handlers where gRPC streams are used. Handlers were blocking forever because of not respecting to context cancellation and stream.Send() errors.
+- (#358) Fix goroutine leaks on api package handlers where gRPC streams are used. Handlers now doesn't block forever by exiting on context cancellation and stream.Send() errors.
 
 ## [v1.0.0]
 
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -34,3 +34,15 @@ func Subscribe(channel string) chan Message {
 	mu.Unlock()
 	return listener
 }
+
+// Unsubscribe unsubscribes a subscription for listening channel.
+func Unsubscribe(channel string, subscription chan Message) {
+	mu.Lock()
+	defer mu.Unlock()
+	for i, s := range listeners[channel] {
+		if s == subscription {
+			listeners[channel] = append(listeners[channel][:i], listeners[channel][i+1:]...)
+			return
+		}
+	}
+}
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -37,13 +37,12 @@ func Subscribe(channel string) chan Message {
 func Unsubscribe(channel string, listener chan Message) {
 	mu.Lock()
 	defer mu.Unlock()
-	if len(listeners[channel]) == 1 {
-		listeners[channel] = nil
-		return
-	}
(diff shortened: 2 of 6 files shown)


antho1404:
It's ok for me, maybe you can add a line on the changelog and for the comparison with the nil, I guess it was in case the channel becomes nil when the subscriber unsubscribe but it's not really useful, I guess it was a useless safe test...

Also there is an issue on codeclimate, maybe we can extract all the verifications in another function to fix it.

Otherwise it's great 👍

pubsub/pubsub.go
```diff
@@ -15,32 +15,34 @@ type Message interface{}
...
 
-// Unsubscribe unsubscribes a subscription for listening channel.
-func Unsubscribe(channel string, subscription chan Message) {
+// Unsubscribe unsubscribes a listener from listening channel.
+func Unsubscribe(channel string, listener chan Message) {
 	mu.Lock()
 	defer mu.Unlock()
-	for i, s := range listeners[channel] {
-		if s == subscription {
+	if len(listeners[channel]) == 1 {
```
antho1404: I'm not sure about that one. What happen if for any reason (by design it shouldn't be possible but everything is possible) the list of `listeners[channel]` contains another `chan` than `listener`.

I would prefer to iterate over the `listeners` and if it's just about memory optimization we can still at the end of the loop, if the length of `listeners[channel]` is 0 then delete it

Is there any other reason why you did this condition here ?

pubsub/pubsub.go
```diff
@@ -15,22 +15,36 @@ type Message interface{}
...
 	listeners[channel] = append(listeners[channel], listener)
-	mu.Unlock()
 	return listener
 }
+
+// Unsubscribe unsubscribes a listener from listening channel.
+func Unsubscribe(channel string, listener chan Message) {
+	mu.Lock()
+	defer mu.Unlock()
+	if len(listeners[channel]) == 1 {
```
antho1404: I'm not sure about that one. What happen if for any reason (by design it shouldn't be possible but everything is possible) the list of `listeners[channel]` contains another `chan` than `listener`.

I would prefer to iterate over the `listeners` and if it's just about memory optimization we can still at the end of the loop, if the length of `listeners[channel]` is 0 then delete it

Is there any other reason why you did this condition here ?

ilgooz: just a memory optimization. if it contains more than 1 listener then program is going to skip the if statement and **a** listener will be removed. This will happen until there is only 1 one listener left in the slice and for that one execution will go through the if statement and slice will be nilled.

I think both ways are totally of equivalent each other. If you prefer the second way as a pattern, I don't mind to change.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - (#303) Deprecate command `service test` in favor of `service dev` and `service execute`
 
 #### Fixed
+- (#358) Fix gorutinue leaks on api package handlers where gRPC streams are used. Handlers were blocking forever because of not respecting to context cancellation and stream.Send() errors.
```
NicolasMahe: Typo on "gorutinue"

NicolasMahe: Can you improve "not respecting to context cancellation"

ilgooz: done

pubsub/pubsub.go
antho1404: I'm just afraid about some possible side effects like another part add a new chan in the `listeners[channel]` directly and then there is an `unsubscribe` called with nil for the listener for example. In this case the first chan added will be deleted even we unsubscribed with nothing.

This can of thing should not happen but if for any reasons it happen we will have a side effect here but if we clear the memory when the slice is empty already (after the check of the chan) then it's no problem

ilgooz: I don't understand: `I'm just afraid about some possible side effects like another part add a new chan in the listeners[channel] directly`

A channel to `listeners[channel]` only added within `Subscribe()`

ilgooz: Call to a `Unsubscribe()` will not remove any channel from `listeners[channel]` because we compare pointer values of channels.

ilgooz: oh, i got it. i'll change the line

ilgooz: done

PR #362, opened 2018-08-13
Feature/tag executions

### LINKED ISSUE #201: Use a tag system to identify requests
It will be nice to be able to listen to a specific `executionID` in the API `Core.ListenResult`.
The core could also close the stream when the result with the right `executionID` is send back.

antho1404:
We can now add some tags for an execution and filter results with specific tags. The filter passed in the `ListenResult` API is an all included filter. The list of tags given in this call needs to be present in the execution's tags

fixes #201


api/core/api.proto
```diff
@@ -79,13 +79,15 @@ message EventData {
...
 // }
 // ```
 message ListenResultRequest {
-  string serviceID = 1;     // The Service ID. Generated when using the [`DeployService` API](#deployservice).
-  string taskFilter = 2;    // __Optional.__  The task's key to filter. The task must match this key. The default is `*` which matches any task.
-  string outputFilter = 3;  // __Optional.__ The output's key from the task to filter. The task must return this output's key. The default is `*` which matches any output.
+  string serviceID = 1;           // The Service ID. Generated when using the [`DeployService` API](#deployservice).
+  string taskFilter = 2;          // __Optional.__  The task's key to filter. The task must match this key. The default is `*` which matches any task.
+  string outputFilter = 3;        // __Optional.__ The output's key from the task to filter. The task must return this output's key. The default is `*` which matches any output.
+  repeated string tagFilter = 4;  // __Optional.__ List of tags to filter. This is am "match all" list. All tags in parameters should be included in the execution to match.
```
NicolasMahe: Typo on "This is am "match all" list."

NicolasMahe: What about renaming `tagFilter` to `tagsFilter` to explicitly show the "many" tags

api/core/api.proto
```diff
@@ -97,14 +99,16 @@ message ListenResultRequest {
...
 message ResultData {
-  string executionID = 1; // The unique identifier of the execution.
-  string taskKey = 2;     // The key of the executed task.
-  string outputKey = 3;   // The output's key from the returned task.
-  string outputData = 4;  // The output's data from the returned task, encoded in JSON.
+  string executionID = 1;   // The unique identifier of the execution.
+  string taskKey = 2;       // The key of the executed task.
+  string outputKey = 3;     // The output's key from the returned task.
+  string outputData = 4;    // The output's data from the returned task, encoded in JSON.
+  repeated string tags = 5; // List of tags associated to the task
```
NicolasMahe: > "List of tags associated to the task"

This is not the task's tags but the execution's tags.

Should be updated to: "The list of tags associated to the execution"

api/core/api.proto
```diff
@@ -114,13 +118,15 @@ message ResultData {
...
 // }
 // ```
 message ExecuteTaskRequest {
-  string serviceID = 1; // The Service ID. Generated when using the [`DeployService` API](#deployservice).
-  string taskKey = 2;   // The task's key to execute.
-  string inputData = 3; // The inputs of the task to execute, encoded in JSON.
+  string serviceID = 1;     // The Service ID. Generated when using the [`DeployService` API](#deployservice).
+  string taskKey = 2;       // The task's key to execute.
+  string inputData = 3;     // The inputs of the task to execute, encoded in JSON.
+  repeated string tags = 4; // List of tags to assign to this task
```
NicolasMahe: Again, This is not the task's tags but the execution's tags.

Should be updated to "The list of tags to associate to the execution"

api/core/api.proto
NicolasMahe: What about to rename `tags` to `executionTags`. It provide more context and match the `executionID` style.

api/core/api.proto
NicolasMahe: Same here about renaming `tags` to `executionTags`.

api/core/listen_result.go
```diff
@@ -71,3 +72,15 @@ func isSubscribedTask(request *ListenResultRequest, e *execution.Execution) bool
 func isSubscribedOutput(request *ListenResultRequest, e *execution.Execution) bool {
 	return array.IncludedIn([]string{"", "*", e.Output}, request.OutputFilter)
 }
+
+func isSubscribedExecution(request *ListenResultRequest, e *execution.Execution) bool {
```
NicolasMahe: I suggest to rename this function `isSubscribedExecution` to something that explicitly show the tags like: `isSubscribedExecutionTags`

api/core/api.proto
antho1404: isn't it better `tagFilters` ?

api/core/api.proto
antho1404: I like just `tags`, we are on the result of an execution already and actually maybe the `executionID` should just be `ID`

api/core/api.proto
antho1404: the same here I feel `tags` is better because we are already in an execution `ExecuteTaskRequest`

api/core/listen_result.go
antho1404: what about `isSubscribedToTags` at the same time I can rename the `isSubscribedOutput` to `isSubscribedToOutput`

api/core/api.proto
NicolasMahe: As it is the gRPC API, I suggest to go with the most explicit names.
In the JS and Go libs, of course we will use only "tags" if it makes sense.
Also, all the variable are composed of 2 words. Let's follow the convention we used.

api/core/api.proto
NicolasMahe: Let's continue the conversation on the other comment ;) It's the same subject.

api/core/listen_result.go
NicolasMahe: I'm ok with your suggestion. I just like more context to the names.

api/core/api.proto
NicolasMahe: I'm asking will.

NicolasMahe: Go for `tagFilters`

api/core/api.proto
antho1404: `taskKey`, `outputKey` and `outputData` are composed with 2 words because they are nested data of the execution, and also because it kind of define the type of value needed here either a key or a data (that we use to symbolize json encoded payload). In the case of `tags` it's neither a nested data/object of the execution and we don't define the "type" otherwise we would call it something like `tagList`.

Really I don't see much the point of adding this. It is attached to a payload so we have the context already. We don't add `executionTaskKey` or `executionOutputKey` so there is no reason to do the same for the tags for me

api/core/api.proto
```diff
@@ -79,13 +79,15 @@ message EventData {
...
 // }
 // ```
 message ListenResultRequest {
-  string serviceID = 1;     // The Service ID. Generated when using the [`DeployService` API](#deployservice).
-  string taskFilter = 2;    // __Optional.__  The task's key to filter. The task must match this key. The default is `*` which matches any task.
-  string outputFilter = 3;  // __Optional.__ The output's key from the task to filter. The task must return this output's key. The default is `*` which matches any output.
+  string serviceID = 1;           // The Service ID. Generated when using the [`DeployService` API](#deployservice).
+  string taskFilter = 2;          // __Optional.__  The task's key to filter. The task must match this key. The default is `*` which matches any task.
+  string outputFilter = 3;        // __Optional.__ The output's key from the task to filter. The task must return this output's key. The default is `*` which matches any output.
+  repeated string tagFilters = 4; // __Optional.__ List of tags to filter. This is a "match all" list. All tags in parameters should be included in the execution to match.
```
NicolasMahe: Can you add "The" at the beginning of the comment to look the same as other comments

api/core/api.proto
```diff
@@ -97,14 +99,16 @@ message ListenResultRequest {
...
 message ResultData {
-  string executionID = 1; // The unique identifier of the execution.
-  string taskKey = 2;     // The key of the executed task.
-  string outputKey = 3;   // The output's key from the returned task.
-  string outputData = 4;  // The output's data from the returned task, encoded in JSON.
+  string executionID = 1;   // The unique identifier of the execution.
+  string taskKey = 2;       // The key of the executed task.
+  string outputKey = 3;     // The output's key from the returned task.
+  string outputData = 4;    // The output's data from the returned task, encoded in JSON.
+  repeated string tags = 5; // The list of tags associated to the execution
```
NicolasMahe: Can you add "The" at the beginning of the comment to look the same as other comments

api/core/api.proto
```diff
@@ -114,13 +118,15 @@ message ResultData {
...
 // }
 // ```
 message ExecuteTaskRequest {
-  string serviceID = 1; // The Service ID. Generated when using the [`DeployService` API](#deployservice).
-  string taskKey = 2;   // The task's key to execute.
-  string inputData = 3; // The inputs of the task to execute, encoded in JSON.
+  string serviceID = 1;     // The Service ID. Generated when using the [`DeployService` API](#deployservice).
+  string taskKey = 2;       // The task's key to execute.
+  string inputData = 3;     // The inputs of the task to execute, encoded in JSON.
+  repeated string tags = 4; // The list of tags to associate to the execution
```
NicolasMahe: My argument to name it `executionTags` is to be as explicit as possible without needing anymore context.
Also, we name everything by using 2 words. So let's continue this way.

That's what we did for all other parameters. Eg:

https://github.com/mesg-foundation/core/blob/5f98a3f23a1653664fbf8e3a61bbbb1027cd378d/api/core/api.proto#L70-L73

https://github.com/mesg-foundation/core/blob/5f98a3f23a1653664fbf8e3a61bbbb1027cd378d/api/service/api.proto#L39-L43

antho1404: for the rest of the conversation
https://github.com/mesg-foundation/core/pull/362/files/12d3d3cba19155ab54cb58118d3b4809656957e6#r209819975

I would love to have other feedbacks on that @ilgooz @krhubert ?

Naming problem again :)

For the naming of the tag parameter sent when executing a task or listening for the result of a task.

- tags (simple, related to the context, can be reused in any other kind of resources to add metadata)
- executionTags (strongly coupled to the execution, gives a lot of details without knowing the context, we often have 2 words attribute)

What do you think ?

ilgooz:
Why not using key, value pair for filtering where _key_ is the tag name and _value_ is the tag's value?

But it is also possible using _key:value_ pair with this implementation by combining key and value into one string like `companyID:123` as a tag in the current implementation. But this leaves more work to devs and requires some type casting.

What are your thoughts?

antho1404:
I like this idea and maybe even more following your idea to have `companyID:123`.
Why not do like most rabbitmq for example with their routing to be able to pass some wildcard.

We could listen for the tag `companyID:*`
and execute with the tag `companyID:123`

Of course after it's up to the developer to use it's own system `company:123`, `company-123`, `company{123}`... whatever. We can be really flexible and just check based on this wildcard.

This could add a lot of flexibility, for a pattern like that `application:feature:id` where we could listen for everything or thing from a specific feature `applicationX:featureX:*`, or any features of a specific application `applicationX:*:*` and so on

What do you think ? Is it coherent with the idea you proposed ?

antho1404:
I feel the key value is less intuitive and you need to add a value so in some cases it might not be needed.

Example: I want to create an app and save all the executions in a time series database. With tags I can just put a tag `metrics` or even `metrics:*` and listen for the same "tag".
With a key value I would need maybe to have a filter like `{metrics: true}` or `{metrics: '*'}`.
I feel it harder to understand

NicolasMahe:
I prefer the RabbitMQ style rather than the key:value.
But, at the end, RabbitMQ's `application:feature:id` is the same as having a list of tags: `[application, feature, id]`.

I don't think we need a dedicated key:value system.
As the tags are given by the application, if the dev want to implement a key:value system, he can concat the key and value to a string and give it as a tag (and vice versa).

So providing the simplest system allow dev to built on top of it what they want do use.

antho1404:
Agree to do that wildcard feature later if we all agree on the tag system, this feature can be built on later and we can merge this as soon as possible. For me it's good to merge, just want to wait some feedbacks on https://github.com/mesg-foundation/core/pull/362#discussion_r209860921 to make sure to choose the right naming but I'm fine with the actual implementation :)

api/core/api.proto
ilgooz: maybe name of _ResultData_ should be changed as _ExecutionResult_, replace _executionID_ with ID and just use _tags_.

cmd/service/execute.go
```diff
@@ -77,11 +80,12 @@ func executeHandler(cmd *cobra.Command, args []string) {
 	fmt.Println(aurora.Bold(execution.OutputData).String())
 }
 
-func executeTask(serviceID string, task string, data string) (execution *core.ExecuteTaskReply, err error) {
+func executeTask(serviceID string, task string, data string, tags []string) (execution *core.ExecuteTaskReply, err error) {
```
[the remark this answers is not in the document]
antho1404: we have the types for all arguments in all the codebase so let's keep it that way I think, I feel it gives more details so it's good

cmd/service/execute_test.go
```diff
@@ -3,7 +3,10 @@ package service
 import (
 	"testing"
 
+	"github.com/spf13/cobra"
+
```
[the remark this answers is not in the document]
antho1404: FYI : result of my gofmt

screen shot 2018-08-17 at 6 00 18 pm

The empty line is still here but I updated it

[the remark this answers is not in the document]
krhubert: Yes, because gofmt don't remove empty lines from imports.

api/core/listen_result.go
```diff
@@ -32,13 +32,14 @@ func (s *Server) ListenResult(request *ListenResultRequest, stream Core_ListenRe
 			return ctx.Err()
 		case data := <-subscription:
 			execution := data.(*execution.Execution)
-			if isSubscribedTask(request, execution) && isSubscribedOutput(request, execution) {
+			if isSubscribedToTags(request, execution) && isSubscribedToTask(request, execution) && isSubscribedToOutput(request, execution) {
 				outputs, _ := json.Marshal(execution.OutputData)
```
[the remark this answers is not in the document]
antho1404: good catch, it's present in other places too, I created a separated issue for that https://github.com/mesg-foundation/core/issues/374

cmd/service/execute.go
```diff
@@ -50,12 +51,14 @@ func executeHandler(cmd *cobra.Command, args []string) {
 	})
 	utils.HandleError(err)
 	taskKey := getTaskKey(cmd, serviceReply.Service)
-	taskData, err := getData(cmd, taskKey, serviceReply.Service)
+	taskData, err := getData(cmd, taskKey, serviceReply.Service, executeData)
 	utils.HandleError(err)
 
+	tags := []string{strconv.FormatInt(time.Now().UnixNano(), 10)}
```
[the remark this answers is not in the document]
antho1404: This is to have a unique tag to listen to this exact execution and only this one. But I'm replacing with an UUID and a comment

api/core/execute_test.go
```diff
@@ -127,14 +127,14 @@ func TestExecuteFunc(t *testing.T) {
 			"test": {},
 		},
 	}
-	id, err := execute(srv, "test", map[string]interface{}{})
+	id, err := execute(srv, "test", map[string]interface{}{}, []string{})
```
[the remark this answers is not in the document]
antho1404: yes nil is fine too. Do you want me to replace with nil instead

api/core/listen_result.go
```diff
@@ -32,13 +32,14 @@ func (s *Server) ListenResult(request *ListenResultRequest, stream Core_ListenRe
 			return ctx.Err()
 		case data := <-subscription:
 			execution := data.(*execution.Execution)
-			if isSubscribedTask(request, execution) && isSubscribedOutput(request, execution) {
+			if isSubscribedToTags(request, execution) && isSubscribedToTask(request, execution) && isSubscribedToOutput(request, execution) {
```
ilgooz: to 
```go
if isSubscribedToTags(request, execution) && 
   isSubscribedToTask(request, execution) && 
   isSubscribedToOutput(request, execution) {
}

ilgooz:
lgtm

PR #368, opened 2018-08-15
Use a specific tag for the mesg/core docker image in CLI, fixes #264

### LINKED ISSUE #264: Use a specific tag for the mesg/core docker image in CLI
Right now, we are using the latest tag when starting the Core in docker. The problem is if the user already has a core image on his computer, it may not update to the real latest.

To fix this problem, we should specify a tag in the CLI when starting the `core` in docker:
`mesg/core:v1.0.0` instead of `mesg/core:latest`.

krhubert:
It also fixes the hardcoded core container port.

fixes #264


config/api.go
```diff
@@ -40,5 +42,7 @@ func setAPIDefault() {
 	viper.SetDefault(ServicePathDocker, filepath.Join("/mesg", "services"))
 	os.MkdirAll(viper.GetString(ServicePathDocker), os.ModePerm)
 
-	viper.SetDefault(CoreImage, "mesg/core:latest")
+	// Keep only the first part if Version contains space
+	coreTag := strings.Split(version.Version, " ")
+	viper.SetDefault(CoreImage, "mesg/core:"+coreTag[0])
```
NicolasMahe: I fix a bug here when the dev release, available on GitHub, start the image with `mesg/core:dev build XXXXX`.
Now it start with image `mesg/core:dev`.

dev-cli
```diff
@@ -1,9 +1,7 @@
 #!/bin/bash
 
-image=mesg/core:local
-version="dev local"
-env MESG_CORE_IMAGE=$image \
-  go run \
+version="local"
+go run \
```
NicolasMahe: Renaming all local version to local

scripts/build-test-core.sh
```diff
@@ -2,5 +2,5 @@
 
 echo "Building mesg/core:local..."
 docker pull mesg/core:latest
-docker build -t mesg/core:build --target build --build-arg version="dev local" .
-docker build -t mesg/core:local --build-arg version="dev local" .
\ No newline at end of file
+docker build -t mesg/core:build --target build --build-arg version="local" .
+docker build -t mesg/core:local --build-arg version="local" .
```
NicolasMahe: Same. Renamed all local version to local

PR #371, opened 2018-08-16
move service deployment related server side logic from cli

ilgooz:
* @antho1404 can you please update the comments on _api/core/api.proto_ to match with the changes?

### about the update DeployService API
I have merged Git url and uploading service context options to same gRPC DeployService handler for deploying services.

What do you prefer? Current one or split this as _DeployService (uploading service context)_ and create new _DeployServiceFromURL_ handler?
One **small** advantage of splitting them might be to not requiring to deal with gRPC stream for simple url deployments.

### about the new mesg package
* please give feedbacks about new mesg package. goal of mesg package is to serve core's all features with a high level API. gRPC server handlers will use methods from mesg package, instead of directly accessing to internal packages like _service_, _execution_, etc. we shouldn't keep any business logic within api handlers and should move them under mesg package.
* there are only two methods in the mesg package to deploy services for now. i'm thinking about adding more methods to cover all features used by gRPC server handlers. 
* i'm planning to move _mesg/deploy_deployer.go_ to service package in future, while refactoring service to make it newable. APIs will be like: `Manager.CreateService()`, `Manager.CreateServiceFromURL()`


mesg/deploy_deployer.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	if err := d.gitClone(url, path); err != nil {
+		return nil, nil, err
+	}
+	return d.deploy(path)
+}
+
+func (d *serviceDeployer) gitClone(repoURL string, path string) error {
```
krhubert: I think gitClone should be moved to some utils pkg (maybe xgit?) but it's not part of service Deploeyr

krhubert: + missing docs

mesg/deploy_deployer.go
```diff
@@ -0,0 +1,135 @@
...
+	if u.Fragment != "" {
+		options.ReferenceName = plumbing.ReferenceName("refs/heads/" + u.Fragment)
+		u.Fragment = ""
+	}
+	options.URL = u.String()
+	_, err = git.PlainClone(path, false, options)
+	return err
+}
+
+func (d *serviceDeployer) FromGzippedTar(r io.Reader) (*service.Service, *importer.ValidationError, error) {
```
krhubert: missing docs + few other methods.

mesg/deploy_deployer.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	if err := archive.Untar(r, path, &archive.TarOptions{
+		Compression: archive.Gzip,
+	}); err != nil {
+		return nil, nil, err
+	}
+	return d.deploy(path)
+}
+
+func (d *serviceDeployer) deploy(path string) (*service.Service, *importer.ValidationError, error) {
```
krhubert: I know golint will not highlight this, but even for unexported methods/funcs it's nice to have some docs (the docs are for mainly us, not to satisfy linters :))

api/core/api.proto
```diff
@@ -31,8 +31,8 @@ service Core {
   rpc StopService (StopServiceRequest) returns (StopServiceReply) {}
 
   // Deploy a service to [Core](../guide/start-here/core.md). This will give you an unique identifier which is used to interact with the service.
-  rpc DeployService (DeployServiceRequest) returns (DeployServiceReply) {}
-
+  rpc DeployService (stream DeployServiceRequest) returns (stream DeployServiceReply) {}
+
```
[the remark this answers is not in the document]
ilgooz: done

api/core/deploy.go
```diff
@@ -1,19 +1,83 @@
...
+	i    int64
+}
+
+func newDeployServiceStreamReader(stream Core_DeployServiceServer) *deployServiceStreamReader {
+	return &deployServiceStreamReader{stream: stream}
+}
+
+func (r *deployServiceStreamReader) GetURL() (url string, err error) {
+	message, err := r.stream.Recv()
+	r.data = message.GetChunk()
```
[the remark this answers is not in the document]
ilgooz: done

docs/api/core.md
```diff
@@ -747,7 +750,9 @@ The reply's data of `DeployService` API.
 
 | Field | Type | Description |
 | ----- | ---- | ----------- |
-| serviceID | [string](#string) | The generated identifier of the deployed service. Use this ID with other APIs. |
+| status | [string](#string) | status will be sent after each status change. |
```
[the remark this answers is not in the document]
ilgooz: i guess we can keep this one as it is

cmd/service/deploy.go
```diff
@@ -24,11 +29,110 @@ To get more information, see the [deploy page from the documentation](https://do
...
+			sp.Suffix = " " + status
+		case serviceID != "":
+			sp.Stop()
+			deployment <- deploymentResult{serviceID: serviceID}
+			return
+		case validationError != "":
+			sp.Stop()
+			fmt.Println(aurora.Red(validationError))
+			fmt.Println("Run the command 'service validate' for more details")
+			os.Exit(0)
```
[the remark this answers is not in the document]
ilgooz: done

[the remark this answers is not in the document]
ilgooz: we need better error handling in cmd package

dev-core
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 
-./scripts/build-test-core.sh
-./dev-cli stop
-./dev-cli start
+# chain these to stop execution on failure
```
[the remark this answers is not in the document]
ilgooz: done

mesg/deploy_deployer.go
ilgooz: i see that we have quite a lot _utils.go_, _utils_ package inside the project. i think we should reduce them and serve under a root level utils package as possible but use different package names for each *different* util under _utils/_ e.g.:

```
util/
  gitutil/ > gitutil.Clone()
```

lets also give a second check to this article about package namings: https://blog.golang.org/package-names

api/core/api.proto
```diff
@@ -199,7 +199,15 @@ message StopServiceReply {
 // }
 // ```
 message DeployServiceRequest {
-  service.Service service = 1; // The service's definition to deploy. [Details here](./service-type.md).
+  oneof value {
+    // Git repo url of service. If url provided, stream will be closed after 
+    // first receive.
+    string url = 1;
```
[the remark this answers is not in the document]
ilgooz: done

api/core/api.proto
```diff
@@ -211,7 +219,18 @@ message DeployServiceRequest {
 // }
 // ```
 message DeployServiceReply {
-  string serviceID = 1; // The generated identifier of the deployed service. Use this ID with other APIs.
+  oneof value {
+    // status will be sent after each status change.
+    string status = 1;
```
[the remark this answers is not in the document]
ilgooz: done

mesg/mesg.go
```diff
@@ -0,0 +1,34 @@
...
+type Option func(*MESG)
+
+// New creates a new MESG with given options.
+func New(options ...Option) (*MESG, error) {
+	m := &MESG{}
+	for _, option := range options {
+		option(m)
+	}
+	var err error
+	if m.container == nil {
```
krhubert: Avoid nested if's 

```
if m.container != nil {
  return m, nil
}
var err error
m.container, err = container.New()
if err ....
```

ilgooz: thanks, in this case i kinda don't like making an early return here because we might add some more lines under these code that not related with initializing a container.

mesg/deploy_deployer.go
ilgooz: docs are done

mesg/deploy_deployer.go
ilgooz: done

mesg/deploy_deployer.go
ilgooz: done

api/server.go
```diff
@@ -63,8 +64,18 @@ func (s *Server) Stop() {
 
 // register all server
 func (s *Server) register() {
```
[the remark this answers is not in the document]
ilgooz: done

ilgooz:
@antho1404 please note that name of package _mesg_ changed as _api_ in this next PR. https://github.com/mesg-foundation/core/pull/383

cmd/service/dev.go
```diff
@@ -28,7 +29,10 @@ func init() {
 }
 
 func devHandler(cmd *cobra.Command, args []string) {
-	serviceID, err := createService(defaultPath(args))
+	serviceID, isValid, err := createService(defaultPath(args))
+	if !isValid {
+		os.Exit(1)
```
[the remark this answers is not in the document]
ilgooz: In the gRPC side, validation errors send as a part of gRPC [response](https://github.com/mesg-foundation/core/blob/b17faa6/api/core/api.proto#L238) stream. But server-side errors of core returned/send as an error in gRPC handlers as we do always. In this case I like separating both of them like this. Validation errors should not be returned as a gRPC handler error because they are different kind of errors that needed to be take action on in the client side. So to identify them as a separate kind we need to split them like the current implementation.

_isValid_ set to _false_ when there is a validation error sent by server [here](https://github.com/mesg-foundation/core/blob/5e7fd59/cmd/service/deploy.go#L140) and as you can see there is some _Print_ function above that line to let devs know there is a validation error. If there is a validation error, we should exit with 1 after we printed those messages but i quite don't like exiting in goroutine because this makes things un-trackable. So instead we send isValid false value through a channel which is read within _ deployService()_ and returned as its return value. so callers of deployService() should check for isValid and immediately exit with 1. If you don't like this approach we can change isValid as _validationErrors_ and fill it with validation errors instead. And callers of deployService() can check it and print its errors with a util function like 	_utils.HandleValidationError(validationError)_. So this function can print the validation error and exit with 1.

For the second question, _err_ return value of deployService() still filled with server errors and we check them in the code with _utils.HandlerError(err)_. I don't see a problem here.

database/services/save.go
```diff
@@ -6,16 +6,17 @@ import (
...
 	db, err := open()
 	defer close()
 	if err != nil {
-		return "", err
+		return err
 	}
-	hash = service.Hash()
-	return hash, db.Put([]byte(hash), bytes, nil)
+	hash := service.Hash()
+	service.Id = hash
```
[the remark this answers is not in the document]
ilgooz: I made this change because I needed to return _*service.Service_ type in [deploy api](https://github.com/mesg-foundation/core/blob/5e7fd59d0394b9a0cb387df96261f43ac1bd1850/mesg/deploy.go#L21). And its callers needs to read service's ID. 

I know this is messy but it is a necessary evil during this big refactor :) If you really don't like it I can remove it from this PR and make changes on deploy api to cover it but eventually deploy api  should return a proper Service type with id. Also if you review the corresponding commits related with this you can see all the changes made to _service.ID_ in one place. Easier to review.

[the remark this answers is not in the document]
ilgooz: After second check, I see that I didn't separate changes for service.Id to different commits. fyi.

ilgooz:
needs review

cmd/service/deploy.go
```diff
@@ -24,11 +29,125 @@ To get more information, see the [deploy page from the documentation](https://do
...
+
+func deployServiceSendServiceContext(path string, stream core.Core_DeployServiceClient) error {
+	archive, err := archive.TarWithOptions(path, &archive.TarOptions{
+		Compression: archive.Gzip,
+	})
+	if err != nil {
+		return err
+	}
+
+	buf := make([]byte, 1024)
```
NicolasMahe: 1ko seems a bit small. 

> Spoiler alert: it's 16KiB - 64KiB, according to oral tradition.
https://github.com/grpc/grpc.github.io/issues/371

cmd/service/deploy.go
```diff
@@ -24,11 +29,125 @@ To get more information, see the [deploy page from the documentation](https://do
...
+}
+
+type deploymentResult struct {
+	serviceID string
+	err       error
+	isValid   bool
+}
+
+func readDeployReply(stream core.Core_DeployServiceClient, deployment chan deploymentResult) {
+	sp := spinner.New(utils.SpinnerCharset, utils.SpinnerDuration)
```
NicolasMahe: What don't you use `utils.StartSpinner` or `utils.ShowSpinnerForFunc`?

NicolasMahe: If the existing helper function are not useful for what you are doing, I suggest to create a helper function in order to reuse it if it make sense.

cmd/service/deploy.go
ilgooz: please check this out https://ops.tips/blog/sending-files-via-grpc/

ilgooz: _It seems like at 1024bytes (1KB) we have a decent throughput._

_I was expecting something like 32KB: (1 << 15) to be the most optimal but it looks like it’s one of the worst scenarios regarding variance._

cmd/service/deploy.go
ilgooz: I had to create it manually in the first for some reason... I'll see if i can use current util for this

ilgooz: done

Issue #377, opened 2018-08-20
Reorganise api package


NicolasMahe:
The package `api` should contains ALL functions (business logic) of MESG that are available through gRPC, CLI (and more in the future).
The functions of the `api` package should be generic, meaning that they should not depend on any communication library (gRPC for instance) and call other packages of Core.

The gRPC and CLI's commands should be move to a subpackage of the api package.

Api folder structure:
```
api/
interface/
|  grpc/
|  cli/
```

NicolasMahe:
Related to https://github.com/mesg-foundation/core/issues/337 for the Manager type.

PR #383, opened 2018-08-22
introducing new api package

ilgooz:
resolves https://github.com/mesg-foundation/core/issues/377.
depends on https://github.com/mesg-foundation/core/pull/371, so it needed to be merged first.


ilgooz:
needs review

ilgooz:
needs review. it'd be much simpler if you review this PR per commit. since we move old api folder to interface/grpc you'll not be able to see diff on each file if you review from total view. But if you review by commits you'll see each single diffs including renamed files and their further changes in the next commits.

NicolasMahe:
I agree with Anthony,

It feel like you go halfway from no types in one package to many types in many packages.
I suggest to keep one type per package and a few packages for any group on function (could be: one package for `service`, one `event`, etc...)

ilgooz:
@antho1404 
### 1-
having a specific type for each feature is done to isolate functions from each other. you can see this kind of implementation detail in some of the [standard](https://golang.org/src/net/http/client.go?s=24325:24403#L738) packages too _(there are more but i don't have the time to do a code research for now)_. 

this kinda use is specially ideal when we expose some package level functions _(APIs)_ and let them to return basic types and some model types _(e.g. *service.Service)_ instead of returning _doER_ types _(these private structs we have)_ and also keep the isolation between API functions. 

and if we don't use different types for each feature and create one struct instead, then we'll a have generic/giant struct type that does everything which is a poor way to organize this kind of package, not good for testing and we'll have possible collapses in the function names.

### 2-
about the options, I understand your concern. but i  followed a pattern that where we have multiple methods that does the same job with a slight difference and receives the same options. like [1](https://github.com/mesg-foundation/core/blob/9eef3b9/api/deploy.go#L23) and [2](https://github.com/mesg-foundation/core/blob/9eef3b9/api/deploy.go#L32). 

In this case i like to pass options to initializer function and it'll iterate options one time and set it to struct fields. This way we don't need to write the same code for each function.

### conclusion
I understand your concern about this, but lets not forget that we're composing almost all of the internal packages in this new api package. api package is a general purpose package unlike a single purpose package like service, execution, container etc. Being a general purpose package can easily result with a messy code if we don't apply necessary isolations between features.  I like to hear more suggestions about this.

ilgooz:
@NicolasMahe separating api package to child packages can be nice but might complicate things a bit, currently we don't have 10s of API methods so this version is also pretty maintainable if we keep  the isolation. I think we can switch to child packages in future, when we think it's needed and in another PR. Happy to hear more suggestions about this.

ilgooz:
@antho1404 
_We have services for tests in different packages. I like the idea to centralize these tests in on place but let's do that for all of them or none of them. We have more in cmd/service/test, service/importer/test Also I would recommend to do that in another PR._

I agree to this except we should keep test services for importer in it's own context. Because importer is a parser, it might need lots of test services to test syntax errors etc. and the test services it uses will not needed to be used to test things in other packages.

Issue #354, opened 2018-08-09
database/service: out of sync when mesg-core restarted


krhubert:
Database may be out of sync when mesg-core is killed/restarted and there is no way to sync it back.

Following some cli commands may show outdated output - for example `mesg-core servcie list`

NicolasMahe:
Maybe it's because the core doesn't quit gracefully and and the gRPC server get killed and make the db crash.

https://github.com/mesg-foundation/core/blob/d017e194432cfce501a2b41bd8ffb103eb1eaa39/core/main.go#L23-L28

krhubert:
I can't reproduce it anymore. Closing

PR #401, opened 2018-08-27
gracefully stop gRPC servers

ilgooz:
related with https://github.com/mesg-foundation/core/issues/354#issuecomment-416206710


core/main.go
```diff
@@ -16,21 +16,37 @@ func main() {
...
+	unixServer := &api.Server{
 		Network: "unix",
 		Address: viper.GetString(config.APIServerSocket),
-	})
+	}
+
+	go startServer(tcpServer)
+	go startServer(unixServer)
+
+	closing := make(chan struct{}, 2)
```
antho1404: I'm not sure I understand how this is useful.
Why not doing something like 
```go
defer tcpService.Close()
defer unixServer.Close()
```
instead of having to create this chan that we pass as a parameters of the `closeServer` we could close the server directly and not in some goroutines. I might be missing something here

ilgooz: yes actually this is needed. Close() will stop accepting incoming gRPC requests and gracefully wait current ones to end. so Close() will block until all accepted gRPC requests are done. We should only exit the program after both Close() calls returned.

ilgooz: and the reason why we use a goroutine and a chan is to close both server concurrently and don't force one to wait until another one closes.

Issue #348, opened 2018-08-08
Improve service errors - Returns errors directly from the `validate` functions


NicolasMahe:
https://github.com/mesg-foundation/core/blob/bd84ad1d54eadfb8aa2bd79ac0c7c68f775f9c03/execution/complete.go#L20-L26
This previous code is one of many similar code in the app when the caller have to create an error that the called function should return.

I think the functions `validate` or `isValid` should return specific errors even if it means to pass more  argument.
We could un-export the `validate` functions in favor of exported function on the service type like:
```go
func (service *Service) ValidateEvent(eventKey string, data map[string]interface{}) error
func (service *Service) ValidateTask(taskKey string, data map[string]interface{}) error
func (service *Service) ValidateTaskOutput(taskKey, outputKey string, data map[string]interface{}) error
```
https://github.com/mesg-foundation/core/blob/bd84ad1d54eadfb8aa2bd79ac0c7c68f775f9c03/service/validation.go#L30-L58

@mesg-foundation/core what do you think?

NicolasMahe:
@mesg-foundation/core please give feedback

ilgooz:
related with #350, #349, #337.

NicolasMahe:
What about returning directly the error instead of the warnings?

NicolasMahe:
Yes I like it!

NicolasMahe:
After dicussion with @antho1404, we suggest 2 family names for thoses validation functions:

- **Validate....** family: those functions should returns an array of warnings
- **Require....** family: those functions should call the validate functions and return an error if some warnings exist

Both family should be implemented in the service package.

PR #402, opened 2018-08-28
service: update error types to be more logicless, simplify validations

ilgooz:
depends on #383, it should be merged before this one. this PR will be rebased from dev after depended PR makes its way through dev.

please review this PR by commits.

### DIFF
--- a/api/execute_executor.go
+++ b/api/execute_executor.go
@@ -49,21 +49,6 @@ func (e *taskExecutor) checkServiceStatus(s *service.Service) error {
 // execute executes task.
 func (e *taskExecutor) execute(s *service.Service, taskKey string, taskInputs map[string]interface{},
 	tags []string) (executionID string, err error) {
-	serviceTask, taskFound := s.Tasks[taskKey]
-	if !taskFound {
-		return "", &service.TaskNotFoundError{
-			TaskKey:     taskKey,
-			ServiceName: s.Name,
-		}
-	}
-	warnings := s.ValidateParametersSchema(serviceTask.Inputs, taskInputs)
-	if len(warnings) > 0 {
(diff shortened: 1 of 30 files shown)


api/execute_executor.go
```diff
@@ -47,9 +47,24 @@ func (e *taskExecutor) checkServiceStatus(s *service.Service) error {
...
+	}
+	warnings := s.ValidateParametersSchema(serviceTask.Inputs, taskInputs)
+	if len(warnings) > 0 {
+		return "", &service.InvalidTaskInputError{
+			TaskKey:  taskKey,
+			Warnings: warnings,
+		}
+	}
+
+	exc, err := execution.Create(s, taskKey, taskInputs, tags)
```
antho1404: I'm curious to have some arguments to move the validation outside of the create. For me the validation here is really related to the execution so it make sense that the execution package does the validation. Same point for the event and the results

ilgooz: I can't 100% agree with this. calling methods of specialized packages like _execution_ with invalid inputs seemed to me a bit messy. I believe that these validations should be done in the api package, a place that where we compose lots of types and different internal packages. This compositions requires some validations and I think doing this in package api feels natural. For _Create()_, to me, there is no difference between checking if the given service is nil -which doesn't make sense- and checking the existence of task. 

But I'm not strongly against for your version, @NicolasMahe what do you think about this?

NicolasMahe: Also, I thought you will put the `validate` functions in the service package and they returns directly the error, like in https://github.com/mesg-foundation/core/issues/348#issuecomment-416183529

ilgooz: I'll do it in another PR

PR #409, opened 2018-08-29
service: add Getters and Validate, Require methods for task, event and output

ilgooz:


execution/complete.go
```diff
@@ -4,27 +4,24 @@ import (
...
-	output, ok := execution.Service.Tasks[execution.Task].Outputs[outputKey]
-	if !ok {
-		return &service.TaskOutputNotFoundError{
-			TaskKey:       execution.Task,
-			TaskOutputKey: outputKey,
-			ServiceName:   execution.Service.Name,
-		}
+	var (
+		s       = execution.Service
+		taskKey = execution.Task
```
krhubert: Remove var section??
It is used only once in func

api/deploy_deployer.go
```diff
@@ -105,10 +105,27 @@ func (d *serviceDeployer) deploy(path string) (*service.Service, *importer.Valid
...
-	d.injectConfigurationInDependencies(service, imageHash)
+	d.adjustFields(s)
+	d.injectConfigurationInDependencies(s, imageHash)
 
 	d.sendStatus(fmt.Sprintf("%s Completed.", aurora.Green("✔")), DONE)
-	return service, nil, services.Save(service)
+	return s, nil, services.Save(s)
+}
+
+func (d *serviceDeployer) adjustFields(s *service.Service) {
```
krhubert: No comment on method + adjustFileds tells nothing about what function does.

event/event.go
```diff
@@ -17,20 +17,12 @@ type Event struct {
 
 // Create creates an event.
```
krhubert: More docs about an create event - it creates based on event definition in service and checks if all data are valid/exist.

api/deploy_deployer.go
ilgooz: i'll remove that func within the next or 2nd PR

execution/complete.go
ilgooz: i just like gofmt to align multiple vars together, more readable

api/deploy_deployer.go
krhubert: kk

event/event.go
ilgooz: done

service/service.proto
```diff
@@ -15,32 +15,42 @@ message Service {
...
 // Events are emitted by the service whenever the service wants.
+// TODO(ilgooz) remove key field when Event type crafted manually.
 message Event {
-  string name = 1;                  // The event's name.
-  string description = 2;           // The event's description.
-  map<string, Parameter> data = 3;  // The list of data of this event.
+  string key = 4;                   // Event's key.
+  string name = 1;                  // Event's name.
+  string description = 2;           // Event's description.
+  string serviceName = 5;           // Event's service name.
```
NicolasMahe: Do you plan to keep `serviceName ` in `Event` in the future?
If not, how do you plan to access this data in the future from the Event type?

ilgooz: no i don't. i'll add private service field to _Event_ type.
```go
type Event struct {
   service *Service
}
```

antho1404: I really don't like to have the service in the event, we should try to keep things really separated and ask ourselves why we really need the service in the event. If it's just to have the name of the service in one error this can be resolved with a event error that is catch and where the calling function can add the service.
Have a double dependencies between resources always makes things way more complicated in the end

PR #410, opened 2018-08-29
Remove socket server, closes #304

### LINKED ISSUE #304: Remove socket server
### Problem

- Windows is not compatible with unix socket. So on windows, the socket server will never start.
- We already have a TCP server. So having a second server increase the security issues and maintenance.
- gRPC on Node is not (yet) compatible with socket

### Solution

Remove the socket server in favor of the TCP server.

Related to https://github.com/mesg-foundation/core/pull/297

krhubert:
#304


antho1404:
I did some more cleanup and also made sure that there is no breaking change. We still use `MESG_ENDPOINT` and `MESG_ENDPOINT_TCP`

NicolasMahe:
@ilgooz can you review this PR please. As Anthony work on it, he's not qualified to review

config/api_test.go
```diff
@@ -15,16 +15,11 @@ func assertViperDefault(t *testing.T, key string, expected string) {
 
 func TestAPIDefault(t *testing.T) {
```
ilgooz: test all config constants in config/api.go

antho1404: @NicolasMahe is working on a refactoring of the configs, this will be done on his PR. I don't want to touch much the configs to avoid a lot of conflicts

PR #411, opened 2018-08-29
Fix container status info, closes #293

### LINKED ISSUE #293: deamon.Status() should check container status
the problem found here is that in the current implementation when the cli checks if mesg-core is running, it just goes and checks if the service is around, which leads to that both mesg-core start and mesg-core status report that everything is fine, but as illustrated in the screenshot below that is far away from the truth.

image

in fact in this example mesg-core even panic'd and still the client is reporting that all is fine. i won't go into details about why it panic'd, that will be part of a PR i am going to make soon, to also allow service and app development under windows.

krhubert:
https://github.com/mesg-foundation/core/issues/293

To think, should we have `unknown` status? or print both statuses (or have verbose mode). 

For now, it fixes the pbm.

### DIFF
--- a/daemon/status.go
+++ b/daemon/status.go
@@ -6,5 +6,19 @@ import (
 
 // Status returns the Status of the docker service of the daemon.
 func Status() (container.StatusType, error) {
-	return defaultContainer.ServiceStatus(Namespace()) //TODO: should it be containerStatus?
+	//TODO: should it be containerStatus?
+	serviceStatus, err := defaultContainer.ServiceStatus(Namespace())
+	if err != nil {
+		return container.STOPPED, err
+	}
+
+	containerStatus, err := defaultContainer.Status(Namespace())
+	if err != nil {
+		return container.STOPPED, err
+	}
+
+	if serviceStatus == containerStatus {
+		return serviceStatus, nil
+	}
+	return container.STOPPED, nil
 }


NicolasMahe:
After some brainstorm with @antho1404, we need to add more features to this PR about the `Status` function of the package `container`.

- [ ] Add 2 new status: `UNKNOWN` and `STARTING`:
```go
// Possible status for services.
const (
	UNKNOWN StatusType = iota
	STOPPED
	STARTING
	RUNNING
)
```

- [ ] Merge `container.ServiceStatus` to `container.Status`
We need to merge the status functions into one. This new function should:
  - find the docker's service. if not found, return status `STOPPED`. if err, return status `UNKNOWN` with error. otherwise, continue
  - find the docker's container. if not found, return status `STARTING`. if err, return status `UNKNOWN` with error. otherwise, continue
  - check the container's state. if it's `Running`, return `RUNNING`. otherwise return `STARTING`

- [ ] use the updated `container.Status` function in `daemon/status.go` and `service/status.go`

@krhubert what do you think?

container/container_test.go
```diff
@@ -116,19 +116,11 @@ func TestNonExistentContainerStatus(t *testing.T) {
 	dt := dockertest.New()
 	c, _ := New(ClientOption(dt.Client()))
 
-	dt.ProvideContainerList(nil, dockertest.NotFoundErr{})
+	dt.ProvideServiceInspectWithRaw(swarm.Service{}, nil, dockertest.NotFoundErr{})
```
ilgooz: can you also add lines to check call arguments of _ServiceInspectWithRaw()_ we need to know if _Status()_ calling _ServiceInspectWithRaw()_ with right namespace and options.

container/container_test.go
```diff
@@ -140,19 +132,20 @@ func TestExistentContainerStatus(t *testing.T) {
...
 	c, _ := New(ClientOption(dt.Client()))
 
+	dt.ProvideServiceInspectWithRaw(swarm.Service{}, nil, nil)
 	dt.ProvideContainerList(containerData, nil)
 	dt.ProvideContainerInspect(containerJSONData, nil)
 
 	status, err := c.Status(namespace)
 	require.Nil(t, err)
-	require.Equal(t, STOPPED, status)
+	require.Equal(t, RUNNING, status)
```
ilgooz: maybe this one should test _STARTING_ state, it kinda equivalent of _STOPPED_ in the previous implementation? it would be good if we test all the branches in the implementation of _Status()_ with different tests.

container/service_test.go
```diff
@@ -36,25 +36,13 @@ func TestStartService(t *testing.T) {
...
-		},
-	}
-
 	dt := dockertest.New()
 	c, _ := New(ClientOption(dt.Client()))
 
-	dt.ProvideContainerList(containerData, nil)
-	dt.ProvideContainerInspect(containerJSONData, nil)
+	dt.ProvideServiceRemove(nil)
+	dt.ProvideServiceInspectWithRaw(swarm.Service{}, nil, dockertest.NotFoundErr{})
```
ilgooz: can we test call arguments of _ServiceInspectWithRaw()_ as well, with _LastServiceInspectWithRaw()_ util.

krhubert: Done

container/container_test.go
krhubert: Done

NicolasMahe:
@ilgooz @antho1404  please review

NicolasMahe:
@antho1404 is working on fixing this PR

PR #414, opened 2018-08-30
service: use custom Service, Event, Task, Output, Parameter types

ilgooz:
**WIP**

fundamental part of #337.

This is an intermediate PR for service refactoring. This PR only contains transition to new types. Auto-generated gRPC types moved under _interface/grpc/service_.

Next PR will introduce new features to _service_ package and there'll be some refactoring on the code as well.


database/services/all.go
```diff
@@ -16,7 +17,7 @@ func All() ([]*service.Service, error) {
 	iter := db.NewIterator(nil, nil)
 	for iter.Next() {
 		var service service.Service
-		if err := proto.Unmarshal(iter.Value(), &service); err != nil {
+		if err := json.Unmarshal(iter.Value(), &service); err != nil {
```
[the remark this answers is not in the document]
ilgooz: we don't use gRPC types in the application layer anymore. gRPC types restricted be used only in interface/grpc

cmd/service/delete.go
```diff
@@ -43,7 +43,7 @@ func deleteHandler(cmd *cobra.Command, args []string) {
 			return
 		}
 		for _, service := range reply.Services {
-			args = append(args, service.Hash())
+			args = append(args, service.Id)
```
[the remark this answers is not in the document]
ilgooz: ID comes from our custom Service type. Id comes from gRPC Service type because it cannot convert id field described in the proto file to full-uppercase.

database/services/all.go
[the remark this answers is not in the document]
ilgooz: in future we may be need to make service search based on the field names in Service type. using json shouldn't be problem but if we feel like it is, go provides packages to convert Struct types to byte data.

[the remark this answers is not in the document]
antho1404: I'm ok, and anyway we might even change the database in the future and we can always optimize later when we need to

interface/grpc/core/core.go
```diff
@@ -0,0 +1,57 @@
...
+				Data:        toGRPCParameters(output.Data),
+			}
+		}
+		sv.Tasks[taskKey] = t
+	}
+
+	return sv
+}
+
+func toGRPCParameters(params map[string]*service.Parameter) map[string]*Parameter {
```
antho1404: should we use `toProtoParameter` ? I think the data are related to protobuf and sent via grpc. Small detail and I'm fine with that it's just to think about it maybe

interface/grpc/core/service.proto
```diff
@@ -1,10 +1,10 @@
 syntax = "proto3";
 
-package service;
+package core;
 
 // This is the definition of a MESG Service.
 message Service {
-  string id = 10;                           // Service's unique id hash.
+  string id = 10;                           // Service's unique id service hash.
```
antho1404: Can't we use `ID` here and like that we only have `ID` in the codebase and not `Id` and `ID`

interface/grpc/core/core.go
ilgooz: done

interface/grpc/core/service.proto
ilgooz: done

service/parameter.go
```diff
@@ -0,0 +1,17 @@
...
+	Name string `hash:"name:1" yaml:"name"`
+
+	// Description is the description of parameter.
+	Description string `hash:"name:2" yaml:"description"`
+
+	// Type is the data type of parameter.
+	Type string `hash:"name:3" yaml:"type"`
+
+	// Optional indicates if parameter is optional.
+	Optional bool `hash:"name:4" yaml:"optiona"`
```
NicolasMahe: Typo here: `optiona`

NicolasMahe:
I notice that the command logs doesn't work anymore.
I think it's because of the use of Docker volume and of the following line that try to read the database from the cli:
https://github.com/mesg-foundation/core/pull/414/files#diff-2c4fa6ab22f8116fb149c75bbd787027R43

PR #415, opened 2018-08-31
Added hall-of-fame to README

sergey48k:
As discussed with @antho1404 by email.

Hall-of-fame is a widget that helps you recognize new/trending/top contributors in your repo. It is designed to be very visual, and change frequently. It autoupdates every hour and requires no maintenance.

Note that by accepting this PR, you will make me a contributor, which means my face will be featured as ’new’ on your Hall-of-fame for one week.

Also note that the widget uses @antho1404  GitHub account to update stats, and makes 10-15 requests to GitHub API per hour (GitHub rate limit is 5,000 per hour).

Also, per @antho1404 request, here is an explanation how badges are computed:

* There are three categories of contributors: new, trending, and top. 
* 'New' and 'trending' badges contain commit counts for the last 7 days. 
* 'top' badge shows all time commits. 'Top' is used to fill up space up to 7 slots if there are not enough 'new' and 'trending' to take up all 7 slots.

More details can be found here: https://github.com/sourcerer-io/hall-of-fame/blob/develop/README.md


antho1404:
Thank you for this PR 👍 
Good to have the reminder for the rate limit about Github, we will check that and make sure it's fine

PR #429, opened 2018-09-05
Get dependency status based on docker service and docker container

antho1404:
Pull request based on this one https://github.com/mesg-foundation/core/pull/411


container/container.go
```diff
@@ -105,26 +106,65 @@ func (c *Container) FindContainer(namespace []string) (types.ContainerJSON, erro
...
+		if s.container == container && s.service == service {
+			return s.status, nil
+		}
 	}
+	return UNKNOWN, nil // This should never be reached but it's better than a panic :)
+}
+
+// containerExists return true if the docker service can be found, return false otherwise and return
+// any errors that are not `NotFound` errors
+func (c *Container) containerExists(namespace []string) (bool, error) {
```
ilgooz: `// containerExists checks if container with namespace can be found.`

container/container.go
```diff
@@ -105,26 +106,65 @@ func (c *Container) FindContainer(namespace []string) (types.ContainerJSON, erro
...
+	_, err := c.FindService(namespace)
+	return presenceHandling(err)
+}
 
-	if container.State.Running {
-		return RUNNING, nil
+// presenceHandling handle the error the check the presence of a docker resource.
+// It returns any error that are not docker `NotFound` errors
+// It returns a boolean that says if the docker resource exists
+func presenceHandling(err error) (bool, error) {
```
ilgooz:
```
// presenceHandling checks err to see it's a Docker NotFound error and if not
// it'll return the err back.
```

container/container.go
```diff
@@ -105,26 +106,65 @@ func (c *Container) FindContainer(namespace []string) (types.ContainerJSON, erro
...
+	if err == nil && container.State != nil &&
+		xstrings.SliceContains([]string{"exited", "dead"}, container.State.Status) {
+		return false, nil
+	}
+	return presenceHandling(err)
+}
+
+// serviceExists return true if the docker container can be found, return false otherwise and return
+// any errors that are not `NotFound` errors
+func (c *Container) serviceExists(namespace []string) (bool, error) {
```
ilgooz: // serviceExists checks if corresponding container for service namespace can be found.

container/container.go
```diff
@@ -105,26 +106,65 @@ func (c *Container) FindContainer(namespace []string) (types.ContainerJSON, erro
...
 }
 
-// Status returns the status of a docker container and servcie.
+// Status returns the status of the container based on the docker container and docker service
+// If any error happen the status will be `UNKNOWN`
+// Otherwise the following status will be applied:
+//   - RUNNING: When the container is running in docker (whatever the status of the service)
+//   - STARTING: When the service is running but the container is not yet started
+//   - STOPPED: When neither the container nor the service are running in docker
 func (c *Container) Status(namespace []string) (StatusType, error) {
```
ilgooz:
```
// status returns the status of the container based on the docker container and docker service.
// if any error occurs during the status check, status will be shown as UNKNOWN.
// otherwise the following rules will be applied to determine status:
//  - RUNNING: when the container is running in docker regardless of the status of the service.
//  - STARTING: when the service is running but the container is not yet started.
//  - STOPPED: when the container and the service is not running in docker.
```

container/container.go
```diff
@@ -105,26 +106,65 @@ func (c *Container) FindContainer(namespace []string) (types.ContainerJSON, erro
...
 	}
+	return UNKNOWN, nil // This should never be reached but it's better than a panic :)
+}
+
+// containerExists return true if the docker service can be found, return false otherwise and return
+// any errors that are not `NotFound` errors
+func (c *Container) containerExists(namespace []string) (bool, error) {
+	container, err := c.FindContainer(namespace)
+	if err == nil && container.State != nil &&
+		xstrings.SliceContains([]string{"exited", "dead"}, container.State.Status) {
```
ilgooz: `return !xstrings.SliceContains([]string{"exited", "dead"}, container.State.Status), nil`  ?

available statuses:
```created, restarting, running, removing, paused, exited, dead```

PR #441, opened 2018-09-10
drop using internal methods of service package from cmd/execute.go

ilgooz:


cmd/service/cast.go
```diff
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"strconv"
+
```
krhubert: Don't move cast to cmd. It shouldn't be in cmd.

cmd/service/cast_test.go
```diff
@@ -3,12 +3,13 @@ package service
 import (
 	"testing"
 
+	"github.com/mesg-foundation/core/interface/grpc/core"
```
krhubert: Same as above

PR #447, opened 2018-09-11
service: add unit tests

ilgooz:
this PR enables integration tests with build flags and and covers some of the unit tests for service package


service/dependency.go
```diff
@@ -36,7 +36,7 @@ type Dependency struct {
 // error logs.
 func (d *Dependency) Logs() (rstd, rerr io.ReadCloser, err error) {
 	var reader io.ReadCloser
-	reader, err = defaultContainer.ServiceLogs(d.namespace())
+	reader, err = d.service.docker.ServiceLogs(d.namespace())
```
antho1404: Why not injecting docker into `Dependency` like that we don't need to use `.service.docker` and we remove one dependency

ilgooz: we're reading some other _Service_ fields in other places too. I'm creating another issue to getting ridding of these cycling references. https://github.com/mesg-foundation/core/issues/475

antho1404: 👍 let's not block for that, good to track it in an issue

NicolasMahe:
Is it still WIP?

ilgooz:
@NicolasMahe I was evaluating for all techniques for unit testing after @krhubert suggestion with [testify/mock](https://github.com/mesg-foundation/core/pull/466). The original idea was to experiment with dockertest for service and api packages too and bring some knowledge from unit testing of go-service and go-application onto table and give a final decision about refactoring dockertest. But I think, I'll go with completing unit tests of service package today and give my decision about adopting which mocking/faking techniques we should go with for core.

ilgooz:
conflicts are resolved

PR #448, opened 2018-09-12
Change engine image to alpine and update to go 1.13.5

krhubert:
- Use alpine image
- Use -s -w flags for build.
- Update to go1.13.5


NicolasMahe:
TODO
- [ ] add a CI test to build the production container, start it and check if it running correcly. Actually, the e2e test should use this version instead of the local dev version.

NicolasMahe:
The `dockerfile.dev` in alpine is not working in ci e2e tests

krhubert:
Ok I have fixed dockers for dev and prod (We didn't publish the prod image for a while because it wasn't build at all).

Now there is only one dockerfile for engine

NicolasMahe:
> We didn't publish the prod image for a while because it wasn't build at all

what do you mean? the dev has been published 3 days ago and the circleci tasks are all green: https://circleci.com/gh/mesg-foundation/engine/tree/dev

Dockerfile
```diff
@@ -1,21 +1,22 @@
 # base Go image version.
-FROM golang:1.13.0-stretch AS build
+FROM golang:1.13.5-alpine AS build
```
NicolasMahe: let's force the alpine version as well
```suggestion
FROM golang:1.13.5-alpine3.10 AS build
```

Makefile
```diff
@@ -17,26 +17,26 @@ docker-build: check-version
 		-t mesg/engine:$(MAJOR_VERSION) \
 		-t mesg/engine:$(MINOR_VERSION) \
 		-t mesg/engine:$(PATCH_VERSION) \
+		-t mesg/engine:dev \
```
NicolasMahe: Are we sure we want to tag it dev? this version is only for build on the dev branch, not local and not prod.

e2e/testdata/test-complex-service/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.13
+FROM golang:1.13.5-alpine
```
NicolasMahe:
```suggestion
FROM golang:1.13.5-alpine3.10
```

e2e/testdata/test-service/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.13
+FROM golang:1.13.5-alpine
```
NicolasMahe:
```suggestion
FROM golang:1.13.5-alpine3.10
```

scripts/dev.sh
```diff
@@ -53,15 +53,15 @@ function start_engine {
...
+    --label com.docker.stack.image=mesg/engine:dev \
     --env MESG_NAME=$MESG_NAME \
     --mount type=bind,source=/var/run/docker.sock,destination=/var/run/docker.sock \
     --mount type=bind,source=$MESG_PATH,destination=/root/.mesg \
     --network $MESG_NAME \
     --network name=$MESG_TENDERMINT_NETWORK \
     --publish $MESG_SERVER_PORT:50052 \
     --publish $MESG_TENDERMINT_PORT:26656 \
-    mesg/engine:local
+    mesg/engine:dev
```
NicolasMahe: I would definitely prefer to build a local version that have a different tag than the one on docker hub (latest, dev and vX.X.X).
We actually use the dev version as a nighly or pre-prod. Maybe we should rename it to avoid confusion.

krhubert:
We spoke also to not care about the mount and build optimization to have only one Dockerfile (the prod one)

obraz

I measure it on my computer and it rebuilds in 23 seconds. If it takes 1 minute on your computer then we might just create a separate dockerfile with mounting.

krhubert:
Also, we can't use volume/mount during the build, so we can't do more about it (only static linking, but it's tricky)

NicolasMahe:
closing in favor of https://github.com/mesg-foundation/engine/pull/1718

PR #465, opened 2018-09-15
Basic stuff to start working on network part

krhubert:
Support yaml in config
Remove config.Path set in ci
Typo in docs
Run service from plugins on core start
Add plugin to config
Deploy service from path + small refactor
Make database path configurable
Start core without docker

Not all of above features were been tested, but you can leave the comments


krhubert:
P.S.

This is fork of fix/cli-logs, pls merge it first

antho1404:
would be great to have this in multiple PR

antho1404:
`core/main.go` should be merged as a command from the cli `commands/root.go`. I have the fix already I can commit it. Also deleted the `dev-core` script

commands/service_execute.go
```diff
@@ -36,33 +33,36 @@ func newServiceExecuteCmd(e ServiceExecutor) *serviceExecuteCmd {
...
+	c.cmd.Flags().StringVarP(&c.jsonFile, "json-file", "f", c.jsonFile, "Path to a JSON file containing the data required to run the task")
+
+	c.cmd.MarkPersistentFlagRequired("task")
 	return c
 }
 
 func (c *serviceExecuteCmd) preRunE(cmd *cobra.Command, args []string) error {
-	if cmd.Flag("data").Changed && cmd.Flag("json").Changed {
-		return errors.New("Only one of '--data' or '--json' options can be specified")
+	if cmd.Flag("json").Changed && cmd.Flag("json-file").Changed {
```
antho1404: What's the reason behind this modification ? the `--data` is really good so why removing it in favor of a raw json (which is really unfriendly). I would remove this change

config/config.go
```diff
@@ -2,30 +2,43 @@ package config
...
+	Path string
+	ID   string
+}
 
 // Config contains all the configuration needed.
 type Config struct {
 	Server struct {
 		Address string
+
+		Plugins []plugin
```
antho1404: I'm still not convinced using a plugin with a path. Also I was thinking more something like that:
```
Plugin struct {
  Resolver string
}
```

Like that we can identify each plugins

config/config.go
```diff
@@ -36,55 +49,88 @@ type Config struct {
...
 	}
 	return instance, nil
 }
 
-// Load reads config from environmental variables.
-func (c *Config) Load() {
+// Load reads config from yaml and env.
+// Note that env variables have higher precedence then yaml config.
+func (c *Config) Load(file string) error {
+	if err := c.LoadFromYaml(file); err != nil {
```
antho1404: Didn't we say that for now we will just keep it simple with the support of env variables only and we can always add a `.envrc` if we want to have a "config file". Moreover I don't see where this function is used.

commands/service_execute.go
krhubert: The problem with --data is casting method. We need to know service definition in order to cast data value to proper types. With service definition spread over the network, we need to download service definition first in order to cast the data.
It is complicated right now for writing PoC and moreover for feature usage. I spoke with @NicolasMahe about it and we agree to change data to JSON.

config/config.go
krhubert: I did ID and Path in plugin struct because of new users - they don't want to "hack" with id. Even for me, it would be easier to develop - just giving the path and edit code and I don't need to think about what I have changed in each plugin or core etc...

> Like that we can identify each plugins

yes, but we can easily do it in the code and by giving slice of plugins with name it gives us more flexibility for next plugins (we don't need to change the config). Moreover, we don't limit plugins that someone may start with the core.

In the feature our plugins may start with special prefix or we can even come up with diffrent apporach and resolver is good name so I don't want to take it from users ([Don't steal good names from the user.](https://blog.golang.org/package-names)). 

Even we were not sure about the name (resolver vs discover).

config/config.go
krhubert: > Didn't we say that for now we will just keep it simple with the support of env variables only and we can always add a .envrc if we want to have a "config file".

Yes, we did. We said also that in the future we will have yaml config. I have added yaml parser to support Plugin slice - passing such complicated struct in env is less readable. Envconf support slices but I guess it's only for basic go types and not for the structs.

> Moreover I don't see where this function is used.

That's true. It's not used yet - it will be if we want to.

NicolasMahe: I agree with Antho, let's keep it simple for now. So as we don't load any config file, why keeping the function.

NicolasMahe:
To store the database in a nice dedicated folder, we could use another folder than the `$HOME`.
If find [this package](https://github.com/shibukawa/configdir) that exposes the "User level configuration folder" (eg on Mac: `${HOME}/Library/Application Support`). I think it's way better to save sensitive data in this folder rather than just `$HOME/.mesg`

api/api.go
```diff
@@ -1,18 +1,22 @@
...
 	container *container.Container
 }
 
 // Option is a configuration func for MESG.
 type Option func(*API)
 
 // New creates a new API with given options.
-func New(options ...Option) (*API, error) {
-	a := &API{}
+func New(db *database.ServiceDB, options ...Option) (*API, error) {
```
ilgooz: lets pass _db_ as an _Option_ instead. and check if db is not nil and return an error if needed after options are applied.

database/service_db.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		return nil, err
+	}
+
+	return db.unmarshal(b)
+}
+
+// Save stores service in database.
+func (db *ServiceDB) Save(s *service.Service) error {
+	if s.ID == "" {
```
ilgooz: what this validation for? having a service without ID is not a valid case and it'd be a programmer error.

NicolasMahe:
Commands execute flags:
- [ ] comment --data
- [ ] add `--json-data` flag

api/api.go
krhubert: I don't want to, because it is required option and not an optional one. here you have validation in compile time and not runtime.

commands/root.go
```diff
@@ -42,3 +45,39 @@ func (c *rootCmd) persistentPreRun(cmd *cobra.Command, args []string) {
...
+
+	for _, plugin := range cfg.Server.Plugins {
+		if plugin.Path != "" {
+			if _, err := service.NewFromPath(plugin.Path); err != nil {
+				return err
+			}
+		}
+	}
+
+	tcpServer := &grpc.Server{
```
ilgooz: initializing of core should be made inside package core. _core package_ can have a func named _New_. It can have _Start_ and _Close_ methods. This way we can simply create a binary from a main program that uses _core package_ to start core and we can distribute that binary to world. or/and we can use _core package_ from _mesg-core_ cli to start core with a command.

NicolasMahe:
This pull request has been mentioned on **MESG**. There might be relevant details there:

https://forum.mesg.com/t/trustful-network/35/1

antho1404:
@krhubert should we close this PR ? we have most of the dev done in some other PR

PR #466, opened 2018-09-15
Create docker client mock

krhubert:


ilgooz:
_dockertest_ provides functionalities to test code written in both sync and async syntax where testify's mock only useful for testing sync code. Meaning by async syntax please see go-application's event-oriented package API to get an idea. 

Tests that are hangs out are totally due to programmer error but I understand and agree on the difficulty of having this kinda possibility while creating new tests. We can drop the async testing support to avoid this case after being sure of that we don't have any code logic that cannot be synchronised and we're sure that we'll not have any code in async syntex, in future. 

I'm strongly against of directly using testify/mock while creating unit tests, even if we use it we should still keep something similar to these _ProvideX_ and _LastX_ methods. These APIs provides a nice abstraction and they're more user friendly while creating new tests. For example while writing unit tests for api or service package I don't like to use low level API's from testify/mock.

As a result I don't see any advantage of using testify/mock other then solving tests that are hangs out but this can be achived by dockertest too by removing async testing support. The reason of why we use these channels is to make async testing possible but if programmer misuses them it can block the tests. The reason why I adopted this method is go-application only can be tested this way due to its async API. And I followed the same pattern in here core too but it seems that having this async testing feature doesn't make sense for core. I might consider to drop this async testing feature of dockertest which can reduce to complexity while writing tests for complicated features of core.

We can close this PR and open an issue about reducing complexity on dockertests and it can be assigned to me.

antho1404:
some example based on our code and how it can benefits compare to the mock that we have could be nice. Now we already have some mock so we need a good reason to change it.

I like the idea to use mock. I definitely prefer that is already known by many people instead of a custom mocking system. It's always better to use existing standards but I would love to have a comparison with an actual test before and after this.

I think the best use case would be the delete network in the container package. This one need to do some calls to docker to delete the network and also to listen to the stream of events from docker. I think it's the most complex use that we have right now.

https://github.com/mesg-foundation/core/blob/3dbd069af961501674e0b5fc641f6e79ff9728c7/container/network_test.go#L70-L95

ilgooz:
@antho1404 I also would like to see _testify/mock_ for unit test equivalent of [this](https://github.com/mesg-foundation/core/blob/233be30/service/start_test.go#L95) integration test. Because it is a complex tests.

krhubert:
```go
func TestDeleteNetwork(t *testing.T) {
  var (
    namespace = []string{"namespace"}
    m         = dockermock.NewCommonAPIClientMock()
    id        = "id"
  )

  // New() calls - can be extract to setupNewClient(t *testing.T) for other tests
  m.On("NegotiateAPIVersion", mock.Anything).Once()
  m.On("Info", mock.Anything).Once().
    Return(types.Info{}, nil)
  m.On("SwarmInit", mock.Anything, mock.Anything).Once().
    Return("", nil)
  m.On("NetworkInspect", mock.Anything, mock.Anything, mock.Anything).Once().
    Return(types.NetworkResource{}, nil)
  m.On("NetworkCreate", mock.Anything, mock.Anything, mock.Anything).Once().
    Return(types.NetworkCreateResponse{}, nil)
  c, _ := New(ClientOption(m))

  // DeleteNetowrk() calls
  m.On("NetworkInspect", mock.Anything, Namespace(namespace), types.NetworkInspectOptions{}).Once().
    Return(types.NetworkResource{ID: id}, nil)
  msgC := make(chan events.Message, 1)
  msgC <- events.Message{ID: id} 
  errC := make(chan error)
  m.On("Events", mock.Anything, mock.Anything).
    Return((<-chan events.Message)(msgC), (<-chan error)(errC))
  m.On("NetworkRemove", mock.Anything, id).
    Return(nil)
  require.Nil(t, c.DeleteNetwork(namespace))

  m.AssertExpectations(t)
}
```

This is one with DockerMock usage.

antho1404:
I kind of like it, this example is a bit longer but still really easy to understand and we can definitely extract some parts in other functions (the `setupNewClient` function you propose).

I like the event driven approach, the `On().Once().Return()`

ilgooz:
I'm also going to do some more experimenting on _testify/mock_ and hopefully send some feedbacks today.

ilgooz:
experimenting with https://github.com/mesg-foundation/core/pull/506

ilgooz:
I'm pretty happy with _testing/mock_ as I experiment it creating unit tests for _service_ package. With _testing/mock_ we can conditional have return values depending on different arguments for same methods. This is a very important thing and needed a lot for our cases, while testing complex methods.

Asserting method call counts and arguments are also nicely covert with _testing/mock_ and can be made in a simple way.

NicolasMahe:
What should we do with this PR?

Mocking docker is useful only to test really quickly that we didn't break our codebase. But anyway, it will be tested against the real docker engine in integration test.

I feel it's a lot of code for unit testing and we may want to run integration test to be sure.

krhubert:
@NicolasMahe the point of this package was to slowly replace dockertest and I think we will eventually do this (maybe not now) we can close this PR (and reopen if needed), merge it and start using in new tests or keep it open

NicolasMahe:
Why not generating the mock from the interface with [Mockery](https://github.com/vektra/mockery)?

krhubert:
@NicolasMahe It was before I know about mockery

PR #473, opened 2018-09-17
commands: [fix] use go chans correctly & correct err checking

ilgooz:
* fixed this [line](https://github.com/mesg-foundation/core/pull/473/files#diff-c7a4b2a3a589820b87a05a160af2f76bL50) because err already checked [here](https://github.com/mesg-foundation/core/pull/473/files#diff-c7a4b2a3a589820b87a05a160af2f76bL62)

* putting a [default](https://github.com/mesg-foundation/core/pull/473/files#diff-c7a4b2a3a589820b87a05a160af2f76bL65) here will break the loop even if a goroutine gets late to send a value to _errC_. because default will triggered right away if there is nothing being sended in the senders side.


krhubert:
> fixed this line because err already checked here

true, in that case just remove nil check from select statement

> putting a default here will break the loop even if a goroutine gets late to send a value to errC. because default will triggered right away if there is nothing being sended in the senders side.

this is handled by wg.WaitGroup.

ilgooz:
@krhubert right, I agree with the second one as well. would you like to have a fix on err check only? or lets merge this PR as it is because behavior is the same and the code is less complex.

ilgooz:
@krhubert I don't get the _xerrors.Error_ part, could you explain more? for the second part please check std lib [example](https://github.com/golang/go/blob/master/src/net/http/transport.go#L1177)

commands/provider/core_provider.go
```diff
@@ -32,38 +31,30 @@ func (p *CoreProvider) Start() error {
...
+		errC   = make(chan error, idsLen)
+	)
 
 	for _, id := range ids {
 		go func(id string) {
-			defer wg.Done()
-			if _, err := p.client.StopService(context.Background(), &coreapi.StopServiceRequest{
+			_, err := p.client.StopService(context.Background(), &coreapi.StopServiceRequest{
 				ServiceID: id,
-			}); err == nil {
```
antho1404: This was definitely a bug, we add the err (which is nil in the `errC`) I guess it's related to what @krhubert was talking about with the nil in the error channels but I don't quite understand what was the point. If you can explain a bit more that would be great.

PR #477, opened 2018-09-19
Bind core volume

NicolasMahe:
This PR brings back the bind of the core volume to a directory on the host machine.

The reason is to remove the dependency to Docker's volume for the Core.
The services are still using Docker volumes.

### DIFF
--- a/database/services/db.go
+++ b/database/services/db.go
@@ -16,7 +16,11 @@ func open() (db *leveldb.DB, err error) {
 	instanceMutex.Lock()
 	defer instanceMutex.Unlock()
 	if _instance == nil {
-		storagePath := filepath.Join(config.Path, "database", "services")
+		c, err := config.Global()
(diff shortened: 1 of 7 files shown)


database/services/db.go
```diff
@@ -16,7 +16,11 @@ func open() (db *leveldb.DB, err error) {
 	instanceMutex.Lock()
 	defer instanceMutex.Unlock()
 	if _instance == nil {
-		storagePath := filepath.Join(config.Path, "database", "services")
+		c, err := config.Global()
+		if err != nil {
+			return nil, err
+		}
+		storagePath := filepath.Join(c.Core.Path, "database", "services")
```
ilgooz: maybe keep this path in config package as well?

antho1404: not related to this PR, we can do that in a separate issue/pr maybe the #481

Issue #457, opened 2018-09-12
make Path in config package configurable


ilgooz:
otherwise we cannot run our tests in dev machines, _Path_ is configured to match with docker volume

NicolasMahe:
Done in PR https://github.com/mesg-foundation/core/pull/478

PR #478, opened 2018-09-19
Multiple core runnable

antho1404:
We can now sandbox the core in it's own context, the `.mesg` folder contains now a folder for each core based on the name. The core now has all the docker configurations (service, container, network) based on its name and we can now so run multiple core with some env variable.

```
./dev-core
MESG_CORE_NAME=testx MESG_SERVER_ADDRESS=:50201 ./dev-core
```

and access with the cli 
```
./dev-cli service list
MESG_CLIENT_ADDRESS=localhost:50201 ./dev-cli service dev list
```


NicolasMahe:
This close #457

NicolasMahe:
To do:
- [ ] the command `stop` is stopping mesg services from any core. We should use another system than listing all docker service that have a mesg.hash (add the name of the core as well?)
https://github.com/mesg-foundation/core/blob/bf15f80105e8c1b212098ba7de5175a8c584989f/service/status.go#L72-L87

daemon/logs.go
```diff
@@ -6,5 +6,5 @@ import (
 
 // Logs returns the core's docker service logs.
 func Logs() (io.ReadCloser, error) {
-	return defaultContainer.ServiceLogs(Namespace())
+	return defaultContainer.ServiceLogs([]string{})
```
ilgooz: I think container package should be more agnostic. maybe it is better to not access config package inside container. the thing with passing empty _Namespace()_ is kinda weird to me

NicolasMahe: I agree with you, `container.Namespace` function should not access the config to add the Core.Name as prefix. It should be pass the Namespace every time we need it. We could create an helper function in config package to get it simply.

antho1404: what do you think about doing that in another PR / issue. I don't want to make this one overcomplicated. There is already some complicated parts in this PR

config/config.go
```diff
@@ -37,8 +37,9 @@ type Config struct {
 	}
 
 	Core struct {
-		Image string
-		Path  string
+		Image    string
```
krhubert: I don't get this part, why do we need name and rootpath? 
What rootpath means - consider remaining it to ConfigPath as this is what you mean right?
Also when the path is explicit given the app shouldn't create new directories (`.mesg`) under it.

daemon/logs.go
ilgooz: https://github.com/mesg-foundation/core/pull/478#pullrequestreview-157643597

config/config.go
antho1404: The goal of that is to have the database stored in it's own directory so the root path is the one that will store all the different "core" instances

Right now we have 
```
~/.mesg/
├── database
```

And this change to
```
~/.mesg/
├── mesg-core/
|   ├── database
```
Like that we can now run different instances of the core with their own separated database
```
~/.mesg/
├── mesg-core1/
|   ├── database
├── mesg-core2/
|   ├── database
...
```

The `RootPath` is the one to store the `.mesg` and the `CorePath()` is the one to use to know where to write. This one returns the `RootPath + Name`.

`ConfigPath` is a bit ambiguous because it's the config of the core so this should be the `~/.mesg/mesg-core`

> Also when the path is explicit given the app shouldn't create new directories (.mesg) under it.

Why shouldn't we do that ? The goal of this property is to be the "namespace" for the folder

antho1404: Based on the discussion on discord and the feedbacks of @krhubert we can remove the `RootPath` and just keep a `CorePath` that is customizable by the user and so we would just have the `MESG_CORE_PATH` to edit more

antho1404:
@mesg-foundation/core ready to review again, I added the `MESG_CORE_PATH` to run the core with different config/db repository like that we can run multiple core with their own database or even choose to run multiple core with the same database

PR #484, opened 2018-09-20
Add multiline option to Pretty.ColorizeJSON

NicolasMahe:
Very simple PR to add the multiline option to Pretty.ColorizeJSON function.
It is useful for a following PR.

### DIFF
--- a/utils/pretty/pretty.go
+++ b/utils/pretty/pretty.go
@@ -215,7 +215,7 @@ func (p *Pretty) ColorizeJSON(keyColor *color.Color, valueColor *color.Color, mu
 	}
 
 	f := prettyjson.NewFormatter()
-	if false == multiline {
+	if !multiline {
 		f.Indent = 0
 		f.Newline = ""
 	}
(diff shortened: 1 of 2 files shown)


utils/pretty/pretty.go
```diff
@@ -209,14 +209,16 @@ func (p *Pretty) Colorize(c *color.Color, msg string) string {
...
-func (p *Pretty) ColorizeJSON(keyColor *color.Color, valueColor *color.Color, data []byte) []byte {
+func (p *Pretty) ColorizeJSON(keyColor *color.Color, valueColor *color.Color, multiline bool, data []byte) []byte {
 	if p.noColor {
 		return data
 	}
 
 	f := prettyjson.NewFormatter()
-	f.Indent = 0
-	f.Newline = ""
+	if multiline == false {
```
krhubert: !multiline

NicolasMahe: Updated

krhubert: It's go way:

```
➜  a git:(dev) ✗ gometalinter .
main.go:7:5:warning: should omit comparison to bool constant, can be simplified to !a (S1002) (megacheck)
```

Also go source code:

```
➜  src gogrep 'if !' |wc -l 
8661
➜  src gogrep 'if.*== false' |wc -l
17
➜  src 
```

antho1404: I don't like the `false == multiline` it's more "conventional but we never do that anywhere in the code. I think it's better to have consistency. Or if you want just change the parameter with `inline` like that you have a `if inline` ;)

NicolasMahe: Ok guys, I give up on this ;)

PR #486, opened 2018-09-20
commands: reimplements printing deployment statuses, closes #462

### LINKED ISSUE #462: commands: missing deploy status messages
we [used to have](https://github.com/mesg-foundation/core/blob/33fb3a31bb84eeee8ade9b1b04f193cddfb56bd7/cmd/service/deploy.go#L128-L136) some lines in [this file](https://github.com/mesg-foundation/core/blob/401f8f9/commands/provider/service_deployer.go) that prints out deploy statuses. 

related with https://github.com/mesg-foundation/core/pull/444

ilgooz:
* utils/pretty: add UseSpinner() and DestroySpinner() funcs.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -23,7 +23,7 @@ type ServiceExecutor interface {
 	ServiceByID(id string) (*coreapi.Service, error)
 	ServiceDeleteAll() error
 	ServiceDelete(ids ...string) error
-	ServiceDeploy(path string) (id string, valid bool, err error)
+	ServiceDeploy(path string, statuses chan provider.DeployStatus) (id string, validationError, err error)
 	ServiceListenEvents(id, eventFilter string) (chan *coreapi.EventData, chan error, error)
 	ServiceListenResults(id, taskFilter, outputFilter string, tagFilters []string) (chan *coreapi.ResultData, chan error, error)
 	ServiceLogs(id string, dependencies ...string) (logs []*provider.Log, closer func(), err error)
--- a/commands/provider/service_deployer.go
+++ b/commands/provider/service_deployer.go
@@ -14,8 +14,10 @@ import (
 type StatusType int
 
 const (
+	_ StatusType = iota // skip zero value.
+
 	// RUNNING indicates that status message belongs to an active state.
-	RUNNING StatusType = iota + 1
+	RUNNING
 
 	// DONE indicates that status message belongs to completed state.
 	DONE
--- a/commands/provider/service_deployer.go
+++ b/commands/provider/service_deployer.go
@@ -2,11 +2,13 @@ package provider
 
 import (
 	"context"
+	"fmt"
 	"io"
 
 	"github.com/asaskevich/govalidator"
 	"github.com/docker/docker/pkg/archive"
 	"github.com/mesg-foundation/core/protobuf/coreapi"
+	"github.com/mesg-foundation/core/utils/pretty"
 )
 
 // ServiceDeploy deploys service from given path.
@@ -87,17 +89,31 @@ func readDeployReply(stream coreapi.Core_DeployServiceClient, deployment chan de
 		}
 
 		var (
+			status          = message.GetStatus()
 			serviceID       = message.GetServiceID()
 			validationError = message.GetValidationError()
 		)
 
 		switch {
+		case status != nil:
+			switch status.Type {
+			case coreapi.DeployServiceReply_Status_RUNNING:
+				pretty.UseSpinner(status.Message)
+
+			case coreapi.DeployServiceReply_Status_DONE:
+				pretty.DestroySpinner()
+				fmt.Println(status.Message)
+			}
+
 		case serviceID != "":
+			pretty.DestroySpinner()
 			result.serviceID = serviceID
 			deployment <- result
 			return
 
 		case validationError != "":
+			pretty.DestroySpinner()
+			fmt.Println(pretty.Fail(validationError))
 			result.isValid = false
 			deployment <- result
 			return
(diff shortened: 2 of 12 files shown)


commands/provider/service_deployer.go
```diff
@@ -87,17 +89,31 @@ func readDeployReply(stream coreapi.Core_DeployServiceClient, deployment chan de
 		}
 
 		var (
+			status          = message.GetStatus()
 			serviceID       = message.GetServiceID()
 			validationError = message.GetValidationError()
 		)
 
 		switch {
+		case status != nil:
```
krhubert: This one of the reason why I drop it in refactoring. Only commands package is responsible for input/output for cli. 

Please move all output printing/spinner usage to commands package to do it properly.

ilgooz: I agree 👍

commands/service_dev.go
```diff
@@ -48,21 +50,19 @@ func (c *serviceDevCmd) preRunE(cmd *cobra.Command, args []string) error {
...
 func (c *serviceDevCmd) runE(cmd *cobra.Command, args []string) error {
-	var (
-		id    string
-		valid bool
-		err   error
-	)
-	pretty.Progress("Deploying the service...", func() {
-		id, valid, err = c.e.ServiceDeploy(c.path)
-	})
+	statuses := make(chan provider.DeployStatus, 0)
```
krhubert: Small change, but if you're putting `0`, just skip it and write `make(chan provider.DeplyStatus)`

```warning: should use make(chan provider.DeployStatus) instead (S1019) (megacheck)```

File: commands/service_deploy.go [hunk 3]
```diff
@@ -4,6 +4,9 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/mesg-foundation/core/x/xerrors"
+
```
krhubert: remove empty line

commands/commands.go
```diff
@@ -23,7 +23,7 @@ type ServiceExecutor interface {
 	ServiceByID(id string) (*coreapi.Service, error)
 	ServiceDeleteAll() error
 	ServiceDelete(ids ...string) error
-	ServiceDeploy(path string) (id string, valid bool, err error)
+	ServiceDeploy(path string, statuses chan provider.DeployStatus) (id string, validationError, err error)
```
krhubert: Update commands_mock_test.go with new api

commands/service_deploy.go
```diff
@@ -38,21 +41,32 @@ func (c *serviceDeployCmd) preRunE(cmd *cobra.Command, args []string) error {
...
+	statuses := make(chan provider.DeployStatus, 0)
+	go printDeployStatuses(statuses)
+	id, validationError, err := c.e.ServiceDeploy(c.path, statuses)
 	if err != nil {
 		return err
 	}
-	if !valid {
-		return errors.New("Service is invalid. To get more information, run: mesg-core service validate")
+	if validationError != nil {
+		pretty.DestroySpinner()
```
krhubert: Here just wonder, shouldn't deployer return `DONE` DeployedStatus on an error, it's kind of done but with error.

Because Here stoping spinner that hasn't been started in the same function looks suspicious.

The same situation is in service_dev.go

commands/provider/service_deployer.go
```diff
@@ -2,41 +2,68 @@ package provider
...
 	"github.com/docker/docker/pkg/archive"
 	"github.com/mesg-foundation/core/protobuf/coreapi"
 )
 
+// StatusType indicates the type of status message.
+type StatusType int
+
+const (
+	// RUNNING indicates that status message belongs to an active state.
+	RUNNING StatusType = iota + 1
```
krhubert: One small change - use `iota` - there is no obvious reason why counting should be started from 1.

commands/commands.go
ilgooz: done

commands/service_deploy.go
ilgooz: done

commands/service_deploy.go
ilgooz: i had the similar thing in my mind but i wasn't sure to implement this because this is not an exception to validation error. an error may occour at any time during deployment and because of this we may need to destroy spinner unexpectedly. actually, I should move this line to out of _validationError_'s nil check to make sure we call it all the time, thanks for the heads up.

commands/provider/service_deployer.go
ilgooz: in goi it's a best practice to count constants from 1. i guess this due to 0 is a _zero_ value in go

krhubert: Can you provide some article/best practice about this? I want to read about it.
thanks

commands/service_deploy.go
krhubert: What about sending DONE to deploy statuses?

ilgooz: i think errors should be only returned from funcs otherwise we'll have duplication of errors in the place where they're handled. feel free to open an issue about this so we can discuss there.

commands/provider/service_deployer.go
ilgooz: https://github.com/golang/go/wiki/Iota

antho1404: We could add a status `UNKNOWN` and start at `iota` like that it's ok. I agree that we shouldn't start at 0, we can have some side effects with that.

Like here https://github.com/mesg-foundation/core/blob/a6887f298acf9e1e1d09f6b853d44a660b3b3a46/service/status.go#L11-L17

or here

https://github.com/mesg-foundation/core/blob/a6887f298acf9e1e1d09f6b853d44a660b3b3a46/container/type.go#L7-L12

Issue #445, opened 2018-09-11
add CI task to run integration tests when dev branch is updated


ilgooz:
`go test ./... -tags=integration`

PR #489, opened 2018-09-21
Add integration test CI task and fix integration test

NicolasMahe:
This PR was meant to be only to add integration test on CircleCI but as integration test failed I had to fix them.

Closes https://github.com/mesg-foundation/core/issues/445


NicolasMahe:
To do:
- [ ] create 3 workflows:
  - [ ] one for PR: only unit testing
  - [ ] one for dev branch: integration test + dev release
  - [ ] one for master branch: integration test + prod release
- [ ] Add cache system: https://medium.com/misc-tech/making-your-build-faster-with-go-1-10-on-circleci-2-0-915fea0ef867
- [ ] Share locally build image between tasks

.circleci/config.yml
```diff
@@ -25,30 +25,33 @@ runOnDocker: &runOnDocker
...
+  "unit_test":
+    <<: *runOnDocker
     steps:
       - checkout
-      - run: docker pull mesg/core:latest
-      - run: docker build -t mesg/core:$CIRCLE_SHA1 --build-arg version="dev build `echo $CIRCLE_SHA1 | cut -c1-7`" .
-
-  "test":
+      - setup_remote_docker # this should be remove when no test are actually using docker
+      - run: go get -t ./...
```
krhubert: Remove go get

.circleci/config.yml
```diff
@@ -25,30 +25,33 @@ runOnDocker: &runOnDocker
...
+      - run: docker swarm init # this should be remove when no test are actually using docker
+      - run: docker build -t sleep docker-images/sleep/ # this should be remove when no test are actually using docker
+      - run: docker build -t http-server docker-images/http-server/ # this should be remove when no test are actually using docker
+      - run: env MESG_CORE_IMAGE=mesg/core:NONEXISTINGIMAGE go test -timeout 180s -p 1 -coverprofile=coverage.txt ./...
```
krhubert: What about syntax:
```
run:
  name:
  command
  envrioment:
```
I think is more clear than putting everything in one line

.circleci/config.yml
```diff
@@ -25,30 +25,33 @@ runOnDocker: &runOnDocker
...
+      - run: bash <(curl -s https://codecov.io/bash)
```
krhubert: We're not using `codecov` - maybe think of removing it and include when we actually start using this in our dev pipeline

.circleci/config.yml
antho1404: yes good point like that we will directly see if we forgot to put some stuff on the vendor folder

container/network.go
```diff
@@ -34,7 +34,10 @@ func (c *Container) CreateNetwork(namespace []string) (id string, err error) {
 }
 
 // DeleteNetwork deletes a Docker Network associated with a namespace.
-func (c *Container) DeleteNetwork(namespace []string) error {
+// event parameter can be "destroy" or "remove". If the network was used by a service, the event to use is "destroy". If the network has not been used, the event is "remove".
+// Remove removes the reference from Docker to the network.
+// Destroy removes the network from Docker active network.
+func (c *Container) DeleteNetwork(namespace []string, event string) error {
```
krhubert: Just a thought, maybe listen to both of these events inside the method. In that case, a developer doesn't need to know about internal events of docker.

NicolasMahe: I tried and it doesn't work because the events "destroy" and "remove" have no definitive order and also "destroy" is fired only when the network has been recently (few seconds) attached to a container...
This is not a perfect solution but the perfect solution requires to reorganize the container package to match our logic (mesg's service have multiple docker's service, one network) rather than staying close to the docker logic (service, container, network). By doing this, we could extract a lot of docker related logic from the service package to the container package (like start and stop mesg's service with dependencies and network).

container/service_integration_test.go
```diff
@@ -54,38 +54,38 @@ func TestIntegrationStopNotExistingService(t *testing.T) {
 	require.Nil(t, err)
 	namespace := []string{"TestStopNotExistingService"}
 	err = c.StopService(namespace)
-	require.Nil(t, err)
+	require.NotNil(t, err)
```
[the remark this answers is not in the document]
NicolasMahe: I will do it in another PR. There is many places to replace `require.Nil(t, err)` and `require.NotNil(t, err)` by `require.NoError(t, err)` and `require.Error(t, err)`.
Also, one test file is still using `stvp/assert`.

.circleci/config.yml
NicolasMahe: I would like to keep it as a reminder of the code coverage stats.
It's true we don't enforce it strictly, but still we should have the stats in the PR.

container/service_integration_test.go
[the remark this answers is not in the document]
krhubert: Sure, I didn't want to do everything at once. Just to make sure new assertions have more meaningful errors messages. But it could be one big simple PR as well

container/network.go
krhubert: Ok, got it

.circleci/config.yml
NicolasMahe: We don't have the env variable anymore and CircleCi doesn't provide to anyway to pass arguments through a nice yaml list.
I'm ok to read the command and not a short name.
What do you think @krhubert?

NicolasMahe:
@mesg-foundation/core ready to review

container/service_integration_test.go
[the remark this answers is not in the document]
antho1404: why this test changed in the first place ? The stop should not return an error if the service doesn't exists. If it doesn't exists it means that it's not running. Agree that we should also have a kind of NotFound error just want to make sure that this change in the test really needs to be done

.circleci/config.yml
```diff
@@ -81,27 +57,44 @@ jobs:
...
             - "publish_docker_dev"
+
+  release_prod:
+    jobs:
+      - "test":
+          filters:
+            tags:
+              only: /^v.*/
+            branches:
+              only: master
```
antho1404: The tests here will be executed only when we are on master AND we have a tag. In this case we skip all the tests when we merge to master and these are important.

We should add the check on the tag only for the `publish_docker_prod` and `release_cli_prod` not for the test

container/network.go
```diff
@@ -8,6 +8,15 @@ import (
 	docker "github.com/docker/docker/client"
 )
 
+// NetworkEventType is a type to define the kind of event related to the network
+type NetworkEventType string
+
+// List of Network status to listen to confirm a network deletion
+const (
+	NetworkRemove  NetworkEventType = "remove"
+	NetworkDestroy NetworkEventType = "destroy"
```
krhubert: Small change, but in go the common rule is to put a type of variable only in the first line.

Also I'd remove `Network` prefix and stick to `type EventType string`, following:

```
const (
  EventRemove EventType = "remove"
  // ...
)
```

Check out https://github.com/moby/moby/blob/master/api/swagger.yaml#L7138 -> the name of events are the same for different docker objects

antho1404: Good point, fixing this

container/network.go
```diff
@@ -34,7 +43,10 @@ func (c *Container) CreateNetwork(namespace []string) (id string, err error) {
 }
 
 // DeleteNetwork deletes a Docker Network associated with a namespace.
-func (c *Container) DeleteNetwork(namespace []string) error {
+// event parameter can be "destroy" or "remove". If the network was used by a service, the event to use is "destroy". If the network has not been used, the event is "remove".
+// Remove removes the reference from Docker to the network.
+// Destroy removes the network from Docker active network.
+func (c *Container) DeleteNetwork(namespace []string, event EventType) error {
```
ilgooz: it's a bit odd to me having _event_ as an argument to this func because it does not change the behavior of network removing. we can listen for both events and return after the first received event. this way use of this func will be agnostic and caller will not need to figure out which event type it should pass.

antho1404: The problem with this one is that we cannot wait for both. In some cases we will have both (when we create the service and attach the network to this service) but in some other cases when we just create and destroy the network without any service we only have the "remove" event.

I agree this is kind of weird to have this `event` maybe we can rename it to `deletionEvent` or something like that.

I don't see any ways to make this function agnostic because it's really related to docker and how we use it but what thing that can be done in the future is to remove this `CreateNetwork` and `DeleteNetwork` from the public API and manage the network internally in the container package which totally makes sense.

NicolasMahe: Totally agree with the last part.
> This is not a perfect solution but the perfect solution requires to reorganize the container package to match our logic (mesg's service have multiple docker's service, one network) rather than staying close to the docker logic (service, container, network). By doing this, we could extract a lot of docker related logic from the service package to the container package (like start and stop mesg's service with dependencies and network).

ilgooz: i don't quite get the limitation of waiting for both _destory_ and _remove_ here. does docker produces two events for both _destory_ and _remove_ when we have a service attached to network?

NicolasMahe: Yes it does. But as we run integration test that only create a network and them delete it, we need a way to change the behaviour of `DeleteNetwork` otherwise we have timeout and the integration test on `CreateNetwork` doesn't work (as well as others).

NicolasMahe: That's why, after transferring more logic to the container package (where we ALWAYS create a network attached to service), we will be able to remove the event parameter.

ilgooz: I understand now. maybe we should inspect the network to see if there is any services attached to it before removing, this way we can guess the event type fore network removing without doing changes on container package. but there can be a race condition with this approach, while we're inspecting the network some services can be removed or started.

ilgooz: i think, the race condition can be prevented by also listening for service start/stop events in this func.

container/service_integration_test.go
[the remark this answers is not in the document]
NicolasMahe: I will investigate but I think the function `StopService` is not behaving the same way since it wait for actual deletion of Docker's ressources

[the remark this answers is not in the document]
NicolasMahe: I confirm the function is behaving differently since this commit https://github.com/mesg-foundation/core/commit/0366e060934827a179e417c0b6a25170ebc6b246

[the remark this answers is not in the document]
NicolasMahe: Fix in new PR: https://github.com/mesg-foundation/core/pull/505

Issue #437, opened 2018-09-07
Break internal dependencies in service


antho1404:
Event, Outputs etc... have a reference on the service and the only reason they have this is to have a nice error message. The service is never needed in the actual business logic, it's just user cosmetic so I would recommend to get rid of this because it's not really needed for the business logic and it can make things complicated in the future

related PR https://github.com/mesg-foundation/core/pull/417#pullrequestreview-153245567

ilgooz:
#475 includes this

PR #491, opened 2018-09-24
service: reduce cycling refs

ilgooz:
depends on #447
closes https://github.com/mesg-foundation/core/issues/437


ilgooz:
good to review

krhubert:
Can you explain why there was a *Sevice in the first place?

Because for me the serviceName (previous service) should be completely removed from task/event struct.

Why Task/Event have to have knwoleage of service they are in?

ilgooz:
@krhubert _Service_ info used while producing errors like [here](https://github.com/mesg-foundation/core/pull/491/files#diff-fb716fbb7bbe727b6611577fbd120d0eR114)

krhubert:
yes, but besides errors/logs?

> If the errors/logs are the only reason, then info about task/event services should be added somewhere else, not on struct level

So if for errors, then it should be removed, and error should be without ServiceName, or the service name should be add by `executor` of the task/event

About the code everthing is ok, but chaing from *Service to string is cosmetic, it should be removed at all.

ilgooz:
@krhubert I understand but the change is not for cosmetic reasons. we had these cycling references before between _Service_ type and others and now we don't keep pointers around. I don't quite agree on removing this meta data because they provide more detailed info about the errors. We can also get these meta info from the caller's level but it'll require us to create an additional error to combine all info about the error. /cc @mesg-foundation/core any more feedbacks?

ilgooz:
about the removed lines from _fromService()_ it should not introduce any bugs, they are just forgotten there, Getters sets the needed values. tests are also passes, if you see any issue please let me know. 

for keeping the parent's meta data info in the child, i think this is a grey area, even std packages introduces this kind of stuff. there should be a balance between where to do this or not. to me, this one is not "evil" and ok. but i'm ok with the both ways. we can combine the errors in the caller's side but we should not modify the returned error

antho1404:
I'm not sure I really understand the part where it's removed from the `fromService` but it's ok because the getters set the value.

What happen if we never call the getters ? In this case we will not have the service/serviceName. This can be a bit dangerous no ?

ilgooz:
the whole idea of exposing service definition from _Service_ type is a bit ambiguous to me. I'd rather having a _Definition_ struct and keep it under _Service_ type privately. While refactoring service package, I didn't choose do this to reduce the complexity of PR and leave it as it is to experiment with it a bit and later decide for an improvement. I was thinking about having a _Definition_ type where it has all these service definition and keeps the values without using pointers, this way it also prevents cycling references.

we don't have any piece of logic that depends on accessing _Service_ definitions directly instead of using Getters and I think that we'll never change. I agree with @antho1404 that it's a bit ambiguous to not have _serviceName_ when you directly access it from _Service_ type's fields. 

lets discuss what direction we should go with I suggest the first option below:
* let _FromService()_ to get a _Definition_ and have a private _def Definition_ field under _Service_ type. Keep _Service_ reference in _Task_, _Output_ etc. _(Service field will not have a reference to these types, so it'll not cause to cycling refs)_. 
* set _serviceName_ etc inside _fromService()_, optionally rm them completely.

with the first option we can also get rid of cycling ref for _Dependency_ type too.

NicolasMahe:
> * let _FromService()_ to get a _Definition_ and have a private _def Definition_ field under _Service_ type. Keep _Service_ reference in _Task_, _Output_ etc. _(Service field will not have a reference to these types, so it'll not cause to cycling refs)_.

I'm not sure to understand. `Definition` will be a new struct?

----

I suggest to get rid of any reference (`*Service` and even Service Name).
Task, Events, Parameters don't need Service in order to work properly.
For the error message, yes the message itself will not say which service it is. But we can catch this type of error and convert them to errors than have more context (like the service name or ID).

Example with the unmarshal function in database package that return a new error rather than the one returned from `json.Unmarshal` that have zero knowledge about the service id.
https://github.com/mesg-foundation/core/blob/c2527ae17bad7f57fe9b1fa8c4027c617bf214f2/database/service_db.go#L58-L64

PR #498, opened 2018-09-25
Remove warning about .mesgingnore close #494

### LINKED ISSUE #494: Remove warning about .mesgingnore
We're not using `.mesgignore` and we're still under active development. I think we don't need to have this warning anymore.

Also in some places, this warning is invalidly marked as `error`.

krhubert:
close #494


krhubert:
Done, sorry for that!

Issue #495, opened 2018-09-24
don't access config inside container package


ilgooz:
related with https://github.com/mesg-foundation/core/pull/478#pullrequestreview-157643597

NicolasMahe:
Will be finally done in PR https://github.com/mesg-foundation/engine/pull/1151 🎊

krhubert:
Nope, because part of container package was moved to sdk. So we need to update issue:

don't access config inside sdk package.

PR #499, opened 2018-09-25
extend ListServices api to filter by running services, closes #496

### LINKED ISSUE #496: core_provider.Stop() should get running services from the Core
The CLI access directly the function `service.ListRunning()`.
It should use the core API `ServiceList` to list the services and then stop them.
For now, `ServiceList` doesn't return the service's status. We can implement a first dummy version that try to stop ALL of them.
**Then in the future**, we can add the status to the `ServiceList` and stop only the running ones.

https://github.com/mesg-foundation/core/blob/c3e23fd91747f05750f9f18c699bb4910a082b7e/commands/provider/core_provider.go#L33-L66

ilgooz:
* coreapi.Service type now has IsRunning field, relevant APIs accordingly updated to set this value.
* update dev-core script.
* update database API to filter services by id, add more tests.


ilgooz:
some options about setting _isRunning_ field on _Service_ proto:
* implement _StatusType_ on Service proto definition for detailed status info or only keep _isRunning_ state.
*  we explicitly call _service.Status()_ on _GetService()_ grpc api but we don't do the same for _ListServices()_. I prefer explictily calling _service.Status()_ for _ListServices()_ too, what do you think?

api/api.go
```diff
@@ -9,18 +10,23 @@ import (
 type API struct {
 	db        *database.ServiceDB
 	container *container.Container
+	cfg       *config.Config
```
krhubert: There is already issue https://github.com/mesg-foundation/core/issues/495 that you've created, about not stroing config inside [Container](https://github.com/mesg-foundation/core/blob/dev/container/container.go#L22).

It should be passed as args/option to New func.

api/api_test.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-const testdbname = "db.test"
+const testdbname = "/tmp/db.test"
```
krhubert: What is the purpose of changing the path?

Not every OS (windows for example) has /tmp dir (or not every user has access to /tmp dir)

database/service_db.go
```diff
@@ -64,7 +65,8 @@ func (db *ServiceDB) unmarshal(id string, value []byte) (*service.Service, error
 }
 
 // All returns every service in database.
-func (db *ServiceDB) All() ([]*service.Service, error) {
+// If ids > 0, it'll only return services with matching ids.
```
krhubert: It's better to keep the functionality of `All` as it's was by original.

You can add an GetByIDs or similar method, to keep API more transparent

api/api.go
```diff
@@ -9,24 +12,32 @@ import (
...
 }
 
 // Option is a configuration func for MESG.
 type Option func(*API)
 
 // New creates a new API with given options.
-func New(db *database.ServiceDB, options ...Option) (*API, error) {
-	a := &API{db: db}
+func New(options ...Option) (*API, error) {
```
antho1404: The API should be in parameter here, it's a required arguments so it has to be here. If we call `New()` without any options (what most developers will do without knowing the codebase) we will have some crash in the code.

Of course you add the error in the end but what's the point to remove a required argument and add a error, it will just be more confusing. I like the pattern to use options but for what is optional and not the stuff required, it doesn't make sense.

Also and more important, let's keep the pull requests simple. This has nothing to do here, it's a pull request about extending the API why do we have some stuff in the database. Especially that this modifications has been done in another pull request that was just merged today and stayed open for more than a week so for this kind of things it better to talk about this in the other PR and not try to include that in a non related PR.

api/api.go
ilgooz: yes, it is better to not access config inside container package. in this case we access it from api package.

actually i'd rather doing all this initilazitions in the core package and propagate it down to initialization of _grpc server_ & _api_. this way we can reuse everything.

ilgooz:
lets discuss https://github.com/mesg-foundation/core/pull/499#issuecomment-424257774

interface/grpc/core/list_services.go
```diff
@@ -3,14 +3,53 @@ package core
...
+				}
+			}
+			errC <- err
+		}(s)
+
+	}
+
+	var errs xerrors.Errors
+
+	for i := 0; i < servicesLen; i++ {
```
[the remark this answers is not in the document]
antho1404: +1 for the wait group, I think it makes the code more readable than a for that is waiting for some events in the chan.

@krhubert I don't get the difference between the buffered and unbuffered version. I would go myself on the first one, what will be the advantage of using the second one ?

interface/grpc/core/list_services.go
```diff
@@ -3,14 +3,53 @@ package core
...
+
+	protoServices := toProtoServices(services)
+
+	var (
+		servicesLen = len(services)
+		errC        = make(chan error, servicesLen)
+	)
+
+	// fill services status info.
+	for _, s := range services {
```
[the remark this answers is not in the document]
ilgooz: done. i made it that way to have a logical separation while updating _Service_ fields in case we do set additional fields like this in future. but i agree and i'd rather reducing the complexity of code.

interface/grpc/core/list_services.go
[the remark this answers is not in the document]
ilgooz: thanks, i cannot see an advantage of using _WaitGroup_ here except using _range_ syntax while receiving from _errC_ which makes readability slightly better. also check std package [example](https://github.com/golang/go/blob/master/src/net/http/transport.go#L1175-L1177).

lets give a decision about which way to go. i'm neutral about this, i see that using _range_ has an upside for readability but it also introduces more code.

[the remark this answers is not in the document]
krhubert: The example you gave it's slightly different. Here they don't wait for goroutines to finish but provide [timeout](https://github.com/golang/go/blob/master/src/net/http/transport.go#L1164) for TLSHandshake execution. WIthout timeout, there will be no point to put TLSHandshake inside goroutine.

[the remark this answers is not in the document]
NicolasMahe: I also suggest to use a WaitGroup to close the chan and use a simple `range` to iterate over the errors.
I suggest to avoid the classic `for i := 0; i < len(arr); i++ {` and anything that directly deal with index on array. It's a source of error.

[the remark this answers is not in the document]
ilgooz: do three of you have an agreement on this? please like this comment, i'll switch to _WaitGroup_ if i see likes >= 2. :)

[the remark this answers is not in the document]
krhubert: About the second example, the code is inside the test package and it's quite old. 

```
8e69d43b32b src/net/net_test.go     (Brad Fitzpatrick 2016-09-27 20:50:57 +0000 451)    for i := 0; i < 2; i++ {
8e69d43b32b src/net/net_test.go     (Brad Fitzpatrick 2016-09-27 20:50:57 +0000 452)            if err := <-errc; err != nil {
8e69d43b32b src/net/net_test.go     (Brad Fitzpatrick 2016-09-27 20:50:57 +0000 453)                    t.Fatal(err)
8e69d43b32b src/net/net_test.go     (Brad Fitzpatrick 2016-09-27 20:50:57 +0000 454)            }
8e69d43b32b src/net/net_test.go     (Brad Fitzpatrick 2016-09-27 20:50:57 +0000 455)    }
```

Maybe there are no strict reviews for tests or maybe you can allow a little bit not standard code in tests. I don't know, but ...

Such cases of errc usage are only 20 in the whole codebase (and 14 of them are for tests).

```
➜  net git:(master) gogrep 'err := <-err' | wc -l      
20
➜  net git:(master) gogrep 'err := <-err' | grep _test.go | wc -l
14
```

It may happen for everyone to produce not "optimal" code. And about your code - the solution is perfectly fine and it works. But it's not the go way - what I mean by that - over the time go developers used to some patterns. WaitGroup is one of it. It just feels more natural for others if they see the pattern like this because everyone used to it and recognize it.

The other example is much simpler from today:

```go
const (
  A int = 1
  B int = 2
)
```

The code works and it's nothing wrong with it, but it looks wired because everyone used to

```go
const (
  A int = 1
  B       = 2
)
````

There are some patterns that everyone understands and it's preferable to use them (for example `Option` for optional arguments :))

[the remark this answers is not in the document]
Comment (ilgooz) [reply interface/grpc/core/list_services.go]: thank you for the research! i still think that using err chans to wait end of the goroutines is an anti-pattern but i'm ok with the two ways.

**Comment by ilgooz** [comment]
@antho1404 

>I like the part to have the status for the API but I don't think we should do a filter like that. We will need to have a filter for this and maybe for name after or the task or whatever so let's not rush on a feature that is not really necessary and will block us.

>Also the part to remove the services when we stop the core is something that the cli should not do but the core itself should when it exits.

>I think we should do the part for the status of the service but the part for the filter we should take some time to make this reusable kind of like docker manage it's filters and for the part of deleting the services when we stop, this in the end should not even be in the cli so we don't need this api to only have the running services and most of the things in this PR will no longer be needed like the config for example in the API struct

I also agree with this. I'm seeing this PR as intermediate one. Of course, it'd be better if we first move the related logic from _cli_ to _daemon_. Maybe, lets merge this as it is and within another PR, we can rm the filter from api and move the logic to daemon.

**Comment by ilgooz** [comment]
@antho1404 you're right

**Comment by ilgooz** [comment]
good to review

File: commands/commands_mock_test.go [hunk 7]
```diff
@@ -6,11 +6,15 @@ import (
...
 	"github.com/mesg-foundation/core/container"
 	"github.com/mesg-foundation/core/protobuf/coreapi"
-	"github.com/mesg-foundation/core/service"
 	"github.com/mesg-foundation/core/utils/servicetemplate"
 	"github.com/stretchr/testify/mock"
 )
 
+var (
+	_ RootExecutor    = (*mockRootExecutor)(nil)
+	_ ServiceExecutor = (*mockServiceExecutor)(nil)
```
antho1404: I don't see why you added these 2 variables that are not used anywhere

ilgooz: it's to make sure to check if types implements the interfaces correctly or not in the compiler time

dev-core
```diff
@@ -4,8 +4,8 @@ echo "Building mesg/core:local..."
 docker build -t mesg/core:local --build-arg version="local" .
 
 # chain these to stop execution on failure
-./dev-cli stop && \
-./dev-cli start && \
+./dev-cli stop
+./dev-cli start
```
NicolasMahe: Why removing the chain of execution? as the comment says, it's to stop on failure.

interface/grpc/core/list_services.go
```diff
@@ -2,8 +2,11 @@ package core
 
 import (
 	"context"
+	"sync"
 
 	"github.com/mesg-foundation/core/protobuf/coreapi"
+	service "github.com/mesg-foundation/core/service"
```
krhubert: One small change, remove service as it's already the name of pkg

dev-core
NicolasMahe: @ilgooz Please revert the change

commands/service_list_test.go
```diff
@@ -0,0 +1,44 @@
+package commands
+
+import (
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
```
krhubert: Remove empty line

interface/grpc/core/list_services.go
ilgooz: done

commands/service_list_test.go
ilgooz: done

dev-core
ilgooz: it was due to a bug while stopping core. reverted it back. i'll open an issue about the bug later

ilgooz:
good to review

ilgooz:
in progress

ilgooz:
good to review

commands/service_list.go
```diff
@@ -44,18 +45,23 @@ func (c *serviceListCmd) runE(cmd *cobra.Command, args []string) error {
...
+	fmt.Fprintf(w, "STATUS\tSERVICE\tNAME\t\n")
 	for _, s := range services {
-		// status, err := s.Status()
-		// if err != nil {
-		// 	return err
-		// }
-
-		fmt.Fprintf(w, "%s\t%s\n", s.ID, s.Name)
-		//fmt.Fprintf(w, "%s\t%s\t%s\n", status.String(), s.ID, s.Name)
+		var status string
```
krhubert: The service has [String](https://github.com/mesg-foundation/core/blob/3ae39191f1b216586938d30bb764084295bb64f1/protobuf/coreapi/api.pb.go#L105) method, why you recreated it here (and also there is "semi" implementation in the _test.go file)

commands/service_list_test.go
```diff
@@ -0,0 +1,59 @@
...
+		`STATUS\s+SERVICE\s+NAME`,
+	} {
+		matched, err := regexp.Match(fmt.Sprintf(`^\s*%s\s*$`, s), readLine(t, r))
+		require.NoError(t, err)
+		require.True(t, matched)
+	}
+
+	for _, s := range services {
+		var status string
+		switch s.Status {
```
krhubert: Here the same - we can use service.String method

api/api_test.go
```diff
@@ -22,7 +22,9 @@ func newAPIAndDockerTest(t *testing.T) (*API, *dockertest.Testing, func()) {
 	db, err := database.NewServiceDB(testdbname)
 	require.NoError(t, err)
 
-	a, err := New(db, ContainerOption(container))
+	a, err := New(db,
```
krhubert: Why this line was splited into 3 lines? The length of line is fine

commands/service_list.go
ilgooz: I didn't know that, thanks for the heads up!

commands/service_list_test.go
ilgooz: done

ilgooz:
deleted @krhubert's comment  by mistake, it was about removing extra lines while initializing _api.API_ in tests. so, posting this as a reminder.

PR #501, opened 2018-09-25
Add unit test for commands flags

krhubert:


commands/logs_test.go
```diff
@@ -0,0 +1,26 @@
...
+	// create reader for docker stdcopy
+	buf := new(bytes.Buffer)
+	w := stdcopy.NewStdWriter(buf, stdcopy.Stdout)
+	w.Write([]byte("core: 2018-01-01 log\n"))
+
+	m.On("Status").Return(container.RUNNING, nil)
+	m.On("Logs").Return(ioutil.NopCloser(buf), nil)
+	c.cmd.Execute()
+
+	m.AssertExpectations(t)
```
ilgooz: this line can be removed, we don't have any expected arguments while calling the methods. or maybe always keep this to not forget to put it in needed places? 

/cc @mesg-foundation/core

commands/logs_test.go
```diff
@@ -0,0 +1,26 @@
...
+	m := newMockExecutor()
+	c := newLogsCmd(m)
+
+	// create reader for docker stdcopy
+	buf := new(bytes.Buffer)
+	w := stdcopy.NewStdWriter(buf, stdcopy.Stdout)
+	w.Write([]byte("core: 2018-01-01 log\n"))
+
+	m.On("Status").Return(container.RUNNING, nil)
+	m.On("Logs").Return(ioutil.NopCloser(buf), nil)
```
ilgooz: we still need to test if the logs are correctly piped to output streams, otherwise having this test will have no meaning

commands/logs_test.go
krhubert: It isn't only for arguments, it is also for the number of calls (for example if Logs wasn't more times etc...)

commands/logs_test.go
krhubert: This test doesn't check the output. I've just added one test ( as an example ) for checking StartCmdTest output. The rest will be in another PR.

commands/start_test.go
```diff
@@ -6,6 +6,23 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestStartCmdRunE(t *testing.T) {
+	m := newMockExecutor()
+	c := newStartCmd(m)
+
+	closeStd := captureStd(t)
```
ilgooz: i think we should avoid modifying global declarations in tests (like overwriting _os.Stdout_) when it is possible. for this case, it's very easy to provide an in memory writer from _io.Pipe()_ to [here](https://github.com/mesg-foundation/core/blob/3860236/commands/logs.go#L46). please also check [the example](https://github.com/mesg-foundation/core/pull/499/commits/186b264422c80f7cd91dabdc7483bba3cfbb5029).

commands/logs_test.go
ilgooz: yes, but in this case what is this test for exactly? i cannot see

commands/logs_test.go
ilgooz: yes, i see now.

commands/logs_test.go
krhubert: It tests if after Stats returns `running` Logs was called.

I need to add more of this in the feature.

For example if status return `STOPPED` the logs shouldnt be called, etc...

commands/start_test.go
krhubert: If you want to pass os.Stdout and os.Stderr to every command it obscure code and the test (you need to add readLine, run each runE/execute inside goroutines)

```go
func TestServiceList(t *testing.T) {
  var (
    services = []*coreapi.Service{
      {ID: "1", Name: "a", Status: coreapi.Service_RUNNING},
      {ID: "2", Name: "b", Status: coreapi.Service_PARTIAL},
    }
    m        = &mockServiceExecutor{}
    closeStd = captureStd(t)
    c        = newServiceListCmd(m)
  )

  m.On("ServiceList").Return(services, nil)
  c.Execute()
  stdout, _ := closeStd()

  require.True(t, regexp.MustCompile(`\s*^STATUS\s+SERVICE\s+NAME\s*$`).MatchString(stdout))
  require.True(t, regexp.MustCompile(`\s*^running\s+1\s+a`).MatchString(stdout))
  require.True(t, regexp.MustCompile(`\s*^partial\s+2\s+b`).MatchString(stdout))
}  
```

Check out this version. You don't need to readLine, introduce test table writer, create pipe and readers,  pass os.Stdout (and os.Stderr for other funcitons) alwasy to newXXXCmd. 

NOTE: the code hasn't been tested.

ilgooz: * for the _os.Stdout_ & _os_Stderr_ i think that they should be configurable like in the example i provided, there is nothing bad by doing this nor it doesn't complicate the code.

* we can make [this](https://github.com/mesg-foundation/core/pull/499/commits/186b264422c80f7cd91dabdc7483bba3cfbb5029#diff-3ae8740d8b3eaadfa999f0729b2ee664R43) implementation _newable_ and it can have an additional field called _ReadLine()_, it doesn't even need to return errors, we can pass _t_ in the initialization time and it can make err assertion inside. It can be used to test output streams in _commands_ package as a replacement of your _captureStd()_. this kinda abstraction is always good, i'd rather avoiding modifying globals and have the flexibility of reading by line instead of copying all the stream into one big string and doing assertion after that.

```go
type outputWriter struct {
	t *testing.T
	w io.Writer
	r *bufio.Reader
}

// by using this newOutputWriter we can initialize an outputWriter
// and directly pass it to newXXXCmd,
// run the cmd.Execute(),
// and read output stream with outputWriter.ReadLine()
func newOutputWriter(t *testing.T) *outputWriter {
	r, w := io.Pipe() // this can be initialized here, so we don't need to repeat this for all tests.
	return &outputWriter{
		t: t,
		w: w,
		r: bufio.NewReader(r),
	}
}

func (o *outputWriter) Write(b []byte) (n int, err error) {
	return o.w.Write(b)
}

func (o *outputWriter) ReadLine() []byte {
	line, _, err := o.r.ReadLine()
	require.NoError(o.t, err)
	return line
}
```

what do you think about this?

ilgooz: i moved this [outputStream](https://github.com/mesg-foundation/core/pull/499/commits/9a431bf4e85856ff8fa818f5c8196f90025896d1) (outputWriter) to the test util file, here in my PR, maybe we can use it here too

NicolasMahe: as decided in https://forum.mesg.com/t/caputre-stdout-in-command-package/70, we going to capture std.out

PR #506, opened 2018-09-28
full unit test coverage for service package

ilgooz:
by using testify/mock and mocking container package under the hood


service/start_test.go
```diff
@@ -272,7 +292,7 @@ func TestPartiallyRunningService(t *testing.T) {
 	require.Len(t, serviceIDs, len(s.Dependencies))
 
 	for i := range s.Dependencies {
-		require.Equal(t, containerServiceIDs[i], serviceIDs[i])
+		require.True(t, xstrings.SliceContains(serviceIDs, containerServiceIDs[i]))
```
krhubert: there is require.Contains Method

container/container_api.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"io"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/swarm"
+)
+
+// ContainerAPI describes the API of container package.
+type ContainerAPI interface {
```
antho1404: I think it would be nice to have this renamed in `Container` and for the concrete implementation use `DockerContainer` for example, what do you think ?

scripts/build-mocks.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -e
+
+# go get mockery
+go get github.com/vektra/mockery/...
```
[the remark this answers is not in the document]
ilgooz: _vendor/_ folder doesn't include binaries, it's the equivalent of _$GOPATH/src/_. there is a waiting proposal  [here](https://forum.mesg.com/t/create-container-for-development-scripts/62/2) to use a docker container for these kinda cases.

container/container_api.go
ilgooz: I don't understand the part related with _DockerContainer_. Could you explain more? If we rename _ContainerAPI_ to _Container_, we need to change current _Container_ type with another name. But I think we should keep that one as it is because package's name is _container_ and _Container_ is the main type of this package.

antho1404: The idea is to have the container package that has the `Container` interface so just the API. After that we can provide some concrete implementations, here for example Docker but we could have something different and so Docker is the actual implementation.

The optimal solution would be to have the following structure:
- container/
  - container.go
  - docker/
    - start.go
    - stop.go
    - ...

but for now having the interface and the implementation at the same level is fine.

So yes the actual `Container` should be `DockerContainer` or `Docker` or `ContainerDocker` something like that and the interface should be `Container`, that's the interface of the package

NicolasMahe: Totally agree with @antho1404

NicolasMahe: Totally agree with @antho1404.
It makes more sense to have the nicer name for the interface.

But let's keep the same file hierarchy for now: everything in the container folder.

ilgooz: I'm ok to update names but I'm not sure about having multiple implementations of container like _docker_ or _kubernetes_ because the API may vary between those and we may end up with totally different _Container_ APIs.

We have the idea of refactoring _container package_ to collect more logic inside but even then I'm not sure that we'll have the same _Container_ API for different implementations.

Please lighten me if I'm missing something and this is an another topic, we can discuss it in forum.

NicolasMahe: The multiple implementation of the container interface is just a future idea and will not be done now.
I will explain it more in post on the forum.

service/service_integration_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func newContainer(t *testing.T) *container.Container {
+func newIntegrationContainer(t *testing.T) *container.DockerContainer {
```
antho1404: It should be better to return the interface instead of the current implementation 
```
func newIntegrationContainer(t *testing.T) container.Container {
```

ilgooz:
@NicolasMahe I made a fix on your code, please review https://github.com/mesg-foundation/core/pull/506/commits/c8cf92fc0343eda8cb9d6c847f2be59803c5b96d

We cannot automatically set the types within constants if we're not using special _iota_ keyword and it's only for number.

service/service_integration_test.go
ilgooz: done

ilgooz:
good to review

PR #518, opened 2018-10-03
mesg-dev: introduce development container, closes #517

### LINKED ISSUE #517: create a container for development
_related proposal on [forum](https://forum.mesg.com/t/create-container-for-development-scripts/62)._

this container is going to be used to have the same dependencies for development env in developers machines.

it'll mainly be used to run _scripts_ that has 3rd party dependencies and it's also possible to use it to execute commands like _go run_, _go test_.

this container can be named as **mesg-dev**.

usage example:
```
$ cd $GOPATH/src/github.com/mesg-foundation/core

// current usage
$ ./scripts/generate-assets.sh

// new usage > only has the container prefix
$ docker run -v `pwd` mesg-dev ./scripts/generate-assets.sh

// note that since there is no much change on between the current and new usage,
// the new syntax is very straightforward to use and can be easily guessed.
```

ilgooz:
* this provides a development container for devs to provide the 3rd party dependencies and even for running go commands.
* to run things in development container add ./mesg-dev prefix to any command.
* usage examples:
    - ./mesg-dev ./scripts/build-proto.sh
    - ./mesg-dev ./dev-cli status
    - ./mesg-dev ls


ilgooz:
what do you think about _mesg-dev_ file name? do you prefer _dev_ instead?

ilgooz:
maybe we should also add following section to _vendor/_ as a library to have the same version for everybody and we already have the _go install_ on vendor/ in Dockerfile, this way it can produce the binaries from the vendor/.

```
# install dependencies of ./scripts.
RUN go get github.com/golang/protobuf/protoc-gen-go && \
    go get github.com/jteeuwen/go-bindata/... && \
    go get github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc
```

ilgooz:
I've removed the line about removing _vendor/_ that added by _ADD_ inside _Dockerfile-dev_ because it didn't work since files are already cached.

We're mounting the host path of core to the container and _vendor/_ has two copies inside host and container. I though there can be conflicts but there is nothing to worry about, docker only cares about files in the [host machine](https://stackoverflow.com/a/39647631/2175257).

ilgooz:
_go-bindata_ somehow behaves differently when running in dev container. might be relevant with runtime arguments.

corresponding error:
```
$ ./mesg-dev ./scripts/generate-assets.sh
bindata: open /go/src/github.com/mesg-foundation/core/schema.json: no such file or directory
```

ilgooz:
> _go-bindata_ somehow behaves differently when running in dev container. might be relevant with runtime arguments.
> 
> corresponding error:
> 
> ```
> $ ./mesg-dev ./scripts/generate-assets.sh
> bindata: open /go/src/github.com/mesg-foundation/core/schema.json: no such file or directory
> ```

solved by updating import path of _go-bindata_ with the correct one.

ilgooz:
action required, forum [discussion](https://forum.mesg.com/t/create-container-for-development-scripts/62/11) is not completely finalized.

ilgooz:
requires action @mesg-foundation/core

ilgooz:
> Ok for me, if we can make sure that the scripts cannot be called on the host machine but only using the container that would be perfect.

done. /cc @antho1404

ilgooz:
@mesg-foundation/core please review and merge

PR #520, opened 2018-10-03
Add a dedicated task for integration test

NicolasMahe:
The integration test task is still running all the time because so package are still not fully unit tested based.

The goal of this PR is to see the difference between the time and codecov between the unit test and integration test.


NicolasMahe:
I'm closing this PR because it doesn't make sense to separate integration test.
On the opposite, we should not consider different type of test but which type is best to test a specific feature.
See https://forum.mesg.com/t/merge-logic-of-unit-and-integration-test/85

PR #521, opened 2018-10-03
Remove point at the end of sentence

NicolasMahe:


ilgooz:
Please see [this](https://forum.mesg.com/t/error-prone-behavior-while-running-assertions-in-goroutines-in-tests/80) forum post about why tests are not passing in this PR.

PR #524, opened 2018-10-04
fix sync issue on status send chans and sync issue on gRPC deploy stream sends

ilgooz:


commands/service_dev.go
```diff
@@ -51,8 +51,11 @@ func (c *serviceDevCmd) preRunE(cmd *cobra.Command, args []string) error {
 
 func (c *serviceDevCmd) runE(cmd *cobra.Command, args []string) error {
 	statuses := make(chan provider.DeployStatus)
-	go printDeployStatuses(statuses)
+	statusPrintDone := make(chan struct{})
```
krhubert: Use sync.Waitgroup as it's perfect for sync. goroutines (it applies everywhere where done channel is used)

```go
var wg sync.WaitGroup
wg.Add(1)
go func() {
  printDeployStatuses()
  wg.Done()
}()

Deploy()
wg.Wait()
```

ilgooz: done chans are pretty common and really simple to use. could you create a proposal for this on forum by referencing this PR please?

NicolasMahe:
What was the issue that this PR fixes?

ilgooz:
> What was the issue that this PR fixes?

it solves several synchronization bugs in the code.

for the case with gRPC, if we don't have [this](https://github.com/mesg-foundation/core/pull/524/files#diff-cc307140c73bc10debbf2d78fb4eb4b3R33) waiter, there is a possibility of ending the stream by sending the[ last message](https://github.com/mesg-foundation/core/blob/dev/protobuf/coreapi/api.proto#L250-L254) (serviceID or validationErr). In that case if we cannot send all the status messages before, they will be lost.

same thing happens in some other places where status messages are send. for example for in commands package there is a possibility that execution of dev/deploy commands will be done before printing all of the status messages to _os.Stdout_.

commands/service_dev.go
ilgooz: this is a good argument, changes will be made 👍

Issue #528, opened 2018-10-05
Documentation is not accurate for the deploy API


antho1404:
https://docs.mesg.com/api/core.html#deployservice

We cannot send service definition to the API we have either URL or chunk of data based on the tar of the service folder

antho1404:
Idem for all APIs that returns the service, we now have the status

ilgooz:
I added some missing **basic** fields to returned _Service_ type by API. There are other missing but we need to decide which ones to keep or remove first. As I remember @NicolasMahe was talking about removing some fields from the _Service_ type in past.

ilgooz:
solved by https://github.com/mesg-foundation/core/pull/530

PR #530, opened 2018-10-05
protobuf: update docs

ilgooz:
closes https://github.com/mesg-foundation/core/issues/528

### DIFF
--- a/docs/api/core.md
+++ b/docs/api/core.md
@@ -1262,8 +1262,8 @@ The reply's data of the `ListServices` API.
 
 **Example**
 ```json
-[{
-  "service": {
+{
+  "services": [{
     "id": "idX",
     "name": "serviceX",
     "description": "descriptionX",
@@ -1304,8 +1304,8 @@ The reply's data of the `ListServices` API.
         }]
       }]
     }]
-  }
-}]
+  }]
+}
 ```
 
 
--- a/docs/api/core.md
+++ b/docs/api/core.md
@@ -856,7 +856,7 @@ The request's data for `DeployService` API.
 or
 ```json
 {
-  "url": "__SERVICE_GZIPPED_TAR_FILE_CHUNK__"
+  "chunk": "__SERVICE_GZIPPED_TAR_FILE_CHUNK__"
 }
 ```
 
--- a/docs/api/core.md
+++ b/docs/api/core.md
@@ -1646,7 +1646,7 @@ The data will be received over time as long as the stream is open.
 **Example**
 ```json
 {
-  "dependency":  "__DEPENDENCY_THAT_LOG_BELONGS__",
+  "dependency":  "__SERVICE_DEPENDENCY__",
   "type": __LOG_TYPE__,
   "data":  __LOG_CHUNK__,
 }
@@ -1655,9 +1655,9 @@ The data will be received over time as long as the stream is open.
 
 | Field | Type | Description |
 | ----- | ---- | ----------- |
-| dependency | [string](#string) | dependency is the dependency that data belongs. |
-| type | [LogData.Type](#api.LogData.Type) | type is the log type. |
-| data | [bytes](#bytes) | data is a log data chunk. |
+| dependency | [string](#string) | Service dependency that data belongs. |
+| type | [LogData.Type](#api.LogData.Type) | The log type. |
+| data | [bytes](#bytes) | Log data chunk. |
 
 
 
--- a/docs/api/core.md
+++ b/docs/api/core.md
@@ -857,7 +857,7 @@ Stream should be closed after url or all chunks sent to server.
 or
 ```json
 {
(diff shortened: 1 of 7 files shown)


docs/api/core.md
```diff
@@ -850,38 +850,21 @@ The request's data for `DeployService` API.
...
-      }
-    }
-  }
+  "url": "__SERVICE_GIT_URL__"
+}
+```
+or
+```json
+{
+  "url": "__SERVICE_GZIPPED_TAR_FILE_CHUNK__"
```
antho1404: this should be chunk no ?

ilgooz: yes, I'm still working in this PR. I'll request review after completed.

ilgooz:
@antho1404 _ServiceLogs_ has missing docs also, would you like me to add it to this PR as well?

antho1404:
yes can be good

protobuf/coreapi/api.proto
```diff
@@ -281,7 +264,10 @@ message ListServicesRequest {
 // ```json
 // [{
 //   "service": {
+//     "id": "idX",
 //     "name": "serviceX",
+//     "description: "descriptionX",
```
antho1404: `"` missing

antho1404: same for status

protobuf/coreapi/api.proto
```diff
@@ -328,7 +314,10 @@ message GetServiceRequest {
 // ```json
 // {
 //   "service": {
+//     "id": "idX",
 //     "name": "serviceX",
+//     "description: "descriptionX",
+//     "status: "statusX",
```
antho1404: same here

protobuf/coreapi/api.proto
```diff
@@ -358,29 +347,39 @@ message GetServiceReply {
 }
 
 // The request's data for `ServiceLogs` API.
+//
+// **Example**
+// ```json
+// {
+//   "serviceID": "__SERVICE_ID__",
+//   "dependencies": "__SERVICE_DEPENDENCIES__"
```
antho1404: this is a `repeated string` so i'm not sure we can pass a string here, it should be an array

antho1404: Don’t forget about that one or maybe it’s normal?

ilgooz: this is also done, forgot add the comment about this :)

antho1404: Are you sure you committed it ? I can't see it, I'm talking about replacing the example of `ServiceLogsRequest` with
```
{
  "serviceID": "__SERVICE_ID__",
  "dependencies": ["__DEPENDENCY_X__"]
}
```
Putting the array for the repeated string

ilgooz: oh got it. do you think we should do same for the chunks(byte arrays)

antho1404:
Just saw another issue on the documentation concerning the service `ListServicesReply` and `GetServiceReply`.
The service does not contains map anymore for events, outputs, tasks... it contains repeated data so this should also be updated.

Example ListServicesReply:

```json
{
  "services": [{
    "id": "idX",
    "name": "serviceX",
    "description": "descriptionX",
    "status": "statusX",
    "events": [{
      "key": "eventX",
      "data": [{
        "key": "dataX",
        "type": "String"
      }]
    }],
    "tasks": [{
      "key": "taskX",
      "inputs": [{
        "key": "foo",
        "type": "String"
      }],
      "outputs": [{
        "key": "outputX",
        "data": [{
          "key": "resX",
          "type": "String"
        }]
      }]
    }]
  }]
}
```

Example GetServiceReply:

```json
{
  "service": {
    "id": "idX",
    "name": "serviceX",
    "description": "descriptionX",
    "status": "statusX",
    "events": [{
      "key": "eventX",
      "data": [{
        "key": "dataX",
        "type": "String"
      }]
    }],
    "tasks": [{
      "key": "taskX",
      "inputs": [{
        "key": "foo",
        "type": "String"
      }],
      "outputs": [{
        "key": "outputX",
        "data": [{
          "key": "resX",
          "type": "String"
        }]
      }]
    }]
  }
}
```


To verify with the API but it should be something like that

protobuf/coreapi/api.proto
ilgooz: done

ilgooz: actually, the data we're sending is a [byte slice](https://github.com/mesg-foundation/core/blob/dev/protobuf/coreapi/api.pb.go#L726). _bytes_ in protobufs is equal to a byte array and a _repeated bytes_  would be the equivalent of _array of byte arrays_. I think it makes sense to update them as arrays

antho1404:
Updated the list service reply that needed a services attribute that is a list of services
Also updated the cli documentation with some missing documentation

PR #534, opened 2018-10-09
Fix: Add network aliases

### LINKED ISSUE #533: Cannot connect to other dependencies inside a service
When creating other dependencies for a service, the goal is to be able to connect these other dependencies but right now there is no way to do that.

Let's take the following service:
```yml
name: "influxDB"
dependencies:
  influxdb:
    image: influxdb
    ports:
      - "8086"
    volumes:
      - /var/lib/influxdb
```

In my application I should be able to connect to endpoint of this dependency with something like `influxdb:8086` but instead, to connect to this dependency we need to access using the container name which is based on the service hash and so there is no way to predict it directly.

In this example I need to connect to this endpoint `core-e00a68dcd22cd77d996ce540e9113feaa75f92e8-influxdb.1.xd6ps2z5xt2ye81c36hwvb13k:8086` or `core-e00a68dcd22cd77d996ce540e9113feaa75f92e8-influxdb:8086`

I don't really mind having to use this long name if it's at least given in an env variable like 

```
INFLUXDB_HOST_1=core-e00a68dcd22cd77d996ce540e9113feaa75f92e8-influxdb.1.xd6ps2z5xt2ye81c36hwvb13k # the one created for the specific docker task
INFLUXDB_HOST=core-e00a68dcd22cd77d996ce540e9113feaa75f92e8-influxdb # the one associated to the service
```

Or even better be able to access the host `influxdb` directly but this can be complicated and can create some problems if names already exists.

antho1404:
Fix issue of containers from the same service that couldn't communicate with each others by adding the namespace as aliases for the network.

Now if we have the following mesg.yml
```yml
name: "serviceX"
dependencies:
  depA:
    ports:
      - "1234"
  depB:
    ports:
      - "5678"
```
We will be able to access `depA` server through `http://depA:1234` and `depB` with `http://depB:5678` but also with the service ID but this one is not really important and especially hard to do so I don't think we should document this part.

fix #533

### DIFF
--- a/container/service_options.go
+++ b/container/service_options.go
@@ -97,11 +97,15 @@ func (options *ServiceOptions) swarmMounts(force bool) []mount.Mount {
 	return mounts
 }
 
+// swarmNetworks creates all necessary network attachment configurations
+// Each network will be attached based on their networkID but also based on aliases
+// These aliases will make services accessible from other containers inside the same network
 func (options *ServiceOptions) swarmNetworks() (networks []swarm.NetworkAttachmentConfig) {
 	networks = make([]swarm.NetworkAttachmentConfig, len(options.NetworksID))
 	for i, networkID := range options.NetworksID {
 		networks[i] = swarm.NetworkAttachmentConfig{
-			Target: networkID,
+			Target:  networkID,
+			Aliases: options.Namespace,
 		}
 	}
 	return
--- a/container/service_options.go
+++ b/container/service_options.go
@@ -22,10 +22,14 @@ type ServiceOptions struct {
 	Labels    map[string]string
 }
 
-// Network contains an ID of a docker network and it's associated optional alias
+// Network contains an ID of a docker network and it's associated optional alias.
 type Network struct {
-	ID    string // ID of the docker network
-	Alias string // Alias is an optional attribute to name this network and be able to access it using this name
+	// ID of the docker network.
+	ID string
+
+	// Alias is an optional attribute to name this service in the
+	// network and be able to access it using this name.
+	Alias string
 }
 
 // Port is a simplify version of swarm.PortConfig.
@@ -103,9 +107,10 @@ func (options *ServiceOptions) swarmMounts(force bool) []mount.Mount {
 	return mounts
 }
 
-// swarmNetworks creates all necessary network attachment configurations
-// Each network will be attached based on their networkID but also based on aliases
-// These aliases will make services accessible from other containers inside the same network
+// swarmNetworks creates all necessary network attachment configurations for service.
+// each network will be attached based on their networkID and an alias can be used to
+// identify service in the network.
+// aliases will make services accessible from other containers inside the same network.
 func (options *ServiceOptions) swarmNetworks() (networks []swarm.NetworkAttachmentConfig) {
 	networks = make([]swarm.NetworkAttachmentConfig, len(options.Networks))
 	for i, network := range options.Networks {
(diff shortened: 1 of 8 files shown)


antho1404:
I totally agree, but as long as they don't share the same port, they will all be accessible thanks to docker. That's why I say that this is something possible but I don't think we should document that part.
Or maybe we should be able to set a specific alias for the network that the calling function will define.

ilgooz:
> Or maybe we should be able to set a specific alias for the network that the calling function will define.

I don't understand this part.

In general, I think, we don't need to give this flexibility to user even it's not documented. This kinda use looks dangerous to me. Lets think the case that where multiple docker tasks started for _service_ dependency that listens from the same port for load balancing. In that case it'll get attached to the same network with the same _service ID_ alias for the same ports.

container/service_options.go
```diff
@@ -97,11 +97,15 @@ func (options *ServiceOptions) swarmMounts(force bool) []mount.Mount {
...
+// swarmNetworks creates all necessary network attachment configurations
+// Each network will be attached based on their networkID but also based on aliases
+// These aliases will make services accessible from other containers inside the same network
 func (options *ServiceOptions) swarmNetworks() (networks []swarm.NetworkAttachmentConfig) {
 	networks = make([]swarm.NetworkAttachmentConfig, len(options.NetworksID))
 	for i, networkID := range options.NetworksID {
 		networks[i] = swarm.NetworkAttachmentConfig{
-			Target: networkID,
+			Target:  networkID,
+			Aliases: options.Namespace,
```
NicolasMahe: Using directly `options.Namespace` will use each string of the namespace as a aliases. Eg:
```go
options.Namespace := []string{"service-hash", "dependency-name", "etc..."}
```
will result in creating multiple aliases:
"service-hash",
"dependency-name",
"etc..."

The alias should be ONLY the name of the dependency defined in the mesg.yml file.
Eg from Anthony mesg.yml used in the first post of this PR: "depA" for the first dependency. And "depB" for the second dependency.

This way, the dependencies' servers are accessible at `depA:1234` and `depB:5678`.

NicolasMahe:
@ilgooz or @krhubert could you take a look please?
This is a very important feature for creating services ;)

ilgooz:
@NicolasMahe I agree with you on this one, should be only using dependency names as network alias.

antho1404:
I've done some modifications, now the `container` or the `service` are setting an network alias and the aliases for docker are not related to the namespace anymore, only the custom alias given in the data structure

container/service_options.go
```diff
@@ -101,11 +107,15 @@ func (options *ServiceOptions) swarmMounts(force bool) []mount.Mount {
 // Each network will be attached based on their networkID but also based on aliases
 // These aliases will make services accessible from other containers inside the same network
 func (options *ServiceOptions) swarmNetworks() (networks []swarm.NetworkAttachmentConfig) {
-	networks = make([]swarm.NetworkAttachmentConfig, len(options.NetworksID))
-	for i, networkID := range options.NetworksID {
+	networks = make([]swarm.NetworkAttachmentConfig, len(options.Networks))
+	for i, network := range options.Networks {
+		aliases := make([]string, 0)
+		if network.Alias != "" {
+			aliases = append(aliases, network.Alias)
```
ilgooz: why not just using `aliases := []string{network.Alias}` because we only have one alias all the time.

ilgooz: I mean 1 alias at most. Something like this is fine:
```go
cfg := swarm.NetworkAttachmentConfig{
  Target:  network.ID,
}
if network.Alias != "" {
  cfg.Aliases = []string{network.Alias}
}
```

antho1404: done ;)

ilgooz:
needs review

PR #537, opened 2018-10-12
database: create mocks

ilgooz:
needed to write unit tests for _systemservices_ pkg. ASAP


krhubert:
Why do you want to mock database? there is no need to mock it, you can easily create a new instance of database under /tmp filesystem (or current test directory like in other packages).

ilgooz:
> Why do you want to mock database? there is no need to mock it, you can easily create a new instance of database under /tmp filesystem (or current test directory like in other packages).
> 
> Mocking it only completes the test (that uses a database) with no visible pros.

I like the idea of not depending on external systems _-like filesystem in this case-_ as much as possible. Creating a database mock with mockery is dead simple and providing fakes with _testify/mock_ is very straightforward.

Yes, I agree that having  some piece of code that interacts with filesystem is not a problem with unit testing but if avoiding it also simple, I prefer avoiding. As addition to this, in future we might depend on an external database that runs as an independent server which requires mocking. So having a mock in the beginning is a win win situation.

antho1404:
I agree this one was not really necessary but I like that we have this interface and concrete implementation maybe the mocks are too much but again it's just one command to execute to have this flexibility so it's nice and we could like that create some tests that mimics some issues with the system so i don't see any problem to merge that

PR #535, opened 2018-10-10
introducing systemservices

ilgooz:

PR #541, opened 2018-10-19
prepare a base for new workflow feature

ilgooz:
Depends on https://github.com/mesg-foundation/core/pull/535, related with [introducing workflow files](https://forum.mesg.com/t/introducing-workflow-files/95).

This PR creates a base for new workflow feature. It aims to successfully connect cli workflow commands, workflow gRPC APIs and workflow system service together with a dummy implementation for creating and deleting workflows.

To test this PR please first run _dev-core_. After that, you can try following two new commands:
```
# these commands interacts with WSS but WSS responds with dummy data.
./dev-cli workflow create /Users/ilgooz/Desktop/a.yml 
./dev-cli workflow delete test
```

You can watch the logs of _WSS_ with _docker logs_ command after you find the id with _docker ps_.

### DIFF
--- a/systemservices/workflow/workflow.go
+++ b/systemservices/workflow/workflow.go
@@ -3,6 +3,7 @@ package workflow
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 
 	"github.com/mesg-foundation/core/execution"
 	"github.com/mesg-foundation/core/utils/workflowparser"
@@ -43,7 +44,7 @@ func CreateOutputs(e *execution.Execution) (id string, err error) {
 	case "error":
 		return "", errors.New(e.OutputData["message"].(string))
 	}
-	panic("unreachable")
+	return "", fmt.Errorf("workflow: task create workflow has unknown output %s", e.OutputKey)
 }
 
 // DeleteInputs maps delete task's inputs.
@@ -62,5 +63,5 @@ func DeleteOutputs(e *execution.Execution) error {
 	case "error":
 		return errors.New(e.OutputData["message"].(string))
 	}
-	panic("unreachable")
+	return fmt.Errorf("workflow: task delete workflow has unknown output %s", e.OutputKey)
 }


ilgooz:
This PR is done, It'll be rebased to dev when it's dependency PR (https://github.com/mesg-foundation/core/pull/535) is merged.

ilgooz:
This PR is done, It'll be rebased to dev when it's dependency PR (#535) is merged.

ilgooz:
good to review

commands/provider/workflow_provider.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	defer file.Close()
+
+	fileData, err := ioutil.ReadAll(file)
+	if err != nil {
+		return "", err
+	}
+
+	reply, err := p.client.CreateWorkflow(context.Background(), &coreapi.CreateWorkflowRequest{
+		File: fileData,
```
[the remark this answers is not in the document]
ilgooz: done

systemservices/workflow/workflow.go
```diff
@@ -0,0 +1,70 @@
...
+const (
+	CreateTaskKey = "create"
+	DeleteTaskKey = "delete"
+)
+
+// Workflow is a high level wrapper for Workflow System Service.
+// It calls the WSS's tasks and reacts to its event through network.
+// WSS responsible for managing and running workflows.
+type Workflow struct {
+	api       *api.API
```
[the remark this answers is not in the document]
ilgooz: done

commands/workflow_create.go
```diff
@@ -0,0 +1,50 @@
...
+
+	pretty.Progress("Creating workflow...", func() {
+		workflowID, err = c.e.CreateWorkflow(filePath, c.name)
+	})
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("%s Workflow running\n", pretty.SuccessSign)
+	fmt.Printf("To see its logs, run the command:\n\tmesg-core workflow logs %s\n", workflowID)
```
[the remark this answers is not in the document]
ilgooz: Yes, but creating a workflow also is not implemented. This is just an intermediate PR. If you prefer, I can close this PR and we can directly merge https://github.com/mesg-foundation/core/pull/559. It includes this PR as well.

The only reason why I created this PR first is to make reviewing easier.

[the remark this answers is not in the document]
antho1404: it's fine here ;)

systemservices/sources/workflow/workflow/proto/api.pb.go
```diff
@@ -0,0 +1,2817 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// source: github.com/mesg-foundation/core/protobuf/coreapi/api.proto
+
+package coreapi
```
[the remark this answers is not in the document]
krhubert: It's because of we vendor all packages inside each systemservice. We can use one vendor per core and use it (in such case each service will be importing from main vendor)

commands/provider/workflow_provider.go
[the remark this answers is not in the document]
krhubert: YAML is the format and tells nothing about the purpose of this field, Please change it to `Definition`.

protobuf/coreapi/api.proto
```diff
@@ -42,6 +42,54 @@ service Core {
...
+//
+// **Example**
+// ```json
+// {
+//   "file":   __FILE__,
+//   "name": "__UNIQUE_WORKFLOW_NAME__"
+// }
+// ```
+message CreateWorkflowRequest {
+  bytes file = 1;  // workflow.yml file.
```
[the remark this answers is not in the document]
ilgooz: This is sounds a better way. But, by extracting the yaml parsing logic from core, we're leaving this work to clients. _Workflow.yaml_ should be a spec defined by MESG and core should natively support that language. In this perspective, I like the idea of accepting yaml for workflow creations.

I don't think that we will ever need to compile a representation to another one like json to yaml, in fact, we may even need to create our own language in future and drop yaml for a much dedicated & cleaner syntax. 

For the `direct api call from a web ui`, I agree on the flexibility of this one but yaml will be widely used in forums discussions, tutorial pages or in mesg application market and community will get very familiar with it. Thus, application/yaml builders will be widely used and implemented by community as well. In this case, a JS library created by us or the community can build a yaml from a definition object that you're proposing.

@antho1404 I both like your version and the plain yaml support in the API. I'll be fine with both because both ideas have their own reasonings. As a reminder, on services, we also parse `mesg.yml` inside core instead leaving this parts to clients.

interface/grpc/core/create_workflow.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"context"
+
+	"github.com/mesg-foundation/core/protobuf/coreapi"
+	"github.com/mesg-foundation/core/systemservices/workflow"
+)
+
+// CreateWorkflow creates and runs a new workflow.
+func (s *Server) CreateWorkflow(ctx context.Context, request *coreapi.CreateWorkflowRequest) (*coreapi.CreateWorkflowReply, error) {
+	exec, err := s.api.ExecuteAndListen(s.ss.WorkflowServiceID(), workflow.CreateTaskKey, workflow.CreateInputs(request.File, request.Name))
```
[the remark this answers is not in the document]
ilgooz: Could you provide an example about how to move this piece of code to _core/api_ please? You can describe the new funcs that we need to create in _core/api_ with their arguments and return values. No need for an actual implementation in the funcs' body.

systemservices/sources/workflow/workflow/proto/api.pb.go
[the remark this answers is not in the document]
antho1404: Hum I see...
This API is not supposed to be accessible, a service should not access the host api, even if for now it's technically possible this should be dropped this can be a huge risk and we can always use the service composition for that.
This part should be delegated to the core, I really see the following process
- The core receive an event from a service
- The core ask the workflow system service the matching tasks 
- The core execute the tasks

protobuf/coreapi/api.proto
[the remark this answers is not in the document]
antho1404: For the `mesg.yml` we actually sent the whole directory to the core because there is a dependency between the files and the mesg.yml. We don't have this dependency here.

The reason why I want to be as close as possible as pure data and not the representation of these data is that we are not yet sure about the representation of the workflow file, we always talk about yml but we have to do more research, maybe there is existing workflow templates even maybe in xml that exists already with an existing ecosystem like an ui or some validation etc... If we keep only the data without the representation we can adapt really easily.

Also for the work from the client side, I don't see it as a problem, we have the cli in our case that can parse and send the data and an UI can directly send the data without having to create a fake temporary file just in order to send it to the api.

We could also have a specific language that is compiled to these data kind of like an AST for most programming languages and AST are the necessaries data not the representation of it. Representation are just for users (eg: prettify for js)

systemservices/sources/workflow/workflow/proto/api.pb.go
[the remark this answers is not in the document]
ilgooz: I don't know why it's a risk to use public APIs of core from a service. I need a much detailed overview of service composition to have more opinions. Maybe we can keep discussion moving forward on the related forum proposal. 

If we listen events & results and execute tasks on core itself, that means we're going to move a good amount of logic from WSS. The main goal was to split core to system services and move logic from core to system services as much as possible.

I'd like to hear more to get a better understanding of your view.

interface/grpc/core/create_workflow.go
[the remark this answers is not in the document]
antho1404: Basically it would be something like that (this is not tested and just done quickly to get the idea)

## package `api`
```go
package api

func (a *API) CreateWorkflow(workflowDefinition WorkflowDefinition) (workflowID string, err error) {
	return newWorkflowCreator(a, ss).Create(workflowDefinition)
}

type workflowCreator struct {
	api *API
	ss *systemservices.SystemServices
}

func newWorkflowCreator(api *API, ss *systemservices.SystemServices) *workflowCreator {
	return &workflowCreator{
		api: api,
		ss: ss,
	}
}

func (w *workflowCreator) Create(workflowDefinition WorkflowDefinition) (workflowID string, err error) {
	exec, err := w.api.ExecuteAndListen(w.ss.WorkflowServiceID(), "create", workflow.CreateInputs(definition))
	if err != nil {
		return nil, err
	}
 	return workflow.CreateOutputs(exec)
}
```

## package `interface/grpc/core`

```go
package core

func (s *Server) CreateWorkflow(ctx context.Context, request *coreapi.CreateWorkflowRequest) (*coreapi.CreateWorkflowReply, error) {
	id, err := s.api.CreateWorkflow(request.WorkflowDefinition)
	if err != nil {
		return nil, err
	}
	return &coreapi.CreateWorkflowReply{
		id: id,
	}, 
}
```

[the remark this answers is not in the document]
ilgooz: This is great, I had the same idea but we can't import _systemservices_ inside _api_ because it's [cycling](https://github.com/mesg-foundation/core/blob/62ae275/systemservices/systemservices.go#L6).

I think this is relevant with the second bullet you mentioned in your [comment](https://github.com/mesg-foundation/core/pull/535#issuecomment-436565176) in system services PR. I'd like to know what kind of structure you're thinking to create in the _systemservices_ & _api_ pkgs while moving service deploy/start logic to _api_ pkg.

protobuf/coreapi/api.proto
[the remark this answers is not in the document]
ilgooz: Having API to accept an object instead yaml makes a bit more sense, I agree with it. In future, if we decide to create our own definition language, maybe then we can support both ways. So clients can build the object files or send the definition file directly depending on their cases. I'm updating the code!

systemservices/sources/workflow/workflow/proto/api.pb.go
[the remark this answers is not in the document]
ilgooz: we discussed about this one on the weekly meeting and we decided that core itself will be responsible for listening events & results and executing tasks. 

so we'll have two new tasks on the WSS to lookup for what tasks should be executed depending on results or events. 

proposed in the chat:
```
task: lookupTaskFromEvent -> output with array of tasks with their input data that needed to be executed
task: lookupTaskFromResult -> output with array of tasks with their input data that needed to be executed
```

interface/grpc/core/create_workflow.go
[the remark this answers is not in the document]
ilgooz: we discussed about this one on the weekly meeting and we decided to create a new pkg under named _systemservices/deployer_ and it'll be responsible to deploy and start system services. _systemservices_ pkg will call method(s) from this pkg in the initialization process.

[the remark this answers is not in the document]
ilgooz: originally related with https://forum.mesg.com/t/the-import-cycle-between-api-systemservices-packages/105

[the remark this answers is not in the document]
ilgooz: solved by https://github.com/mesg-foundation/core/pull/567

ilgooz:
this PR waiting for https://github.com/mesg-foundation/core/pull/567 to be merged

ilgooz:
@mesg-foundation/core good to review

docs/api/service.md
```diff
@@ -38,12 +38,12 @@ The event and its data must be defined in the [service's definition file](../gui
...
 The request's data for the `EmitEvent` API.
 
-**Example:**
+Example:
 ```json
 {
-  "token":     "__SERVICE_TOKEN_FROM_ENV__",
-  "eventKey":  "__EVENT_KEY__",
-  "eventData": "{\"foo\":\"hello\",\"bar\":false}"
+"token":     "__SERVICE_TOKEN_FROM_ENV__",
```
antho1404: take care there is some problems with the generation of the doc, all the json are flatten and the example is not in bold anymore

https://deploy-preview-541--mesg-docs.netlify.com/api/core.html#getservice

ilgooz: Do you know how to solve this without manually editing the docs? Running _build-proto.sh_ doesn't correct the docs.

ilgooz: What is your version?

antho1404: For protoc I have `libprotoc 3.6.1` (I think this one is the same, there is not much versions)
For protoc-gen-go I cannot have the version but the commit hash is `b27b920f9e71b439b873b17bf99f56467623814a` this one should be the problem

ilgooz:
@mesg-foundation/core good to review

api/workflow_create.go
```diff
@@ -0,0 +1,40 @@
...
+
+// newWorkflowCreator returns a new workflow creator with given api.
+func newWorkflowCreator(api *API) *workflowCreator {
+	return &workflowCreator{
+		api: api,
+	}
+}
+
+// Create creates and runs a new workflow with optionally given unique name.
+func (w *workflowCreator) Create(definition yaml.WorkflowDefinition, name string) (id string, err error) {
```
krhubert: Move Create method body to CreateWorkflow

api/workflow_delete.go
```diff
@@ -0,0 +1,35 @@
...
+
+// newWorkflowCreator returns a new workflow creator with given api.
+func newWorkflowDeletor(api *API) *workflowDeletor {
+	return &workflowDeletor{
+		api: api,
+	}
+}
+
+// Delete stops and deletes a workflow. id can be unique id or name.
+func (w *workflowDeletor) Delete(id string) (err error) {
```
krhubert: Same here - move Delete body to DeleteWorkflow

commands/provider/workflow_provider.go
```diff
@@ -0,0 +1,45 @@
...
+	return reply.ID, nil
+}
+
+// DeleteWorkflow stops and deletes workflow with id.
+func (p *WorkflowProvider) DeleteWorkflow(id string) error {
+	_, err := p.client.DeleteWorkflow(context.Background(), &coreapi.DeleteWorkflowRequest{ID: id})
+	return err
+}
+
+func (p *WorkflowProvider) toProtoWorkflowDefinition(definition yaml.WorkflowDefinition) *coreapi.CreateWorkflowRequest_WorkflowDefinition {
```
krhubert: Is the struct left empty intentionaly here?

commands/provider/workflow_provider.go
```diff
@@ -0,0 +1,45 @@
+package provider
+
+import (
+	"context"
+
+	"github.com/mesg-foundation/core/protobuf/coreapi"
+	"github.com/mesg-foundation/core/utils/workflowparser/yaml"
```
krhubert: Change github.com/mesg-foundation/core/utils/workflowparser/yaml to github.com/mesg-foundation/core/utils/workflowparser

Yaml is good package name for parsing yaml (github.com/go-yaml/yaml), but here shouldbe `workflowparser` or `workflow/parser`. If you want to put emphasize on format, you can create
```go
type Parser interface {
  Parse() (Definition, error)
}

type YAMLParser struct {} // ....
```

Or just create 
```go
workflowparser.ParseFromYAMLFile()
```

Gopkg.toml
```diff
@@ -32,6 +32,10 @@ required = [
   "github.com/vektra/mockery/cmd/mockery"
 ]
 
+[[override]]
+  version = "v1.1.0"
+  name = "github.com/pseudomuto/protoc-gen-doc"
```
krhubert: Why this protoc-gen-doc is included in this PR? Are we actually using this? They already have docker image and posted in the readme how to use this and thier using this like we do with our docker scripts.

commands/workflow_create.go
```diff
@@ -0,0 +1,50 @@
...
+	)
+
+	pretty.Progress("Creating workflow...", func() {
+		workflowID, err = c.e.CreateWorkflow(filePath, c.name)
+	})
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("%s Workflow running\n", pretty.SuccessSign)
```
krhubert: is running

api/workflow_create.go
ilgooz: I kind of agree with this, but we follow the same pattern all around the _api_. I think, for now we should keep this one as it's and if we want to make a change about this pattern we can do it for all methods together in a different PR.

api/workflow_delete.go
ilgooz: I kind of agree with this, but we follow the same pattern all around the _api_. I think, for now we should keep this one as it's and if we want to make a change about this pattern we can do it for all methods together in a different PR.

commands/provider/workflow_provider.go
ilgooz: Yes, it'll be used by the next PR.

Gopkg.toml
ilgooz: We already require this pkg in the Gopkg file, but I just added these lines to lock its version. Otherwise _dep ensure_ may install an older version which causes to this issue: https://github.com/mesg-foundation/core/pull/541#pullrequestreview-175688242

commands/workflow_create.go
ilgooz: done

commands/provider/workflow_provider.go
ilgooz: done

commands/provider/workflow_provider.go
krhubert: Ok

api/workflow_create.go
krhubert: There is no need to keep it and then change all at once, we can start one by one and get rid of it eventually.

Gopkg.toml
krhubert: Ok

api/workflow_create.go
ilgooz: I don't think it's something that we need to get rid of. To me, having it good to isolate funcs from different groups, otherwise they all will be defined globally which kills the isolation, causes poor naming on funcs and opens up the possibility of overlapping func names. If we can find a better way, we can switch to it but by having a discussion first. We can think about a guideline and maybe open a forum proposal.

NicolasMahe: I agree with @ilgooz on this one

systemservices/workflow/workflow.go
```diff
@@ -0,0 +1,66 @@
...
+// CreateOutputs maps create task's outputs.
+// id is unique workflow id.
+func CreateOutputs(e *execution.Execution) (id string, err error) {
+	switch e.OutputKey {
+	case "success":
+		return e.OutputData["id"].(string), nil
+	case "error":
+		return "", errors.New(e.OutputData["message"].(string))
+	}
+	panic("unreachable")
```
NicolasMahe: Could we get rid of panic?
Ok, the code should not be unreachable, but for a magical reason that get it reached, then it make the core crash.
Can we just return an error with a nice message?

systemservices/workflow/workflow.go
```diff
@@ -0,0 +1,66 @@
...
+
+// DeleteOutputs maps delete task's outputs.
+func DeleteOutputs(e *execution.Execution) error {
+	switch e.OutputKey {
+	case "success":
+		return nil
+	case "error":
+		return errors.New(e.OutputData["message"].(string))
+	}
+	panic("unreachable")
```
NicolasMahe: Same here about panic

systemservices/workflow/workflow.go
ilgooz: To me, this is where having tests comes in because they're meant to catch programmer errors. 

As you also said, panic should never happen. Our business logic does not depend on having one. If we get a panic anyway, it'd be due to a programmer error. Which means, we have a bug in Core or WSS. We should write tests to avoid buggy code to be merged which could cause to this panic instead of returning an error that will never get handled.

If you still think that we should return an error instead, I'm ok with it.

systemservices/workflow/workflow.go
ilgooz: https://github.com/mesg-foundation/core/pull/541#discussion_r234530261

systemservices/workflow/workflow.go
ilgooz: In Discord, we agreed on using errors instead for now. In future we may change to goroutines again and recover any panic on gRPC handlers to prevent Core to crash.

ilgooz: done

systemservices/workflow/workflow.go
ilgooz: done

ilgooz:
@NicolasMahe good to review

Issue #540, opened 2018-10-18
Possible memory leak with long running core


antho1404:
I started a core for about 3 hours and the memory went from 0 to 250MB, we probably have a memory leak somewhere.

The core was running with an application that use erc20 service and a log service so basically listen to ~100 events every 10 sec and execute a task for each events.

<img width="463" alt="screen shot 2018-10-18 at 6 21 36 pm" src="https://user-images.githubusercontent.com/1783126/47151119-0a8a5280-d303-11e8-9e48-4acc459c6380.png">

ilgooz:
could you share application service stack please so we can reproduce easier

krhubert:
Done

ilgooz:
This one requires a confirmation by running the same stack for a few hours to see if the problem is solved.

antho1404:
<img width="459" alt="screen shot 2018-10-30 at 8 02 04 am" src="https://user-images.githubusercontent.com/1783126/47701458-20104e00-dc1a-11e8-90a6-1b03271dfdea.png">

Seems like everything is fine now

PR #542, opened 2018-10-20
Remove executed executions from memory

antho1404:
This might me really related to #540 
Every time we were executing a task, the executions was kept in memory, making the memory going up more and more over time.
This is a small fix to remove the execution when it's done. We should later on put that in a database but that will be another work

### DIFF
--- a/execution/state.go
+++ b/execution/state.go
@@ -7,7 +7,6 @@ import (
 var mu sync.Mutex
 var pendingExecutions = make(map[string]*Execution)
 var inProgressExecutions = make(map[string]*Execution)
-var processedExecutions = make(map[string]*Execution)
 
 // InProgress returns the matching in progress execution if exists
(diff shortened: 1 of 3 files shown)


execution/state.go
```diff
@@ -7,7 +7,6 @@ import (
 var mu sync.Mutex
 var pendingExecutions = make(map[string]*Execution)
 var inProgressExecutions = make(map[string]*Execution)
-var processedExecutions = make(map[string]*Execution)
 
 // InProgress returns the matching in progress execution if exists
 func InProgress(ID string) (execution *Execution) {
```
ilgooz: we need to Lock in here before accessing to data

antho1404:
Thanks for the fixes, I'm also working on having a database to store all these executions so all these queues should disappear soon hopefully.
This one is ready to merge, I want to do it quickly to update the server and see if it fixes #540

PR #543, opened 2018-10-21
organize initialization of gRPC server and core

ilgooz:


interface/grpc/core/type.go
```diff
@@ -11,24 +9,7 @@ type Server struct {
...
-	return s, nil
-}
-
-// APIOption sets underlying mesg API.
-func APIOption(api *api.API) Option {
-	return func(s *Server) {
-		s.api = api
-	}
+// NewServer creates a new Server.
+func NewServer(api *api.API) *Server {
```
antho1404: I saw it yesterday and wanted to change it 👍

PR #553, opened 2018-10-24
UI issue with service execute command fix #546

### LINKED ISSUE #546: UI issue with service execute command
The loader is displayed weirdly when using `service execute` command without passing task and data

screen shot 2018-10-22 at 11 46 42

krhubert:
fix #546


commands/service_execute.go
```diff
@@ -153,10 +153,17 @@ func readJSONFile(path string) (string, error) {
 	if path == "" {
 		return "{}", nil
 	}
+
 	data, err := ioutil.ReadFile(path)
 	if err != nil {
 		return "", err
 	}
+
+	var raw json.RawMessage
```
antho1404: Do we need this test here ? The test is already done by the API
With this test
```
invalid character '"' after object key:value pair
```
Without the test
```
rpc error: code = Unknown desc = invalid character '"' after object key:value pair
```

I would recommend that if it's an error that the API catch we should let the API catch this. What do you think ?

krhubert: We don't but If we network comes into play, callers of API should do as much validation as possible to save the network overhead. So if we can easily validate it here and get user faster response (which increase user experience) why not to?

PR #560, opened 2018-10-26
Executions stored in Database

antho1404:
I've created a database to store all executions. For now it is a leveldb database but this can be changed in the future if needed. The purpose of this PR is really to not store any execution state in memory to avoid huge memory consumption.


execution/execution_test.go
```diff
@@ -0,0 +1,174 @@
...
+	taskKey       = "task"
+	defaultInputs = map[string]interface{}{
+		"foo": "hello",
+		"bar": "world",
+	}
+	tags = []string{"tag1", "tag2"}
+)
+
+func db(t *testing.T, dir string) DB {
+	fmt.Println(dir)
```
krhubert: remove println

execution/execution_test.go
```diff
@@ -0,0 +1,174 @@
...
+	db, err := New(dir)
+	require.NoError(t, err)
+	return db
+}
+
+func TestCreate(t *testing.T) {
+	db := db(t, filepath.Join(os.TempDir(), "TestCreate"))
```
krhubert: - Files are not deleted after tests
- use https://golang.org/pkg/io/ioutil/#TempFile

execution/leveldb.go
```diff
@@ -0,0 +1,125 @@
...
+	"fmt"
+	"time"
+
+	"github.com/cnf/structhash"
+	"github.com/mesg-foundation/core/service"
+	"github.com/syndtr/goleveldb/leveldb"
+)
+
+// LevelDB is a concrete implementation of the DB interface
+type LevelDB struct {
```
krhubert: Change name as it's not levelDB struct, rather ExecutionDB

execution/leveldb.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return &LevelDB{db: db}, nil
+}
+
+// Create a record in the database to store this execution and returns the id
+// returns an error if any problem happen with the database
+// returns an error if inputs are invalid
+func (db *LevelDB) Create(service *service.Service, taskKey string, inputs map[string]interface{}, tags []string) (*Execution, error) {
+	task, err := service.GetTask(taskKey)
```
krhubert: GetTask + ReuireInputs logic dosen't belong to this package

Create should take Execution as parametr

So in that case it should only return erro

execution/leveldb.go
```diff
@@ -0,0 +1,125 @@
+package execution
```
krhubert: Change file name (it's not leveldb but executiondb)

execution/execution_test.go
```diff
@@ -0,0 +1,174 @@
...
+	defer db.Close()
+	tests := []struct {
+		inputs map[string]interface{}
+		assert bool
```
krhubert: s/assert/hasError/

execution/execution_test.go
```diff
@@ -0,0 +1,174 @@
...
+	}{
+		{inputs: map[string]interface{}{"foo": "hello", "bar": "world"}, assert: false},
+		{inputs: map[string]interface{}{"foo": "hello"}, assert: true},
+	}
+	for _, test := range tests {
+		execution, err := db.Create(srv, taskKey, test.inputs, tags)
+		if test.assert {
+			require.Error(t, err)
```
krhubert: add continue and remove else for readability

execution/leveldb.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/syndtr/goleveldb/leveldb"
+)
+
+// LevelDB is a concrete implementation of the DB interface
+type LevelDB struct {
+	db *leveldb.DB
+}
+
+// New creates a new DB instance
+func New(path string) (DB, error) {
```
krhubert: New should return LevelDB, not interface

api/api_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"os"
 	"testing"
 
+	"github.com/mesg-foundation/core/execution"
+
```
krhubert: Empty line

execution/execution_test.go
```diff
@@ -0,0 +1,174 @@
...
+		} else {
+			require.NoError(t, err)
+			require.NotNil(t, execution)
+			e, err := db.Find(execution.ID)
```
krhubert: This is Find test (not create)

execution/execution_test.go
```diff
@@ -0,0 +1,174 @@
...
+func TestFind(t *testing.T) {
+	db := db(t, filepath.Join(os.TempDir(), "TestFindExecution"))
+	defer db.Close()
+	e, _ := db.Create(srv, taskKey, defaultInputs, tags)
+	tests := []struct {
+		id     string
+		assert bool
+	}{
+		{e.ID, false},
+		{"noid", true},
```
krhubert: keep consistency with tests:

change to "dosen't exists" or "noid" for all tests

execution/leveldb.go
antho1404: In that case the `Complete` function will be the same and this package is only about the database and we need to create another package to store the logic of this execution package or we need to delegate this part to the api but this will make the api way more complicated.

The goal of this package is to be high level and manage this kind of verification.

We could create another package that only handles the database like that we can have it in the database folder but this package will be pretty much useless

execution/leveldb.go
antho1404: I think we should keep the `levelDB`, I can (as the service db package) append the executionDB to have `LevelDBExecutionDB` but we need to keep the kind of implementation in the name.

ilgooz:
this one has some conflicts that needs to be resolved

execution/leveldb.go
krhubert: I was thinking about it and I think it shouldn't be there.

If you want to create Execution obj then do it with `execution.New`. Like every package this one should have execution.New (or NewFromService etc) and pass it to a database. 

Same for `complete` Exeuction object should have something like `func (e *Exection) Verify(outputKey, outputData) (error)` and this one should be called from `Complete`, but I still consider it (I mean verification) shouldn't be there at all.

execution/execution.go
```diff
@@ -0,0 +1,44 @@
...
+	Inputs            map[string]interface{} `hash:"inputs"`
+	Output            string                 `hash:"-"`
+	OutputData        map[string]interface{} `hash:"-"`
+	CreatedAt         time.Time              `hash:"createdAt"`
+	ExecutedAt        time.Time              `hash:"-"`
+	ExecutionDuration time.Duration          `hash:"-"`
+}
+
+// DB exposes all the functionalities
+type DB interface {
```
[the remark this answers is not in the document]
ilgooz: To me, this interface should be named something like _Executor_. It should keep the current content _Create(), Find(), Execute(), Complete() and Close()_.

There should also be a _Storage/DB_ interface that has a content like _Get(), Delete(), Update()_ etc. and _Executor_'s methods should use it.

The advantage of this, it splits the database layer from the business logic. Any change on database can be handled in the implementation of _Storage_ without touching to _Executor_.

This is not a must but good to have.

antho1404:
I've done some update like splitting the database part (and put it in the database package) and also have the execution package that only manage the business logic related to the execution object. The responsibility to save this execution is done directly in the api package.

Waiting more feedbacks on that

database/execution_db.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/cnf/structhash"
+	"github.com/mesg-foundation/core/execution"
+	"github.com/syndtr/goleveldb/leveldb"
+)
+
+// ExecutionDB exposes all the functionalities
+type ExecutionDB interface {
+	Find(executionID string) (*execution.Execution, error)
+	Save(execution *execution.Execution) (*execution.Execution, error)
```
ilgooz: no need for _*execution.Execution_ in return statement

database/execution_db.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	var execution execution.Execution
+	err = json.Unmarshal(data, &execution)
+	return &execution, err
+}
+
+// Save an instance of executable in the database
+// Returns an error if anything from marshaling to database saving goes wrong
+func (db *LevelDBExecutionDB) Save(execution *execution.Execution) (*execution.Execution, error) {
+	id := fmt.Sprintf("%x", sha1.Sum(structhash.Dump(execution, 1)))
```
ilgooz: id shouldn't be generated from hash calculation of execution struct. there is no a specific need for this so we can avoid doing it and use uuid. otherwise while updating something from execution, the ID also needs to be changed for no real reason.

database/execution_db.go
```diff
@@ -0,0 +1,62 @@
...
+	execution.ID = string(id)
```
ilgooz: this one should check if execution has an ID already. if it has, update the existent one, otherwise create a new record in the database.

execution/execution.go
```diff
@@ -0,0 +1,102 @@
...
+	Tags              []string               `hash:"tags"`
+	Inputs            map[string]interface{} `hash:"inputs"`
+	Output            string                 `hash:"-"`
+	OutputData        map[string]interface{} `hash:"-"`
+	CreatedAt         time.Time              `hash:"createdAt"`
+	ExecutedAt        time.Time              `hash:"-"`
+	ExecutionDuration time.Duration          `hash:"-"`
+}
+
+// New a record in the database to store this execution and returns the id
```
ilgooz: legacy docs

execution/execution.go
```diff
@@ -0,0 +1,102 @@
...
+		Inputs:    inputs,
+		TaskKey:   taskKey,
+		Tags:      tags,
+		CreatedAt: time.Now(),
+		Status:    Created,
+	}, err
+}
+
+// Execute a given execution
+// Returns an error if the execution doesn't exists in the database
```
ilgooz: legacy docs

execution/execution.go
```diff
@@ -0,0 +1,102 @@
...
+			ActualStatus:   execution.Status,
+			ExpectedStatus: Created,
+		}
+	}
+	execution.ExecutedAt = time.Now()
+	execution.Status = InProgress
+	return nil
+}
+
+// Complete verifies the output associated to the execution and save this to the database
```
ilgooz: legacy docs

database/execution_db.go
antho1404: I did the exact same behavior than the serviceDB if the ID is not present then we raise an error

antho1404:
I changed slightly the current implementation, now the hash is not calculated on the execution date but based on an `eventID`, executions should always be executed based on an event and every events should have an unique identifier. For now we generate the unique Id only based on an UUID but with decentralized services this should be the ID of the event (like the transaction hash on ethereum for example)

ilgooz:
> I changed slightly the current implementation, now the hash is not calculated on the execution date but based on an `eventID`, executions should always be executed based on an event and every events should have an unique identifier. For now we generate the unique Id only based on an UUID but with decentralized services this should be the ID of the event (like the transaction hash on ethereum for example)

I don't get these changes. Why are we complicating unique id generation? To me, we can create and set _execution.Execution.ID_ with a randomly generated uuid inside _execution.New()_ or _execution.Save()_. Why do we need to generate a hash over _execution.Execution_ data structure? We're not looking for uniqueness of executions. The only thing needed is a random unique id for identifying each execution.

antho1404:
We actually need "uniqueness of executions". When we have the distributed version of MESG, the execution will be created by one node but all other nodes should be able to reproduce the same id based on the same data to ensure that they are verifying/executing the exact same execution. For now it's not mandatory to use a hash but it will be and it doesn't add much complexity, we already have the exact same principle for services for the same reason.

execution/execution_test.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func TestExecute(t *testing.T) {
+	e, _ := New(srv, "xxx", taskKey, map[string]interface{}{"foo": "1", "bar": "2"}, tags)
+	tests := []struct {
+		id       string
+		hasError bool
+	}{
+		{e.ID, false},
+		{"doesn't exists", true},
```
ilgooz: I'm removing this because I think, it's forgotten here.

Issue #558, opened 2018-10-24
dependency volume changes when source code is changed


ilgooz:
[Problematic line](https://github.com/mesg-foundation/core/blob/f099449/service/start.go#L140). I'm working on WSS with mongodb, when I make a change on source code of WSS, mongodb's data volume also changes and I lose all the previous data.

/cc @antho1404

antho1404:
I was thinking about this one too when I worked on the influxdb service and yes it can be annoying. In some cases it's actually a good thing to have this all automatic, but in some other cases it's really annoying so maybe it should be the responsibility of the user and have a parameter "bind" or something like that.

antho1404:
Also the line you are pointing is really risky to remove because what happen if 2 services have the same name of dependency and same name of volume eg: db -> db that might be a big problem because data will be shared

antho1404:
I think the `update` option is not really related and I think this issue will solve the problem https://github.com/mesg-foundation/core/issues/275 but until then I really think we could either force to bind the volume in the `.mesg` folder based on the name like we had before or even make an option for that.
The final solution is really to work on this "alias/ID" for service that will always be static and we can use that as identifier for the volume and this will not change anymore.

ilgooz:
@antho1404 please confirm and close this issue in favor of forum proposal.

antho1404:
This is still a bug that can be solved by the service ID feature but it should be kept open as reminder

antho1404:
can't we for now revert to the initial system that was doing the binding all the time of the different volumes on the host disk based on the name of the service (with dash).

This will solve the problem and we can later on use the alias/id of the service.

What do you think ?

ilgooz:
@antho1404 you mean using the service name in the yaml instead its hash as volume name?

antho1404:
yes

but even this might not be enough because with this we still rely on docker to create volumes but we cannot bind these volumes to the actual host so maybe using the name instead of the hash AND binding this to the host in the `~/.mesg/services/{serviceName}/{dependencyName}/{volumePath}`

ilgooz:
I agree, but I'd like to keep Docker volumes instead of binding volumes to host machine for isolation.

In future we can have new features like volume plugins, providing data backups to users or easily moving core, its whole stack and data volumes to another machine. Having data in Docker volumes may give us more flexibility for these kind of use cases.

Is there a reason to the bind volumes to host machine?

antho1404:
[...]

PR #571, opened 2018-11-13
Possibility to customize volume from mesg.yml

antho1404:
Fix https://github.com/mesg-foundation/core/issues/558

This is still an ongoing discussion on the issue and the goal of this PR is to show a possible implementation but not necessary the one to merge.

Attempt to solve the volume issue with the following solution:

If a volume doesn't have any extra informations (eg: `/path/in/the/container`), this creates a docker volume that is automatically managed by docker and created based on the ID of the service (so will change for every new version of the service).

If the volume have the complete informations (eg: `/local/path:/path/in/the/container`), a directory will be created under `~/.mesg/{serviceName}/{dependencyName}/{path}` and this directory will be bind inside the container to ensure that the data from the container and the host are shared.
This helps to customize any service by providing some configurations in this directory.


ilgooz:
Very nice!
* First we need to make sure that services that use the same name in their _yml_ file will not get the same folder on host machine's filesystem. This causes to a bug. The only solution I see for now is to checking the uniqueness of names on service deployment. Maybe we can figure out more solutions.

* Do we really want devs to force using MESG path? While specifying a path from host machine with `~/myhost/a/:/container/a` syntax, the expected behaviour is to find the folders as the place it's pointed, in this case it's here `~/myhost/a/` not under _.mesg/services/service-name/_ path. What is the point of this? If this is a must we need to figure out a more revealing syntax.

* We should be supporting mounting special files like unix sockets. _Or no?_

* Current implementation only supports persistent data by using the mounting data to host option. We should be supporting the same for Docker volumes as well. Or completely get rid of mounting to host option and require users to create their own data volumes for their configuration files. So, Core can use them. Core, should be able to use pre-existent Docker volumes with services otherwise we'll have problems with databases too, they require data to be permanently stored.

One other, we can have a very similar pattern with _docker-compose_. I need more thinking before having conclusion but for now I can say that: 

If we don't implement giving the ability of setting _unique_ names for services before, for now, we can both let choosing volume names and absolute paths from the host machine, to the devs. Like _docker-compose_ does with the following syntax:
* `data-volume:/data/a`
Using the Docker volume named as _data-volume_. User is responsible to not use same volume name for different type of services. I want to remind that, It's also possible devs to provide their configs within Docker volumes without the need of mounting to host. So if volume is already created we can directly use it. A pre-existing volume might be keeping configuration files or keeping some data for example, from a database. Both ways, we still have control on these volumes and can provide backups or other features to devs in future like I mentioned in my previous [comment](https://github.com/mesg-foundation/core/issues/558#issuecomment-438276429). But, I don't remember if we can attach existent data volumes to Docker services, we should verify that first.

* `/host/a:/data/a` 
Mounting to an absolute path (_/host/a_) on host machine.

* `/var/run/docker.sock:/var/run/docker.sock`
Mounting a socket.

antho1404:
## Name uniqueness
Agree with that, if the name of the service and the name of the dependency and the name of the volume are the same then we have the same volume but in that case isn't it the same service ?
I'm ok to leave it like that as a known issue (probably really rare) and that will be solved by the Alias/ID for services

## Force MESG path
We can't really allow to put any kind of path (absolute path). If we do so it means that I can develop a service that is asking for the path of the docker socket or the logs of my machine and can exploit that.
This can have serious implications so I want to reduce this risk as much as possible by just "namespacing" this on the host machine and after it's the responsibility of the host to change this path, make a symlink or anything like that.

## Unix socket
This work out of the box, everything is considered as a file so sharing a folder, a file or a socket is the same in docker. We already share the socket when we run the daemon and don't have any issues with that.

## Volume persistency
Docker volumes are persistent, every volume created are identified by their key (in this case `hash(ID, dependency, volume)`) so as long as the these information are not updated the same volume will be used and persisted

NicolasMahe:
Actually, after experimentation with @antho1404, we don't need to bind any volume to the host to solve the problem that we have experienced.

There is a `volumesfrom` property in the `mesg.yml` that mount dependencies' volumes in the service docker.
From @krhubert example:
```yaml
name: tendermint
configuration:
  volumesfrom:
     - tendermint
dependencies:
  tendermint:
    image: tendermint/tendermint
    volumes: 
      - /tendermint
```
It will bind the path `/tendermint` in both the dependency and the service itself.

NicolasMahe:
I would like to insist that binding volume to the host machine is to avoid as much as possible.
It doesn't bring any feature because volumes can be shared between containers (and that's actually what docker recommend to do).
Moreover, services should be stateless. Giving access to the host filesystem is a contradiction and developers may use the feature in a wrong way.

ilgooz:
@antho1404 
>...case isn't it the same service ?

It may not, service names might be overlapping and we shouldn't put data folders to the same service namespace when services are different. Even if their dependency names and volume names are the same. Because it's ambiguous, services should not share the same service namespace.

> ...We can't really allow to put any kind of path (absolute path). If we do so it means that I can develop a service that is asking for the path of the docker socket or the logs of my machine and can exploit that....

On the security side while thinking Network feature, yes I agree this one as well. Totally makes sense. A peer in the node should not be revealing its entire filesystem to services. But still, we need a much more clear syntax on this one in order to prevent confusions. Because, current naming with _volumes_ field in the yml configs is too close to Docker's world and people can think that they're functioning in the same way. To prevent the confusion we can change the _volumes_ name to something else or introduce a different syntax or at least have good documentation about it.

> This work out of the box, everything is considered as a file so sharing a folder, a file or a socket is the same in docker. We already share the socket when we run the daemon and don't have any issues with that.

I haven't tested the code but I think current implementation has a small bug that prevents this feature to work. [This line](https://github.com/mesg-foundation/core/blob/571bfe8c7e1338cb010f9c236cc2e5b4c705f303/service/start.go#L149) always think the first part of **:** is a directory and tries to create them where unix socket are files and already exists in the filesystem.

> Docker volumes are persistent, every volume created are identified by their key (in this case hash(ID, dependency, volume)) so as long as the these information are not updated the same volume will be used and persisted

About this one, I was pointing about something else. Current implementation supports persistent data by mounting to the host machine but doesn't cover the same case with Docker volumes because it respects to service id only, which changes when source code changed. I think we should be supporting both ways. This can be done by named services or letting devs to specify volume name like _docker-compose_ does.

PR #574, opened 2018-11-16
systemservices/deployer: improve logs

ilgooz:
I reverted back logging messages like in the first implementation of SS. I lowercased the log messages for convience. Added total system services number. Removed use of _%q_ because their _'_ escaped during logs are printed which is hard to read. And removed some extra logs to cut the noise.


ilgooz:
@antho1404 @krhubert can you review again?

Issue #536, opened 2018-10-10
determining container status


ilgooz:
This [line](https://github.com/mesg-foundation/core/blob/f53b994/container/container.go#L159) is error-prone, it doesn't respect the case where docker services stopped manually by the devs instead of core itself. 

When devs rm all the services with the docker cli command, containers will still remain even if they are shutdown. So they should be removed too. Otherwise, core thinks that containers are still running and because of that services will not be started.

This may not count as a bug but it's definitely something that we should think about it and at least document it in the troubleshooting wikis.

In my opinion if there are no side-effects, a service should be counted as _STOPPED_ when service is not found even if it's container exists.

ilgooz:
this weird behaviour also happening *I think* when do not rm services manually. For example sometimes core won't start even cli says it's started.

@antho1404 suggests:
```
I guess docker that doesn't delete the container when we remove the service
and because there is a container we concider that the docker instance is running
```
```
we should make sure to always delete the container (even force delete after a timeout) and also maybe check the status of an existing container
```

ilgooz:
https://github.com/mesg-foundation/core/pull/586 might solve this

ilgooz:
I did some tests by removing MESG services manually with docker commands. Which ends up that their containers eventually deleted as well within a few seconds.

It believe that, this issue might be the same with the one described in #586.

I'm removing the hard cleanup of containers and services from dev script with this PR https://github.com/mesg-foundation/core/pull/614. So, we can experience if there is any problem left.

PR #577, opened 2018-11-20
Make dev core run faster

krhubert:
Please ignore tests for circleci - the source code wasn't modified and I don't want to run in them forever


antho1404:
I've got some errors when running `./dev-core`

```
➜  core git:(feature/faster-dev-core) ./dev-core
compile cli and core
./dev-core: line 17: sha256sum: command not found
```

I saw this resource https://unix.stackexchange.com/questions/426837/no-sha256sum-in-macos#426838
Maybe using openssl make it more compatible ?

krhubert:
Changed to `openssl md5`

dev-core
```diff
@@ -1,16 +1,49 @@
...
-echo "Building mesg/core:local..."
-docker build -t mesg/core:local --build-arg version="local" .
+set -e
+
+DIR=${GOPATH:-$HOME/go}
+VERSION=local
+LDFLAGS="-X 'github.com/mesg-foundation/core/version.Version=$VERSION'"
+
+echo "compile cli and core"
+go build -o ./bin/core -ldflags="$LDFLAGS" core/main.go
```
antho1404:
```suggestion
GOOS=linux GOARCH=amd64 go build -o ./bin/core -ldflags="$LDFLAGS" core/main.go
```

antho1404: It doesn't work on mac because it's compiled in the host machine. The core needs to be compiled for the container.

antho1404: Error on OSX
```
standard_init_linux.go:190: exec user process caused "exec format error"
```

krhubert: Done

dev-core
```diff
@@ -1,16 +1,49 @@
...
-docker build -t mesg/core:local --build-arg version="local" .
+set -e
+
+DIR=${GOPATH:-$HOME/go}
+VERSION=local
+LDFLAGS="-X 'github.com/mesg-foundation/core/version.Version=$VERSION'"
+
+echo "compile cli and core"
+GOOS=linux GOARCH=amd64 go build -o ./bin/core -ldflags="$LDFLAGS" core/main.go
+GOOS=linux GOARCH=amd64 go build -o ./bin/cli -ldflags="$LDFLAGS" interface/cli/main.go
```
antho1404: CLI is running on the host, only the core is running on the container

```suggestion
go build -o ./bin/cli -ldflags="$LDFLAGS" interface/cli/main.go
```

antho1404: fix done

dev-core
```diff
@@ -1,16 +1,49 @@
...
 # create a symlink for system services.
-rm -rf ~/.mesg/systemservices
-mkdir -p ~/.mesg
-ln -s $GOPATH/src/github.com/mesg-foundation/core/systemservices/sources ~/.mesg/systemservices
+rm -rf "$HOME/.mesg/systemservices"
+mkdir -p "$HOME/.mesg/"
+ln -s "$DIR/src/github.com/mesg-foundation/core/systemservices/sources" "$HOME/.mesg/systemservices"
+
+echo "build mesg/core image"
+if [[ ! $BINCACHED ]]; then
```
ilgooz: I don't quite get why are we doing our own cache check? Previous version of _dev-core_ already had Docker cache and if source code is not changed it doesn't even compile the source code use cache instead. What is the thing that we're making faster?

dev-core
```diff
@@ -1,16 +1,49 @@
...
+  docker build -f Dockerfile.dev -t "mesg/core:$VERSION" .
+fi
+
+# remove docker core service if it wasn't clean up.
+COREID="$(docker service ls --format="{{.ID}} {{.Name}}" | grep core | awk '{{print $1}}')"
```
ilgooz: It's better if we can also _docker rm_ the containers of removed services. Because sometimes Core things that if it can find a service's container then service is still running and it won't start the service when requested.

dev-core
krhubert:
```bash
$ cat Dockerfile.dev
FROM ubuntu:18.04
COPY ./bin/core .
CMD [ "./core" ]

$ time docker build -f Dockerfile.dev -t "mesg/core:local" . > /dev/null
docker build -f Dockerfile.dev -t "mesg/core:local" . > /dev/null  2,98s user 0,61s system 85% cpu 4,224 total

$ # ~3 seconds to copy and build the image

$ time openssl md5 ./bin/core > /dev/null
openssl md5 ./bin/core  0,04s user 0,01s system 95% cpu 0,056 total

$ # compared to ~0s to check if core binary has changed - if not then don't the image.
```

dev-core
krhubert: > Because sometimes Core things that if it can find a service's container then service is still running and it won't start the service when requested.

What is core things? I don't get what you mean.

dev-core
krhubert: So go build also uses go cache and if the source hasn't changed it won't rebuild it (it wasn't possible when we build inside docker build image) so this is first step to make it faster. The second one is to avoid build mesg:core/local if core md5 sum hasn't changed.

dev-core
ilgooz: Yes it should be. But in the case where Core removes services we had this issue https://github.com/mesg-foundation/core/issues/536. Still, sometimes I required to execute this `docker rm -f $(docker ps -a -q)` command to get rid it. We should second check if the issue still persist.

krhubert: ok, I do I know which container was run by core services? Is it possible to know it, do we have some labels? If we have then I will add it right away, if not then I don't `docker rm -f $(docker ps -a -q)` because it will remove everything.

ilgooz: Yes we have labels for containers as well.

krhubert: Done

krhubert:
Closing in favor of https://github.com/mesg-foundation/core/pull/584

PR #579, opened 2018-11-21
Fix gometalinters errors.

krhubert:


Gopkg.lock
```diff
@@ -148,15 +148,14 @@
...
+  digest = "1:b16db30b5311fb0946093d7afdc8352c6528280d8b2e77926229e60dd47672d5"
   name = "github.com/go-bindata/go-bindata"
   packages = [
     ".",
     "go-bindata",
   ]
   pruneopts = "UT"
-  revision = "5ebd2c033d84e12a7f4aae4483d98ec7dde53b66"
-  version = "v1.0.0"
+  revision = "41975c0ccc30077fd791b15b383818e805d0a57e"
```
ilgooz: Can you lock new version inside toml file please?

ilgooz: Actually it's possible, check below for an example :
https://github.com/mesg-foundation/core/blob/1cdab23/Gopkg.toml#L30
https://github.com/mesg-foundation/core/blob/1cdab23/Gopkg.toml#L36-L37

Manually editing _lock_ file should be avoided as much as possible. It can get overwritten by further _dep ensure_ executions. In this case it may try to install the tagged version. 

If we want to force to install _master_, it should be versioned inside toml file.

krhubert: oh, cool, there was no revision/version before it. done

ilgooz: I think, directly putting revision hash not supported. We need to use _branch=master_. I haven't tested but as I remember, this is needed otherwise if we lose related vendor from the vendor dir for example, a folloing _dep unsure_ will erase the changes that manually made to lock file. And it'll try to install last tagged version. This is why we need to tell to _dep_ to use master branch so it can respect to revision given in the lock file.

krhubert: https://golang.github.io/dep/docs/Gopkg.toml.html#constraint

`  # Note that only one of "branch", "version" or "revision" can be specified.`

PR #583, opened 2018-11-23
Alias system

### LINKED ISSUE #605: Merge db service and alias
See comments https://github.com/mesg-foundation/core/pull/583#discussion_r235911954

Wait first for https://github.com/mesg-foundation/core/pull/583 to be merged

Need to merge the db service and alias and use transaction to guarantee in case of error that the database are in sync

### LINKED ISSUE #612: Output alias in command service list
The command `service list` should output the services' alias (if not generated).

Depending on https://github.com/mesg-foundation/core/pull/583

NicolasMahe:
Fix https://github.com/mesg-foundation/core/issues/558 about re-using the same volumes.
Closes https://github.com/mesg-foundation/core/issues/607, closes #605 and closes #612 

This PR introduce a new optional property called **alias** in `mesg.yml`.
The alias can be use instead of the generated service id in any API and command that accept a service ID.
If the alias is omitted, one is generated based on the service hash.
If the alias already exist, it get replace by the last deployed service.

There is **no** breaking change.


database/service_db.go
```diff
@@ -77,39 +82,66 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
...
 
-	return d.unmarshal(id, b)
+	return d.unmarshal(idOrAlias, b)
 }
 
 // Save stores service in database.
 func (d *LevelDBServiceDB) Save(s *service.Service) error {
-	if s.ID == "" {
-		return errors.New("database: can't save service without id")
+	if s.ID == "" || s.Alias == "" {
```
ilgooz: Maybe we should always require an ID and _alias_ can stay as optional. Because we always save services by using their ID as key.

database/service_db.go
```diff
@@ -77,39 +101,76 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
 }
 
 // Delete deletes service from database.
-func (d *LevelDBServiceDB) Delete(id string) error {
-	return d.db.Delete([]byte(id), nil)
+// TODO this should also find and delete alias when id is given.
+func (d *LevelDBServiceDB) Delete(idOrAlias string) error {
```
ilgooz: this should also find and delete alias when id is given

service/namespace.go
```diff
@@ -11,6 +11,7 @@ const resultChannel string = "Result"
 
 // namespace returns the namespace of the service.
 func (service *Service) namespace() []string {
+	return []string{service.Alias}
```
ilgooz: has two returns, let's not forget updating

service/start.go
```diff
@@ -135,9 +136,11 @@ func (d *Dependency) extractVolumes() ([]container.Mount, error) {
 	return volumes, nil
 }
 
+// volumeKey create an key for the volume based on the alias to make sure that the volume
+// will be consistent from one service's version to another.
 func volumeKey(s *Service, dependency string, volume string) string {
 	return xstructhash.Hash([]string{
-		s.ID,
+		s.Alias,
```
ilgooz: @antho1404 I thought alias was optional, but here we're using it all the time.

krhubert:
It would be great if we can provide versioning along with alias.

antho1404:
The idea with the versioning is to have a list of hashes for a specific alias instead of a single hash for a specific alias with by default the first one (or last one) is the latest deployment but we could go back in time with all the deployments and see for the same alias all the different versions updated.

It's not really versions like semver, it's more all the previous versions deployed. (Is it what you were thinking @krhubert ?)

This is something we might do later on but i really wanted to keep it simple for now

database/service_db.go
antho1404: alias should be optional for the user but needs to be required internally

service/namespace.go
antho1404: alias is not good anyway with the system that we have, it will generate a id too long for docker

service/start.go
antho1404: It should be optional for the user but not internally

database/service_db.go
```diff
@@ -10,18 +10,36 @@ import (
...
-	Delete(id string) error
-	Get(id string) (*service.Service, error)
-	Save(s *service.Service) error
 }
 
 // LevelDBServiceDB is a database for storing service definition.
 type LevelDBServiceDB struct {
-	db *leveldb.DB
+	db      *leveldb.DB
+	aliases *leveldb.DB
```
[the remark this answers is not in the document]
antho1404: this makes things a bit more complicated but i agree on the principle

service/importer/assets/schema.json
```diff
@@ -2,22 +2,22 @@
   "$schema": "http://json-schema.org/draft-06/schema#",
   "type": "object",
   "required": [
-    "name"
+    "name",
```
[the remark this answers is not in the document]
antho1404: no i removed it in the last commit, now the alias field can be optional for the user and a new unique alias is generated if not present

database/service_db.go
```diff
@@ -130,8 +191,17 @@ func (e *DecodeError) Error() string {
...
 
-// IsErrNotFound returs true if err is type of ErrNotFound, false otherwise.
+// IsErrNotFound returns true if err is type of ErrNotFound, false otherwise.
 func IsErrNotFound(err error) bool {
 	_, ok := err.(*ErrNotFound)
 	return ok
 }
+
+// ErrSameAlias error returned when there is a service with the same alias.
+type ErrSameAlias struct {
```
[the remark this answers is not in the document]
antho1404: If 2 services have the same alias the second one will override the first one so for now it's the developers responsibility not to choose 2 services with the same alias. Later with the network and services deployed on the network, the author will be set in the service and if the author doesn't have the right to update the service then the developer will have an error and will need to find a new alias

database/service_db.go
```diff
@@ -10,18 +10,36 @@ import (
...
 
+const aliasesPathSuffix = "_aliases"
+
+var (
+	errCannotSaveWithoutID = errors.New("database: can't save service without id")
+)
+
 // ServiceDB describes the API of database package.
 type ServiceDB interface {
+	// Save saves a service to database.
```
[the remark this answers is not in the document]
NicolasMahe: I changed the logic about alias:
if alias already exist, it get replaced by new one.

database/service_db.go
NicolasMahe: done

database/service_db.go
```diff
@@ -77,39 +82,66 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
...
+		}
+	}
 	return d.db.Delete([]byte(id), nil)
 }
 
 // Get retrives service from database.
-func (d *LevelDBServiceDB) Get(id string) (*service.Service, error) {
-	b, err := d.db.Get([]byte(id), nil)
+func (d *LevelDBServiceDB) Get(idOrAlias string) (*service.Service, error) {
+	id, err := d.aliases.Get([]byte(idOrAlias), nil)
```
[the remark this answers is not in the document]
NicolasMahe: Done

database/service_db.go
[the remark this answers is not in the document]
NicolasMahe: I change the logic to allow an alias to be replaced.
Otherwise, during the development of a service, the developer would have to change the service alias at every deployment, exactly the opposite of what we want to do..

NicolasMahe:
Ready to review. Please see PR first message.

service/namespace.go
NicolasMahe: I revert to use ID here.

database/service_db.go
```diff
@@ -64,52 +87,119 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
...
 func (d *LevelDBServiceDB) Close() error {
-	return d.db.Close()
+	if err := d.db.Close(); err != nil {
+		return err
+	}
+	return d.aliases.Close()
+}
+
+// getIDAndAlias returns and separates id from alias.
+// id cannot be empty if an error is returned.
```
ilgooz: I cannot understand what's the meaning of this

database/service_db.go
```diff
@@ -64,52 +87,119 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
 				logrus.WithField("service", decodeErr.ID).Warning(decodeErr.Error())
 				continue
 			}
+			iter.Release()
```
ilgooz: seems forgotten here (it shouldn't be called during iteration?)

database/service_db.go
```diff
@@ -64,52 +87,119 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
...
 			return nil, err
 		}
 		services = append(services, s)
 	}
 	iter.Release()
```
ilgooz: it's better to call this with a differ before starting to iterate

database/service_db.go
```diff
@@ -64,52 +87,119 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
...
+	id = string(idB)
+	return id, alias, nil
+}
+
+// getAliasFromID returns the alias pointing to the id if exist.
+func (d *LevelDBServiceDB) getAliasFromID(id string) (alias string, err error) {
+	iter := d.aliases.NewIterator(nil, nil)
+	for iter.Next() {
+		if string(iter.Value()) == id {
+			alias = string(iter.Key())
```
ilgooz: we can directly return here and release iteration with defer before starting to iterate

service/inject_definition.go
```diff
@@ -10,6 +11,10 @@ import (
 // injectDefinition applies service definition to Service type.
 func (s *Service) injectDefinition(def *importer.ServiceDefinition) {
 	s.Name = def.Name
+	s.Alias = def.Alias
+	if s.Alias == "" {
+		s.Alias = s.generateAlias()
```
ilgooz: why are we forcing to have an alias? it might be better to only use ids when an alias is not provided by user. what's the purpose of creating an alias internally which is never used by the user?

service/inject_definition.go
```diff
@@ -144,3 +149,10 @@ func (s *Service) defParametersToService(params map[string]*importer.Parameter)
 	}
 	return ps
 }
+
+func (s *Service) generateAlias() string {
+	alias := strings.Replace(s.Name, " ", "-", -1)
+	alias = strings.Trim(alias, " _-.,")
+	alias = strings.ToLower(alias)
+	return alias
```
antho1404: If I run the same service with a modification in the code, they still share the same volume with this alias. This is a breaking change.
The current behavior is that every changes in the code of the service generates a new id and a new volume.
Maybe this alias need to be generated at the end based on the hash of the service (without the alias) like that if the service is deployed 2 times with the same code, the volume will be the same but if there is a modification in the code then the volume will be different

antho1404: This fixes the problem

```diff
diff --git a/service/inject_definition.go b/service/inject_definition.go
index a3a6c888..80e8926e 100644
--- a/service/inject_definition.go
+++ b/service/inject_definition.go
@@ -12,9 +12,6 @@ import (
 func (s *Service) injectDefinition(def *importer.ServiceDefinition) {
 	s.Name = def.Name
 	s.Alias = def.Alias
-	if s.Alias == "" {
-		s.Alias = s.generateAlias()
-	}
 	s.Description = def.Description
 	s.Repository = def.Repository
 	s.Events = s.defEventsToService(def.Events)
@@ -154,5 +151,5 @@ func (s *Service) generateAlias() string {
 	alias := strings.Replace(s.Name, " ", "-", -1)
 	alias = strings.Trim(alias, " _-.,")
 	alias = strings.ToLower(alias)
-	return alias
+	return alias + s.computeHash()
 }
diff --git a/service/service.go b/service/service.go
index 15701cc3..1e44fb1c 100644
--- a/service/service.go
+++ b/service/service.go
@@ -207,6 +207,10 @@ func (s *Service) deploy() error {
 	s.configuration.Key = "service"
 	s.configuration.Image = imageHash
 	s.Dependencies = append(s.Dependencies, s.configuration)
+
+	if s.Alias == "" {
+		s.Alias = s.generateAlias()
+	}
 	return nil
 }
```

database/service_db.go
NicolasMahe: It's not forget.
It need to be done at the end of the iteration, the next line is a return, so it will stop the iteration

database/service_db.go
NicolasMahe: i'm not sure, because then we call `iter.Error()`. so release may populate some error.

database/service_db.go
NicolasMahe: this: "id cannot be empty if an error is returned." ?
if there is no error, then ID is never empty

service/inject_definition.go
NicolasMahe: We are relaying on alias for docker volume, but yes we could do a better system that an generated alias.
We could create a alias getter that return it if if it exist or return the service id otherwise.
@antho1404 what do you think?

antho1404: we could also have a test on the docker volume part to take either the alias or the id but i'm pretty sure we might need this in other places so generating an alias is really convenient. Also I don't see any problem of having default properties, we already have something like that for the configuration

database/service_db.go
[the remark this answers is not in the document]
NicolasMahe: We agree to do it in another PR. See https://github.com/mesg-foundation/core/issues/605

database/service_db.go
ilgooz: that's right

database/service_db.go
ilgooz:
```suggestion
// id will be filled if there are no errors.
```

service/inject_definition.go
NicolasMahe: Ok I change the behavior on empty Alias like you suggest.
One little dif, the alias is directly equal to the service hash:
```go
if s.Alias == "" {
    s.Alias = s.computeHash()
}
```

service/inject_definition.go
NicolasMahe: Ok for me

ilgooz: The problem with this is we're generating the alias from the _Name_ field of _mesg.yml_. In other means we're forcing users to define a unique _Name_. If we're going to create an alias anyway depending on the _Name_ why we have both of them, it's a bid odd. If this is an acceptable behavior, I'm ok with this.

NicolasMahe: The alias (if omitted) is now generated from the service hash

service/start.go
```diff
@@ -120,9 +121,11 @@ func (d *Dependency) extractVolumes() ([]container.Mount, error) {
 	return volumes, nil
 }
 
+// volumeKey create an key for the volume based on the alias to make sure that the volume
```
ilgooz:
```suggestion
// volumeKey creates a key for service's volume based on the service's alias to make sure that the volume
```

service/start.go
```diff
@@ -120,9 +121,11 @@ func (d *Dependency) extractVolumes() ([]container.Mount, error) {
...
+// will be consistent from one service's version to another.
```
ilgooz:
```suggestion
// will stay the same for different versions of the service.
```

database/service_db.go
```diff
@@ -88,80 +87,65 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
...
+	// encode service
 	b, err := d.marshal(s)
 	if err != nil {
 		return err
 	}
+	// save service
+	if err := d.db.Put([]byte(s.ID), b, nil); err != nil {
+		return err
+	}
+	// save alias if exist
```
ilgooz: this shouldn't be optional and should return an error if not exists

krhubert:
close #605

database/service_db.go
```diff
@@ -54,57 +76,80 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
 		services []*service.Service
 		iter     = d.db.NewIterator(nil, nil)
 	)
-
 	for iter.Next() {
```
NicolasMahe: Iterator already manage prefix, see example:
```
Iterate over subset of database content with a particular prefix:

iter := db.NewIterator(util.BytesPrefix([]byte("foo-")), nil)
for iter.Next() {
	// Use key/value.
	...
}
iter.Release()
err = iter.Error()
...
```

NicolasMahe:
Also closes https://github.com/mesg-foundation/core/issues/607

NicolasMahe:
I approve the change made by @krhubert, good job 👍

ilgooz:
I approve as well 👍

krhubert:
@ilgooz nice catch 👍  I improved it a little bit - removed unmarshal + remove data race (you didn't use transaction and there was a chance that between `Get` and `Delete` some other goroutine will override the alias)

ilgooz:
> @ilgooz nice catch 👍 I improved it a little bit - removed unmarshal + remove data race (you didn't use transaction and there was a chance that between `Get` and `Delete` some other goroutine will override the alias)

_A few notes about using transactions here:_

* I agree that we should use transactions in _Save_ and _Delete_. By doing that we're going to make sure that saving/deleting the alias and saving/deleting the service will be guaranteed to be made together. So let's apply transactions for _Save_ as well (but we use batch there too, I'm not sure that leveldb batches are atomic and they guarantee all or nothing. if so we don't need transactions).

* For the transaction that added to Delete with the last [commit](https://github.com/mesg-foundation/core/pull/583/commits/fbb72c3fe97349470fb0cbdaf28975081dfd89da), are we sure that other concurrent writes while a transaction is open will not overlap with each other? As I see in docs, they guarantee that there will not be two open transactions at the same time but what about single writes that can be caused by a _Save_?

* Every core has their own databases so only one core can access to database at the same time. Therefore, we can simply use _sync.Lock_ between _Delete_ and _Save_ instead transactions here. Just an idea.

* There is a small bug in the last [commit](https://github.com/mesg-foundation/core/pull/583/commits/fbb72c3fe97349470fb0cbdaf28975081dfd89da). It doesn't get the alias when _Delete_ called called with an id.

database/service_db.go
```diff
@@ -98,14 +98,35 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
...
-	return d.db.Write(batch, nil)
+	batch.Delete([]byte(idKeyPrefix + id))
+	if alias != "" {
+		batch.Delete([]byte(aliasKeyPrefix + alias))
+	}
+	if err := tx.Write(batch, nil); err != nil {
+		tx.Discard()
+		return err
+	}
+	return tx.Commit()
```
NicolasMahe: As @ilgooz noticed:
"It doesn't get the alias when Delete called called with an id."
-> so the alias is not deleted when an id is pass to the function delete

NicolasMahe:
Guys, please stop adding feature to this PR. It gets too big.
The issues https://github.com/mesg-foundation/core/issues/605 and https://github.com/mesg-foundation/core/issues/612 were about waiting this PR to be merged and then create other PR to fix the issues.
Let's fix the logic issue and merge this PR.

ilgooz:
@NicolasMahe 👍 I didn't see we had an issue about having alias in _service list_. https://github.com/mesg-foundation/core/issues/612

database/service_db.go
```diff
@@ -52,59 +75,103 @@ func (d *LevelDBServiceDB) unmarshal(id string, value []byte) (*service.Service,
...
-
-	return services, nil
+	return services, iter.Error()
 }
 
 // Delete deletes service from database.
-func (d *LevelDBServiceDB) Delete(id string) error {
-	return d.db.Delete([]byte(id), nil)
+func (d *LevelDBServiceDB) Delete(idOrAlias string) error {
+	tx, err := d.db.OpenTransaction()
```
NicolasMahe: Why opening a transaction if the get and delete function are not using it?
It should be tx.Get, tx.Delete, etc...
https://godoc.org/github.com/syndtr/goleveldb/leveldb#Transaction

krhubert: Sorry it was in rush, I've pushed a fix. now should be ok

krhubert:
> So let's apply transactions for Save as well (but we use batch there too, I'm not sure that leveldb batches are atomic and they guarantee all or nothing. if so we don't need transactions).

They are atomic. You can check it [here](https://github.com/syndtr/goleveldb/blob/master/leveldb/db_write.go#L263) (it uses transaction under the hood)

> For the transaction that added to Delete with the last commit, are we sure that other concurrent writes while a transaction is open will not overlap with each other? As I see in docs, they guarantee that there will not be two open transactions at the same time but what about single writes that can be caused by a Save?

Single write also waits for a lock.

> Every core has their own databases so only one core can access to database at the same time. Therefore, we can simply use sync.Lock between Get, All, Delete and Save instead transactions here. Just an idea.

It's not about the diffrent core. It's about one core handling several deploys. And transaction already has a lock so there is no need to create another one in the core.

> There is a small bug in the last commit. It doesn't get the alias when Delete called with an id.

Should be fixed now

ilgooz:
@krhubert It's a bit better to do atomic reads (with transactions) for _Get_ & _All_ as well.

database/service_db.go
```diff
@@ -52,59 +75,98 @@ func (d *LevelDBServiceDB) unmarshal(id string, value []byte) (*service.Service,
...
+	}
+
+	// encode service
 	b, err := d.marshal(s)
 	if err != nil {
 		return err
 	}
 
-	return d.db.Put([]byte(s.ID), b, nil)
+	batch := &leveldb.Batch{}
```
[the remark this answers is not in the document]
krhubert: batch are atomic, you can check out source code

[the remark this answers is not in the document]
krhubert: All is atomic as it works on snapschot. Get need a little fix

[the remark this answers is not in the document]
antho1404: The batch doesn't do transaction all the the time 
> // If the batch size is larger than write buffer, it may justified to write
	// using transaction instead. Using transaction the batch will be written
	// into tables directly, skipping the journaling.

Also if we can rely on batches and we are in this condition why not using it for the `Delete` function ?

```go
batch.Delete([]byte(aliasKeyPrefix + s.Alias))
if err := tx.Write(batch, nil); err != nil {
```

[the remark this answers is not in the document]
krhubert: In delete function we need to get the service first , which isn't possible with batches.

And about batch, this docs only apply to a transaction buffer size, but the overall batch is atomic in leveldb (you can check it reading code in depth or read orgninal leveldb docs (as this one from golang leveldb is rather poor)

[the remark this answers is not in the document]
NicolasMahe: I think we should use transaction. I read articles on batch and they said it's not really a transaction. Transactions give us more flexibility and more control. And also it lock the db from concurrent read and write.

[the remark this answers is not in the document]
NicolasMahe: Batch seems to either relaying on transaction if batch is too big, otherwise, it write in a single write (and of course wait for the db to be unlocked)

NicolasMahe:
@ilgooz > It's a bit better to do atomic reads (with transactions) for _Get_ & _All_ as well.

Transaction are only useful to group multiple reads and writes into a single atomic transaction.
If it only writes, batch is enough.
If it only one read, no need for transaction.
If it multiple reads, transactions are useful.

For the `All` function, i don't think we need transaction because `Iterator` is returning a snapshot of the database. So concurrent modifications are not affecting the iterator.

database/service_db.go
```diff
@@ -52,59 +75,105 @@ func (d *LevelDBServiceDB) unmarshal(id string, value []byte) (*service.Service,
...
+
+	batch := &leveldb.Batch{}
+	batch.Delete([]byte(idKeyPrefix + s.ID))
+	batch.Delete([]byte(aliasKeyPrefix + s.Alias))
+	if err := tx.Write(batch, nil); err != nil {
+		tx.Discard()
+		return err
+	}
+	if err := tx.Commit(); err != nil {
+		tx.Discard()
```
NicolasMahe: @krhubert @ilgooz what do you think of calling `tx.Discard` on error of `tx.Commit`?
The doc of `Commit` says:
> Commit commits the transaction. **If error is not nil, then the transaction is not committed, it can then either be retried or discarded**. Other methods should not be called after transaction has been committed.

database/service_db_test.go
```diff
@@ -60,22 +76,68 @@ func TestServiceDBDelete(t *testing.T) {
...
+	_, err = db.Get(s.ID)
+	require.IsType(t, &ErrNotFound{}, err)
+
+	_, err = db.db.Get([]byte(idKeyPrefix+s.ID), nil)
+	require.Equal(t, leveldb.ErrNotFound, err)
+	_, err = db.db.Get([]byte(aliasKeyPrefix+s.Alias), nil)
+	require.Equal(t, leveldb.ErrNotFound, err)
+}
+
+func TestServiceDBDeleteConcurrency(t *testing.T) {
```
NicolasMahe: I add this test in order to make sure that Delete function behave correctly on concurrent deletion.
If you remove the transaction in Delete function, then this test fails.

To test, replace in Delete function:
```
tx, err := d.db.OpenTransaction()
```
By
```
tx = db.db.
```
And remove `tx.Discard()` and `tx.Commit()`

NicolasMahe:
@mesg-foundation/core please review. I add few tests and change some minor code.
Please don't commit anthing on this PR

ilgooz:
> @ilgooz > It's a bit better to do atomic reads (with transactions) for _Get_ & _All_ as well.
> 
> Transaction are only useful to group multiple reads and writes into a single atomic transaction.
> If it only writes, batch is enough.
> If it only one read, no need for transaction.
> If it multiple reads, transactions are useful.
> 
> For the `All` function, i don't think we need transaction because `Iterator` is returning a snapshot of the database. So concurrent modifications are not affecting the iterator.

Here there is multiple reads https://github.com/mesg-foundation/core/blob/64407a543a353286229e8ba76d23f264d2abeeb4/database/service_db.go#L127

database/service_db.go
krhubert: https://github.com/syndtr/goleveldb/blob/master/leveldb/db_write.go#L263

Look here, they just commit a transaction, and in case of fali discard/retried is not needed

NicolasMahe: > OpenTransaction opens an atomic DB transaction. Only one transaction can be opened at a time. Subsequent call to Write and OpenTransaction will be blocked until in-flight transaction is **committed or discarded**. The returned transaction handle is safe for concurrent use.

👍

database/service_db_test.go
```diff
@@ -60,22 +76,68 @@ func TestServiceDBDelete(t *testing.T) {
...
+	defer closer()
+
+	s := &service.Service{ID: "00", Alias: "2", Name: "test-service"}
+	db.Save(s)
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	// service should not be found
+	go func() {
```
ilgooz: There is a chance that goroutine might complete before. Instead of checking _NotFound_ error here we should push it to an error slice and do the same for the other call to _Delete_. After _Wait_ done, we should compare the length of slice to see if it's equal to 1 (GOOD) or 2.

NicolasMahe: I understand what you mean but is it really necessary? In this case, it's a go routine versus the main routine. In the case of 2 go routine, then it will make sense. Maybe I should edit the code to call Delete from 2 go routine to have better concurrency?
@krhubert what do you think?

ilgooz: Actually, this is still two different goroutines, it doesn't matter that one is main. There are no difference.

PR #584, opened 2018-11-23
Log level debug for dev

### LINKED ISSUE #575: Support debug logs in dev
In development we need to support logs that are using `Debug` and not use `Info` for debug.
Some logs are important for dev but might be useless for production. eg: https://github.com/mesg-foundation/core/pull/574

antho1404:
dependency #577 

Set the core to run with the log level in debug.

We can now use `logrus.Debug` function

fix #575


antho1404:
@NicolasMahe @ilgooz this one is ready to merge and includes the changes if @krhubert from his PR #577
could be good to have your feedback on all that :)

dev-core
```diff
@@ -1,16 +1,56 @@
...
+CORE_SUM_PATH="./bin/.core.sum"
+touch "$CORE_SUM_PATH"
+
+# check if core bin was is cached
+CORE_SUM="$(openssl md5 ./bin/core)"
+CORE_SUM_PREV="$(cat $CORE_SUM_PATH)"
+if [[ "$CORE_SUM" == "$CORE_SUM_PREV" ]]; then
+  BINCACHED=true
+else
+  echo "$CORE_SUM" > "$CORE_SUM_PATH"
```
ilgooz:
```suggestion
  echo "$CORE_SUM" > "$CORE_SUM_PREV"
```

PR #588, opened 2018-11-27
use mockery to create Docker API mocks, a part of #587

ilgooz:


utils/docker/docker.go
```diff
@@ -0,0 +1,10 @@
+package dockermock
```
krhubert: Remove this file, you can generate mock directly from vendor

```mockery -name CommonAPIClient -dir vendor/github.com/docker/docker/client -inpkg```

File: scripts/build-mocks.sh [hunk 2]
```diff
@@ -9,6 +9,9 @@ cd $GOPATH/src/github.com/mesg-foundation/core
 # generate mocks for container package.
 mockery -name=Container -dir ./container -output ./container/mocks
 
+# generate mocks for docker.CommonAPIClient that used by container package.
+mockery -name=Docker -dir ./utils/docker -inpkg
```
krhubert: out file - mock_Docker.go -> don't use capital letters in filename.

ilgooz: I couldn't see a configuration related to this in mockery's cli. You can open an issue on their repo page. But I think it's OK this way.

utils/docker/docker.go
ilgooz: I didn't want to access to vendor folder directly. It's a bit messy and we can easily avoid it like with the current implementation.

scripts/build-mocks.sh
krhubert: `-case underscore`

ilgooz: I tried it and applied it to all other mocks but as I see, it messes with the interface names as well. For example, _ServiceDB.go_ mock file for _ServiceDB_ interface becomes to _service_db.go_ and I don't like this naming. I prefer to skip setting the case, it's not that important anyway.

NicolasMahe:
@krhubert can you review this pr please

Issue #589, opened 2018-11-27
Add gometalinter in CI


NicolasMahe:
So we can have a check in PR

PR #590, opened 2018-11-27
Add golang-lint tool

krhubert:
blocking/waiting for https://github.com/mesg-foundation/core/pull/580 first

closes https://github.com/mesg-foundation/core/issues/589

### DIFF
--- a/commands/service_dev.go
+++ b/commands/service_dev.go
@@ -46,7 +46,7 @@ func newServiceDevCmd(e ServiceExecutor) *serviceDevCmd {
 }
 
 func (c *serviceDevCmd) preRunE(cmd *cobra.Command, args []string) error {
-	c.path = getFirstOrDefault(args, "./")
+	c.path = getFirstOrDefault(args)
 	return nil
 }
 
@@ -72,7 +72,7 @@ func (c *serviceDevCmd) runE(cmd *cobra.Command, args []string) error {
 	if validationError != nil {
 		return xerrors.Errors{
 			validationError,
-			errors.New("To get more information, run: mesg-core service validate"),
+			errors.New("to get more information, run: mesg-core service validate"),
 		}
 	}
 	fmt.Printf("%s Service deployed with ID: %v\n", pretty.SuccessSign, pretty.Success(id))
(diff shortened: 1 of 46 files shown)


antho1404:
https://fixmie.com services like that may be a solution and it uses suggestions so that could be nice

NicolasMahe:
@krhubert test TestServiceStartError is failing now. can you investigate and fix please?

krhubert:
@antho1404 I saw it and I even request access for it, but still don't get it. But from thier adds this is standard go vet + go lint tools so everthing we have with golint

@NicolasMahe I was waiting for daemon to be merged, now I can continue it and fix all issues

commands/service_dev.go
```diff
@@ -82,8 +82,7 @@ func (c *serviceDevCmd) runE(cmd *cobra.Command, args []string) error {
 			err = c.e.ServiceDelete(id)
 		})
 		if err != nil {
-			fmt.Printf("%s Error while removing the service\n", pretty.FailSign)
-			fmt.Printf(pretty.Failln(err))
+			fmt.Printf("%s Removing the service error: %s\n", pretty.FailSign, err)
```
ilgooz:
```suggestion
			fmt.Printf("%s Removing the service completed with an error: %s\n", pretty.FailSign, err)
```

krhubert:
Yes, maybe it is simple and useful but it's really hard to tell without seeing the product as it requires access. Maybe is super hard and messy, maybe it is ... and so on.
For now, we have this tool which is also nice and gives colorized output (https://circleci.com/gh/mesg-foundation/core/4510). Also, you can run it on your localhost before commit which isn't possible with fixmie or simillar solutions

service/validation.go
```diff
@@ -18,9 +18,8 @@ func (p *ParameterWarning) String() string {
 // ValidateParametersSchema validates data to see if it matches with parameters schema.
 // TODO(ilgooz) add this as a method to Service type when create custom types for Event, Task etc.
 // TODO(ilgooz) remove pointer from *Parameter.
-func validateParametersSchema(parameters []*Parameter,
-	data map[string]interface{}) []*ParameterWarning {
-	warnings := make([]*ParameterWarning, 0)
+func validateParametersSchema(parameters []*Parameter, data map[string]interface{}) []*ParameterWarning {
+	warnings := make([]*ParameterWarning, 0, len(parameters))
```
ilgooz: Needed cap is not always fixed because invalid parameters' number can change. But I think it's ok to define the **max possible** cap here.

PR #592, opened 2018-11-28
Dockerfile.dev: install certs for secure connections

ilgooz:


krhubert:
I know I have asked, but remind me for which part exactly do we need certs. What is the pbm that certs are solving? Can you give a minimal example?

ilgooz:
> I know I have asked, but remind me for which part exactly do we need certs. What is the pbm that certs are solving? Can you give a minimal example?

We're using secure connections while cloning git repositories while using _https_ or _ssh_. If we don't have this trusted certs installed we won't be able to launch secure connections with Git servers and this opens the possibility of man-in-the-middle attacks.

For example when someone wants to deploy a service from a Git url and if there is an attacker in the middle, attacker can modify the source code as it's getting downloaded. This leads to major security issue but can be avoided by simply installing CA certs.

Issue #564, opened 2018-11-09
service: parameterValidator should support all types


ilgooz:
see [1](https://github.com/mesg-foundation/core/blob/d64035e78f60d3c378a61504b84a860f878978bf/systemservices/workflow/workflow.go#L20), [2](https://github.com/mesg-foundation/core/blob/d64035e78f60d3c378a61504b84a860f878978bf/systemservices/resolver/resolver.go#L18).

[_paramaterValidator_](https://github.com/mesg-foundation/core/blob/dev/service/validation.go#L57-L89) does not recognize struct types and any slice/map type except these two: 
_[]interface{}_, _map[string]interface{}_.

Please note that:

An interface that has a _[]string_ value cannot be asserted as _[]interface{}_ or:
An interface that has a _map[string]string_ value cannot be asserted as _map[string]interface{}_.
These can be only asserted to their *original* types like _[]string_ or _map[string]string_.
Also see https://play.golang.org/p/NVMb6YZh5E9

There are also no support for structs, slice or map of structs and nested structs.

In order to avoid type conversion to interface like in the examples 1 & 2 provided above or using _json.Unmarshal_ to do this with help of a pkg, we need to do reflection and more type assertions in the _validateType_ method to support a good range of types.

antho1404:
I'm not sure I understand this issue, can you provide more informations on the description please

ilgooz:
@antho1404 first post is updated

ilgooz:
@mesg-foundation/core Do we agree on improving it to support more types? Please give thumbs up to this post.

antho1404:
@ilgooz is this issue still valid ? we have new types for the data sent through grpc and we don't have the systemservices anymore. I think this one can be close

PR #596, opened 2018-11-28
possibility to have errors in result data

### LINKED ISSUE #573: service execute command hangs forever on unexpected data type
This happens when received result data has a field with an unexpected type.
For example if a task's _success_ output has a field with _name(string)_ defined in the yml and service responses with _name(bool)_, cli command hangs forever on the following state:

```
✔ Task "findLocation" executed
⣷ Waiting for result... 
```

### LINKED ISSUE #610: Core shouldn't allow multiple output for the same execution
Based on this discussion https://github.com/mesg-foundation/core/issues/562#issuecomment-443436893, the core doesn't return any error when a task is submitting multiple results to the core.

Each task can return only one result per execution. All the following results should be ignored and error should be triggered.

Services should follow a good practice and not return errors from the the `submitResult` api. These errors should be treated differently.

ilgooz:
fixes #610, fixes #573.


protobuf/coreapi/api.proto
```diff
@@ -157,9 +157,23 @@ message ListenResultRequest {
...
+    Output output = 6;   // output is filled when there is matching task execution that completed successfully.
+    Error error = 7;     // error is filled when there is a matching task execution that cannot be completed. 
+  }
+
+  message Output {
+    string key = 3;   // The output's key from the returned task.
+    string data = 4;  // The output's data from the returned task, encoded in JSON.
+  }
+  
+  message Error {
```
krhubert: `message Error` should be outside Result Data as it will be used for diffrent error (like for ListenEvent)

protobuf/coreapi/api.proto
```diff
@@ -157,9 +157,23 @@ message ListenResultRequest {
...
+    string message = 1;
+  }
+
+  string outputKey = 3;   // [Deprecated] The output's key from the returned task.
```
krhubert: I would love to remove this, as it will force us to update services, otherwise we will stuck with it for a long time

protobuf/coreapi/api.proto
```diff
@@ -157,9 +157,23 @@ message ListenResultRequest {
 message ResultData {
   string executionID = 1;             // The unique identifier of the execution.
   string taskKey = 2;                 // The key of the executed task.
-  string outputKey = 3;               // The output's key from the returned task.
-  string outputData = 4;              // The output's data from the returned task, encoded in JSON.
   repeated string executionTags = 5;  // The list of tags associated with the execution
+  oneof result {
```
krhubert: The same should be done to lisent event

commands/service_execute.go
```diff
@@ -105,11 +105,20 @@ func (c *serviceExecuteCmd) runE(cmd *cobra.Command, args []string) error {
...
 	}
-	fmt.Printf("Task %s returned output %s with data:\n%s\n",
-		pretty.Success(c.taskKey),
-		pretty.Colorize(pretty.FgCyan, result.OutputKey),
-		pretty.ColorizeJSON(pretty.FgCyan, nil, true, []byte(result.OutputData)),
-	)
+	errResult := result.GetError()
+	outputResult := result.GetOutput()
+	if errResult != nil {
+		fmt.Printf("Task %s cannot be completed, error:\n%s\n",
```
krhubert: Put a fail sign and remove colors to keep consistency with others commands printf

commands/service_dev.go
```diff
@@ -130,11 +130,20 @@ func (c *serviceDevCmd) runE(cmd *cobra.Command, args []string) error {
...
 		case r := <-listenResultsC:
-			fmt.Printf("Receive result %s %s: %s\n",
-				pretty.Success(r.TaskKey),
-				pretty.Colorize(color.New(color.FgCyan), r.OutputKey),
-				pretty.ColorizeJSON(pretty.FgCyan, nil, false, []byte(r.OutputData)),
-			)
+			errResult := r.GetError()
+			outputResult := r.GetOutput()
+			if errResult != nil {
+				fmt.Printf("Receive execution error on %s task: %s\n",
```
krhubert: Put a fail sign and remove colors to keep consistency with others commands printf

commands/service_dev.go
```diff
@@ -130,11 +130,20 @@ func (c *serviceDevCmd) runE(cmd *cobra.Command, args []string) error {
...
+					pretty.Fail(r.TaskKey),
+					pretty.Fail(errResult.Message),
+				)
+			} else {
+				fmt.Printf("Receive result %s %s: %s\n",
```
krhubert: Put a success sign and remove colors to keep consistency with others commands printf

commands/service_execute.go
```diff
@@ -105,11 +105,20 @@ func (c *serviceExecuteCmd) runE(cmd *cobra.Command, args []string) error {
...
+			pretty.Fail(c.taskKey),
+			pretty.Fail(errResult.Message),
+		)
+	} else {
+		fmt.Printf("Task %s returned output %s with data:\n%s\n",
```
krhubert: Put a success sign and remove colors to keep consistency with others commands printf

api/submit_result.go
```diff
@@ -32,12 +34,14 @@ func (s *resultSubmitter) Submit(executionID string, outputKey string, outputDat
...
 	}
-	if err := exec.Complete(outputKey, outputData); err != nil {
-		return err
-	}
-	if err = s.api.execDB.Save(exec); err != nil {
-		return err
+
+	if err = exec.Complete(outputKey, outputData, rerr); err != nil {
+		exec.Error = err
+	} else {
```
krhubert: s/else/else if 

If there is only one if inside

krhubert: the mroe I think of it, the more it gets complicated, in general we should have 2 error - one for service and one for internal.

For example here if save to db fail , then it's for sure internal error because task execution is fine. The core just failed.

This error is different from the error that service return. For now, we can leave it, but not sure if it's ok

@NicolasMahe @ilgooz

protobuf/coreapi/api.proto
NicolasMahe: Are you sure it's a breaking change to move it to a oneof message?
```
 message Output {
    string key = 3;   // The output's key from the returned task.
    string data = 4;  // The output's data from the returned task, encoded in JSON.
  }
```

If the number are the same, the client should be able to decode it. But yes, the client will not be compatible with the error (current behavior anyway)

protobuf/coreapi/api.proto
NicolasMahe: I agree. We can create custom message in gRPC and reuse them anywhere

protobuf/coreapi/api.proto
```diff
@@ -157,9 +157,23 @@ message ListenResultRequest {
...
+    Error error = 7;     // error is filled when there is a matching task execution that cannot be completed. 
+  }
+
+  message Output {
+    string key = 3;   // The output's key from the returned task.
+    string data = 4;  // The output's data from the returned task, encoded in JSON.
+  }
+  
+  message Error {
+    string message = 1;
```
NicolasMahe: are you sure that using number '1' is ok? isn't override `executionID`?
But I'm not sure how grpc 'message' works.

protobuf/coreapi/api.proto
ilgooz: Let's do this when we need it. It won't be a breaking change when we do this in future. Because we may not need to use _Error_ inside _ListenEvent_. We should first check to see if we need to have errors inside listen event like we need it here.

protobuf/coreapi/api.proto
ilgooz: Can you explain more please?

protobuf/coreapi/api.proto
ilgooz: This is inside another message _(Error)_ so both field that uses _1_ is in separate contexts. There should be no problems with this use.

```
. executionID (1)
. error (7) > message (1)
```

api/submit_result.go
ilgooz: else-if seems better

ilgooz: I agree, that there is two errors as one Core's internal and second is about the invalid behavior of  A Service.

To keep things simple I didn't create two types errors, actually I give it a try but code become unnecessarily complex. I'd like to experiment current implementation for now and after we can add a field like _type_ to _Error_ gRPC message type to indicate the type of error. This won't be a breaking change anyway.

ilgooz: Also, we currently show only the produced error message we don't provide the inputs that are caused to error as a meta info to error. So we may also want to add field like maybe with a generic name like _data_ to keep them alongside _type_ and _message_. So client may print it to inform users. For this case _data_ can contain the _outputKey_ and _outputData_ because any of them can be sent as invalid by server.

protobuf/coreapi/api.proto
krhubert: Yep, ListenEvent is also a stream and by analogy, it might return an error. So ListenEventRequest 
should have oneof { Result, Error } as well

commands/service_execute.go
ilgooz: done

commands/service_execute.go
ilgooz: done

commands/service_dev.go
ilgooz: I think we shouldn't have it here because this is a log stream and there will be too much noise with signs if we put them. What do you think?

commands/service_dev.go
ilgooz: I think we shouldn't have it here because this is a log stream and there will be too much noise with signs if we put them. What do you think?

protobuf/coreapi/api.proto
ilgooz: It's a breaking change like explained here https://github.com/mesg-foundation/core/pull/596#discussion_r237433781, they're defined in different levels. One of them in the root (outputKey) and other is under another field (output) as _output.key_

protobuf/coreapi/api.proto
ilgooz: @NicolasMahe We may extend this specific _Error_ type with another fields like _type_, _data_ that talked about other comments in this PR. So I think that we may not be able to use a generic error type here.

ilgooz: Can I get more feedbacks about the direction of this discussion?

protobuf/coreapi/api.proto
krhubert:
```proto
message ListenEventRequest {
  message Result {
    string serviceID = 1; 
    string eventFilter = 2;
  }
  oneof result {
    Result result = 3;
    Error error = 4;
  }
}
```

commands/service_dev.go
krhubert: Ok

commands/service_dev.go
krhubert: Ok

protobuf/coreapi/api.proto
NicolasMahe: Ok, as you said, let's do it when we need it if it will not break anything to move it later ;)

protobuf/coreapi/api.proto
NicolasMahe: So are you planning to send both (root and output) to the client for now?

protobuf/coreapi/api.proto
NicolasMahe: I quite agree with hubert:
if the service send a not valid event to the core, should we notify the app as well? it could be good yes.
@antho1404 what do you think about this?

commands/service_dev.go
NicolasMahe: I agree with hubert, it will look nicer in the console

commands/service_dev.go
NicolasMahe: I agree with hubert, it will look nicer in the console

api/submit_result.go
NicolasMahe: let's keep it simple for this PR. let's see if it useful and if it is, let's do it in another PR

protobuf/coreapi/api.proto
ilgooz: As this one is not urgent we should merge this PR to fix an actual bug in the codebase. Let's create an another issue for this one and move the discussion there.

ilgooz: moved to https://github.com/mesg-foundation/core/issues/600

protobuf/coreapi/api.proto
ilgooz: Of course, otherwise if we remote the *old ones* from *root* it'd be braking change for other clients.

NicolasMahe: 👍

ilgooz:
In order to

krhubert:
Fix lint errors :)

antho1404:
This PR might not be necessary according to this comment https://github.com/mesg-foundation/core/issues/610#issuecomment-443533797.

We might be solving to issues with one single solution and this current solution will not be necessary.

ilgooz:
> This PR might not be necessary according to this comment [#610 (comment)](https://github.com/mesg-foundation/core/issues/610#issuecomment-443533797).
> 
> We might be solving to issues with one single solution and this current solution will not be necessary.

Why do you think that this PR might not be necessary? When a service sends an invalid result to Core, an application that listens for that result will never have it. In the mean while Core will witness this. So, it can inform application by submitting the result but with the actual error that raised from _SubmitResult_ API which is the purpose of this PR. I read the #610 but I couldn't understand it fully. I'll give a second look tomorrow.

NicolasMahe:
I'm removing this PR from v0.5 in order to do more test and to implement the changes in the application libraries.

antho1404:
This PR could be done a bit differently to include also a fix https://github.com/mesg-foundation/core/issues/610. On this issue, if we have the error in the execution the problem is solved so what about having this error inside the execution instead of outside and in the `api.SubmitResult` set the error in the execution through a `Failed` function in the execution.

Like that the execution is marked as failed and all the time send to the stream and also all the time stored in the database to be able to extract these data if we need to.

I have the modifications of this changes, I can open a PR just to explain with code if it helps

ilgooz:
Currently we pass an error in the _execution.Complete()_ func and if it's not nil execution will complete with an error. I don't mind extracting this part to a separate _ComleteWithFail()_ but I cannot see a benefit.

ilgooz:
As I checked, this PR already includes #610. https://github.com/mesg-foundation/core/issues/610#issuecomment-445720962

ilgooz:
Having an additional failed status is not mandatory -because we already have this _Error_ field that we can check if execution completed with failure- but would be nice.

Passing string value to _api.SubmitResult()_ still fees a bit odd because that will make submit result api to be JSON aware. Keeping this decoding part in the _interface_ pkg where we deal with network feels more natural because JSON is used to transmit data over network so it's more relevant to keep it there. We also do the encoding part of listen result, event, task inside _interface_ pkg as well. Only moving the decoding part to _api_ pkg also feels odd.

Maybe we can discuss/solve this one under another issue because this is just an improvement. But still, if you really think that we should do these changes, I'm glad to do the updates. 👍

antho1404:
This PR might also solve this issue https://github.com/mesg-foundation/core/issues/637

ilgooz:
needs review

api/submit_result.go
```diff
@@ -1,12 +1,16 @@
...
+	"errors"
+	"fmt"
+
 	"github.com/mesg-foundation/core/pubsub"
 	"github.com/mesg-foundation/core/service"
 )
 
 // SubmitResult submits results for executionID.
-func (a *API) SubmitResult(executionID string, outputKey string, outputData map[string]interface{}) error {
+func (a *API) SubmitResult(executionID string, outputKey string, outputData interface{}) error {
```
antho1404: We talked in the meeting to use `[]byte` here. I could be quite ok with the `interface{}` which was my initial idea too. I'm curious to know why you put that ? Is their any places we still send a `map[string]interface{}`, I only see the grpc api. If there is only this I think we shouldn't overcomplicated and just keep one kind of data instead of having to manage string and map[string]interface{}.

But here again i'm more curious about the reason of that.

execution/execution_test.go
```diff
@@ -146,7 +165,39 @@ func TestComplete(t *testing.T) {
...
+	e, _ := New(srv, eventID, taskKeyWithError, map[string]interface{}{"foo": "1", "bar": "2"}, tags)
+	e.Execute()
+	tests := []struct {
+		name string
+		key  string
+		data map[string]interface{}
+		xerr error
+		err  error
+	}{
+		{name: "1", key: "outputX", data: map[string]interface{}{"foo": "bar"}, xerr: errors.New("cannot complete"), err: nil},
```
krhubert: Change name "1" and "2" to the description of a testing case if you want to guide someone in tests.

api/submit_result.go
```diff
@@ -32,12 +36,32 @@ func (s *resultSubmitter) Submit(executionID string, outputKey string, outputDat
 	if err != nil {
 		return err
```
antho1404: We can still wrap this error in the execution. This will happen when we delete a service that is running an execution. The execution will fail but we might for example have a restart policy and when we reinstall the service it will be completed or this kind of behaviors that are possible if we store this error

execution/execution_test.go
```diff
@@ -146,7 +165,39 @@ func TestComplete(t *testing.T) {
...
+			ActualStatus:   Failed,
+		}}, // this one is already proccessed
+	}
+	for _, test := range tests {
+		err := e.Failed(test.xerr)
+		require.Equal(t, test.err, err, test.name)
+		if test.err != nil {
+			continue
+		}
+		require.NotNil(t, e)
```
krhubert: Remove NotNil(t,e) as e is already used and it canot be nil.

execution/execution_test.go
```diff
@@ -146,7 +165,39 @@ func TestComplete(t *testing.T) {
...
+		require.NoError(t, err, test.name)
+		require.NotNil(t, e, test.name)
```
krhubert: The same notNil test as above

execution/execution_test.go
```diff
@@ -146,7 +165,39 @@ func TestComplete(t *testing.T) {
...
+		}}, // this one is already proccessed
+	}
+	for _, test := range tests {
+		err := e.Failed(test.xerr)
+		require.Equal(t, test.err, err, test.name)
+		if test.err != nil {
+			continue
+		}
+		require.NotNil(t, e)
+		require.NoError(t, err, test.name)
```
krhubert: This one is also already checked in L192

api/submit_result.go
```diff
@@ -32,12 +36,32 @@ func (s *resultSubmitter) Submit(executionID string, outputKey string, outputDat
...
 	}
-	if err := exec.Complete(outputKey, outputData); err != nil {
-		return err
+
+	// parse output data into a map.
+	var outputDataMap map[string]interface{}
+	switch d := outputData.(type) {
+	case map[string]interface{}:
+		outputDataMap = d
+	case string:
```
krhubert: Pass json as slice of []bytes as it's a standard way for passing json data. it's standard because every pkg handles json expected bytes. Moreover, convert []bytes to string generates small overhead

api/submit_result.go
```diff
@@ -32,12 +36,32 @@ func (s *resultSubmitter) Submit(executionID string, outputKey string, outputDat
 	if err != nil {
 		return err
 	}
-	if err := exec.Complete(outputKey, outputData); err != nil {
-		return err
+
+	// parse output data into a map.
+	var outputDataMap map[string]interface{}
+	switch d := outputData.(type) {
```
krhubert: Change outputData to slice of []bytes as it looks like a dynamic programming language. Here we should expect only one type as discussed.

api/submit_result.go
```diff
@@ -32,12 +36,32 @@ func (s *resultSubmitter) Submit(executionID string, outputKey string, outputDat
...
+		exec.Failed(err)
+	}
+
 	if err = s.api.execDB.Save(exec); err != nil {
 		return err
 	}
+
 	go pubsub.Publish(exec.Service.ResultSubscriptionChannel(), exec)
-	return nil
+	return err
```
antho1404: We could delegate all this error handling to another function to make sure we always publish the result, here we have a return on the save that might be necessary (and i think it's good to have it) but we might add some stuff in between and forget that this function should publish almost all the time.

I was thinking of something like that 
https://github.com/mesg-foundation/core/blob/05dd0762eb9dd37ece4d66f3bac68fff4875f201/api/submit_result.go#L30-L37

commands/service_execute.go
```diff
@@ -105,11 +105,22 @@ func (c *serviceExecuteCmd) runE(cmd *cobra.Command, args []string) error {
...
 	}
-	fmt.Printf("Task %s returned output %s with data:\n%s\n",
-		pretty.Success(c.taskKey),
-		pretty.Colorize(pretty.FgCyan, result.OutputKey),
-		pretty.ColorizeJSON(pretty.FgCyan, nil, true, []byte(result.OutputData)),
-	)
+	errResult := result.GetError()
+	outputResult := result.GetOutput()
+	if errResult != nil {
+		fmt.Printf("%s Task %s cannot be completed, error:\n%s\n",
```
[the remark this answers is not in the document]
antho1404: @krhubert please use github suggestions for these kind of feedback, that saves a lot of time

interface/grpc/service/submit_result_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"testing"
 
+	"github.com/mesg-foundation/core/x/xerrors"
+
```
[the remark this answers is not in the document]
ilgooz: done

commands/service_execute.go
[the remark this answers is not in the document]
ilgooz: done

execution/execution_test.go
```diff
@@ -146,7 +165,39 @@ func TestComplete(t *testing.T) {
...
+		require.Zero(t, e.Error, test.name)
+	}
+}
+
+func TestFailed(t *testing.T) {
+	e, _ := New(srv, eventID, taskKeyWithError, map[string]interface{}{"foo": "1", "bar": "2"}, tags)
+	e.Execute()
+	tests := []struct {
+		name string
+		key  string
```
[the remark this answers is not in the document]
ilgooz: done

execution/execution_test.go
ilgooz: done

execution/execution_test.go
ilgooz: done

execution/execution_test.go
ilgooz: done

execution/execution_test.go
ilgooz: done

protobuf/coreapi/api.proto
antho1404: I'm a bit worried on this one, I just want to make sure that this is not breaking for the existing libraries (it's ok if that breaks when there is an error but not when there is a success.

We should make sure that libraries are always compatible with that or at least update the libraries and make them as breaking release if needed.

api/submit_result.go
ilgooz: done

api/submit_result.go
ilgooz: done

api/submit_result.go
ilgooz: for the WSS we faced with a similar problem where which explained here https://github.com/mesg-foundation/core/issues/564. But we don't need to accept here in the _service_ api as you said because we don't make calls to inside Core, only service clients use it. So, I'm changing it to bytes and we can always switch to interface if needed.

api/submit_result.go
ilgooz: I cannot follow what you mean exactly. Do you mean calling _Failed_ by "wrapping" This error here is raised when _FromService_ is unsuccessful which it would be an internal error of core if it's not nil. I also checked the implementation of _FromService_ and it always returns nil errors for now.

api/submit_result.go
ilgooz: done

ilgooz: actually, i applied but reverted the changes you proposed. because if there is an error we still may need to publish it because it can be about unmarshalling the output data or invalid output key etc. here in your proposition we don't publish in case of error which we should sometimes.

protobuf/coreapi/api.proto
ilgooz: I tested manually and it works. We should have integration tests to be sure https://forum.mesg.com/t/automatize-manual-testing-for-core-true-integration-testing/94/7

antho1404: Why not having something like that
```proto
message ResultData {
  string executionID = 1;
  string taskKey = 2;
  string outputKey = 3;
  repeated string executionTags = 5;  
  
  oneof result {
    string outputData = 4;
    Error error = 7;
  }

  message Error {
    string message = 1;
  }
}
```

We can actually provide the `outputKey` even when there is an error so let's give more context and like that we don't need to deprecate or change anything on the api, just add the error fields.

ilgooz: two catches here:

* output key might be invalid send by the client. (not defined in the yml)
* output data might be a bad json so it cannot be unmarshaled. therefore we'll not be able to set it _execution.OutputData_ which is not convenient.

antho1404: outputKey can be invalid the point here is that it will always be here (even if empty). If invalid we will have the error as result.

For the second part I don't really understand, it's either we can parse it and we have the outputData or we cannot parse it and we get the error but i'm not sure I understand what you say

ilgooz: what i mean is, even if we can provide the output key when there is an error we never can do it for the output data. because we cannot parse it to a map. because of that we shouldn't set both of them. invalid key and wrong formatted output data can be put inside the error message itself.

api/submit_result.go
NicolasMahe: I guess @antho1404 is saying:
```
if err != nil {
	exec.Fail(err)
	return exec, nil
}
```

NicolasMahe: It's to change the status of the exec to fail and publish it anyway.

ilgooz: >  We should not fail executions on internal errors because service may re-try. I also checked the implementation of FromService and it always returns nil errors for now.

NicolasMahe: The logic @antho1404 want to push is that's not a service's error.
The service did it job, it executed the task and provide a result.
Any errors that happen on the core should not be returned to the service but should be return the "user" by changing the state of the execution to failed.

ilgooz: now, this makes sense 👍

api/submit_result.go
ilgooz: I re-implemented this again because it's now possible after the changes made for https://github.com/mesg-foundation/core/pull/596#discussion_r240911881. But still, we don't return on error and always publish it.

api/submit_result.go
```diff
@@ -23,21 +27,40 @@ func newResultSubmitter(api *API) *resultSubmitter {
...
+	go pubsub.Publish(exec.Service.ResultSubscriptionChannel(), exec)
+	return err
+}
+
+// processExecution processes execution and marks it as complated or failed.
+func (s *resultSubmitter) processExecution(executionID string, outputKey string, outputData []byte) (*execution.Execution, error) {
 	exec, err := s.api.execDB.Find(executionID)
 	if err != nil {
-		return err
+		exec.Failed(err)
```
NicolasMahe: exec is nil in this case.

```suggestion
```

api/submit_result.go
```diff
@@ -23,21 +27,40 @@ func newResultSubmitter(api *API) *resultSubmitter {
...
+		return nil, err
 	}
+
 	exec.Service, err = service.FromService(exec.Service, service.ContainerOption(s.api.container))
 	if err != nil {
-		return err
+		exec.Failed(err)
+		return nil, err
```
NicolasMahe: It should return exec and not the error.
The error is contained in the the exec and will be publish by the parent function.

api/submit_result.go
```diff
@@ -23,21 +27,40 @@ func newResultSubmitter(api *API) *resultSubmitter {
...
+		if err = exec.Complete(outputKey, outputDataMap); err != nil {
+			exec.Failed(err)
+		}
 	}
-	if err = s.api.execDB.Save(exec); err != nil {
-		return err
+
+	if err := s.api.execDB.Save(exec); err != nil {
+		exec.Failed(err)
+		return nil, err
```
NicolasMahe: Same same

api/submit_result.go
ilgooz: removed

api/submit_result.go
ilgooz: I made a change to both return the exec and err but it's not idiomatic to return a not-nill type when an err returned. We should improve this in future.

api/submit_result.go
ilgooz: https://github.com/mesg-foundation/core/pull/596#discussion_r242036717

api/submit_result.go
```diff
@@ -42,24 +45,30 @@ func (s *resultSubmitter) processExecution(executionID string, outputKey string,
...
 	var outputDataMap map[string]interface{}
-	if err = json.Unmarshal(outputData, &outputDataMap); err != nil {
-		err = fmt.Errorf("invalid output data error: %s", err)
-		exec.Failed(err)
-	} else {
-		if err = exec.Complete(outputKey, outputDataMap); err != nil {
-			exec.Failed(err)
-		}
+	if err := json.Unmarshal(outputData, &outputDataMap); err != nil {
+		return exec, s.failedAndSave(exec, fmt.Errorf("invalid output data error: %s", err))
```
ilgooz: notes:

* error here sent to workflow but same error not sent to return value of gRPC handler. gRPC handler should get the same error but it's overwritten by the error returned from _ failedAndSave()_.

NicolasMahe: I changed the logic a little bit with the introduction of `func saveExecution`.
In `func saveExecution`, error on `exec.Failed` and `execDB.Save` are prioritize over possible error from parameters.

antho1404:
Should we close this one in favor of https://github.com/mesg-foundation/core/pull/651

Issue #382, opened 2018-08-22
service: service.Start() can block forever


ilgooz:
Similar to https://github.com/mesg-foundation/core/pull/381.

If dependencies from same service listens from the same port, service.Start() will block forever because it waits for status _RUNNING_ for dependency containers. However, _N-1_ of them will never be in _RUNNING_ state because ports collides.

Also any other misconfiguration on _Service_ or _service file (mesg.yml)_ can cause the same issue.

antho1404:
In this case one of the task created by the docker service will never trigger the task and so we will never get the error from that. We need to find a way to get the errors between the creation of the service (that is fine) and before the creation of the task

ilgooz:
future note: add a test afterwards to _service/start_test.go_
```go
func TestX(t *testing.T) {
    service := &Service{
        Name: "TestServiceDependenciesListensFromSamePort",
        Dependencies: map[string]*Dependency{
            "test": {
                Image: "nginx",
                Ports: []string{"80"},
            },
            "test2": {
                Image: "nginx",
                Ports: []string{"80"},
            },
        },
    }
    fmt.Println(service.Start()) // currently blocks forever
}
```

NicolasMahe:
Let's check if the error is still here

krhubert:
The pbm stills exist

NicolasMahe:
So,
The service that fail to start has one task stuck in "new" status.

What about adding a timeout on `container.waitForStatus` function?

Otherwise, we could add a more sophisticated system that consider a task as error if it didn't change status for let's say 1min.

NicolasMahe:
I really think we should keep the restart always. It's a really nice feature for the service to restart if any error happen.
We have a pretty good error handling so far. But for this issue, docker have a weird behavior.

NicolasMahe:
Actually, I have a solution:
This problem occurs only when the 2 docker services are start at the same time. If the services are start one by one, then the second return an error. Eg:
```go
package service

import (
	"fmt"
	"testing"
	"github.com/mesg-foundation/core/container"
)

func TestX(t *testing.T) {
	c, _ := container.New()
	service, _ := FromService(&Service{
		Name: "TestServiceDependenciesListensFromSamePort1",
		Dependencies: []*Dependency{
			{
				Key:   "1",
				Image: "nginx",
				Ports: []string{"80"},
			},
		},
	}, ContainerOption(c))
	service2, _ := FromService(&Service{
[...]

PR #598, opened 2018-11-28
Start service dependencies one by one

krhubert:
closes https://github.com/mesg-foundation/core/issues/382


NicolasMahe:
What is the order of start?
Does the service start before the dependencies or the other way?

I would say it makes sense to start the dependencies first. (as the name suggest, the service is dependent on them).

krhubert:
Before there was no order of starting. And it's the pbm because some dependencies may eg. connect to service and the other way and the other way, so it's hard to tell the one perfect order.

For example you deploy website as servcie and has db as a dependency so service must start first.

Second, you deploy database and eg. crawlers as a dependency in that case db must start first so crawler can connect to it

NicolasMahe:
For info, the dependencies are started before the service itself. LGFM

antho1404:
Why the dependencies are starting before the service ? This is just based on a map so it shouldn't by default, this needs to be added for me, I don't see how it can be done with the code here

NicolasMahe:
Because of the way we transform the yaml to the go struct. See end of func deploy:
https://github.com/mesg-foundation/core/blob/fcea30b78145f1b27a3979555418560745a8186f/service/service.go#L193-L207

PR #601, opened 2018-11-29
Change error msg to more explaining one

### LINKED ISSUE #562: weird behaviour on task execution with invalid output data
to reproduce:
```go
// 1-
$ ./dev-core

// 2-
$ ./dev-cli service dev https://github.com/mesg-foundation/service-webhook

// 3-
$ ./dev-cli service execute -d url=https://mesg.tech -d data={} -d headers={} 4f7891f77a6333787075e95b6d3d73ad50b5d1e9
// outputs:
✔ Task "call" executed
Task call returned output error with data:
{
  "message": "Error: 2 UNKNOWN: json: cannot unmarshal string into Go value of type map[string]interface {}"
}
```

**3-** shouldn't output with _error_ task key. Because error message is belongs to a Go program whereas _service-webhook_ written in nodejs. I think this error caused from [here](https://github.com/mesg-foundation/core/blob/d23e60d/interface/grpc/service/submit_result.go#L13) but I don't know why it's combined with _error_ output key instead of having just a plain gRPC error.

**1-** One other weird thing related to this is _dev-core_ logs _SubmitResult_ twice as _error_ and _info_. Where I just expect one _error_ log:
```
time="2018-11-01T08:46:22Z" level=error msg="finished unary call with code Unknown" error="json: cannot unmarshal string into Go value of type map[string]interface {}" grpc.code=Unknown grpc.method=SubmitResult grpc.service=api.Service grpc.start_time="2018-11-01T08:46:22Z" grpc.time_ms=28.074 span.kind=server system=grpc
time="2018-11-01T08:46:22Z" level=info msg="finished unary call with code OK" grpc.code=OK grpc.method=SubmitResult grpc.service=api.Service grpc.start_time="2018-11-01T08:46:22Z" grpc.time_ms=1.148 span.kind=server system=grpc
```

krhubert:
close #562 

It turned out that https://github.com/mesg-foundation/service-webhook/blob/master/index.js#L16 here the output is a string (so invalid) and I've changed error message just. With the core everything is fine.

### DIFF
--- a/interface/grpc/service/submit_result.go
+++ b/interface/grpc/service/submit_result.go
@@ -3,6 +3,7 @@ package service
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 
 	"github.com/mesg-foundation/core/protobuf/serviceapi"
 )
@@ -11,7 +12,7 @@ import (
 func (s *Server) SubmitResult(context context.Context, request *serviceapi.SubmitResultRequest) (*serviceapi.SubmitResultReply, error) {
 	var data map[string]interface{}
 	if err := json.Unmarshal([]byte(request.OutputData), &data); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("servcie sent invalid json data in %s output", request.OutputKey)
 	}
 	return &serviceapi.SubmitResultReply{}, s.api.SubmitResult(request.ExecutionID, request.OutputKey, data)
 }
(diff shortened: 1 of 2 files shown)


NicolasMahe:
@krhubert, there was a typo on the error message: 
```
return nil, fmt.Errorf("servcie sent invalid json data in %s output", request.OutputKey)
```
Shouldn't the lint return an error?
Does gometalinter do it and not golangci-lint?

PR #606, opened 2018-11-30
Disable workflows command for release v0.5

NicolasMahe:
Closes https://github.com/mesg-foundation/core/issues/602


krhubert:
Fix lint errors :)

antho1404:
To disable commands you could also simply add the property `Hidden: true` to the `newRootWorkflowCmd` like that there is only this to change and don't need to comment everything, the feature will be hidden (but still accessible for the ones who know it like on the dev). This visibility could also be linked to an env variable eg: `MESG_FEATURE_WORKFLOW=1`

PR #609, opened 2018-12-01
Delete all service in parallel in commands

### LINKED ISSUE #603: commands: delete 'all' services in parallel
https://github.com/mesg-foundation/core/blob/a4950ea/commands/provider/service_provider.go#L40-L53

this code is deleting all services in series but this is a waste on IO time so they should be done in parallel

krhubert:
close #603


x/xerrors/errors.go
```diff
@@ -23,3 +26,16 @@ func (e Errors) Error() string {
 
 	return strings.Join(s, "\n")
 }
+
+// SyncErrors is an error for tracing multiple errors safe to use in multiple goroutines.
+type SyncErrors struct {
```
ilgooz: I prefer to have _Append_ and _AppendSync_ methods for _Error_ type. I think we don't quite need two types here.

PR #611, opened 2018-12-01
Switch to go1.11

### LINKED ISSUE #468: Switch to go 1.11
Switch to go 1.11 , which leads to move from `dep` to `go modules`

krhubert:
close #468 

Are you ready to switch to go 1.11?! 😍😍😍😂 

List of changes:
- switch to go 1.11 in circleci
- replace Gopkg.* with go.mod and go.sum
- use `go mod` instead of dep
- add tools.go to keep tools vendored


ilgooz:
Do we want to keep vendor/ folder or not? I'd like to remove it also but we should create an another repo and keep vendors there and update it with a CI task (for master branch).  We don't need to pull it from there, it is just for securing the vendors. This way if some of them gets deleted from the internet we'll be able recover.

PR #613, opened 2018-12-02
support deleting data volumes while deleting services, closes #422

### LINKED ISSUE #422: Delete docker's volume when deleting a service
This PR https://github.com/mesg-foundation/core/pull/419 uses Docker Volume instead of binding path to the host machine.
But Docker doesn't delete any volume automatically.

I suggest to remove the docker's volume when a service is deleted by the API `deleteService`:

https://github.com/mesg-foundation/core/blob/060812cfe016838c4a29306d720eb1103d7d5e88/api/delete_deleter.go#L18-L27

Also, we could provide a flag to force to re-create the core volume in the command `start` like: `--reset`.

ilgooz:


api/deploy.go
```diff
@@ -139,6 +139,10 @@ func (d *serviceDeployer) deploy(r io.Reader) (*service.Service, *importer.Valid
 	if validationErr != nil {
 		return nil, validationErr, nil
 	}
+
+	if err := s.CreateVolumes(); err != nil {
```
antho1404: Why do we need to create volumes here, they are dynamically created when we start a service so I don't see the benefits of adding this here (and the whole service/volumes file)

ilgooz: I added to set some labels on the volume but after I removed the labels. I'm deleting the creation of volumes part.

commands/service_delete.go
```diff
@@ -41,7 +43,7 @@ func (c *serviceDeleteCmd) preRunE(cmd *cobra.Command, args []string) error {
 	}
 
 	if !c.all || (c.all && c.force) {
-		return nil
+		return c.askDeleteData()
```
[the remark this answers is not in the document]
ilgooz: It has a different [purpose](https://github.com/mesg-foundation/core/blob/a4950ea/commands/service_delete.go#L34). We have a _-delete-data_ flag for force deletion of data volumes.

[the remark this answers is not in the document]
krhubert: But it's unclear, the `--force` is a standard cli flag to avoid any confirmation during command

api/delete.go
```diff
@@ -34,5 +38,13 @@ func (d *serviceDeletor) Delete(serviceID string) error {
 	if err := s.Stop(); err != nil {
 		return err
 	}
+	// delete volumes first before the service. this way if
+	// deleting volumes fails, process can be retried by the user again
+	// because service still will be in the db.
+	if deleteData {
```
NicolasMahe: Why not always deleting the volumes?
If an user delete a service, it's natural to delete the associated volume.

commands/service_delete.go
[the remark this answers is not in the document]
NicolasMahe: > Why not always deleting the volumes?
> If an user delete a service, it's natural to delete the associated volume.

If volumes are always deleting when service are deleting, then the `--force` flag is enough to bypass the confirmation.

api/delete.go
ilgooz: In the real world usage data should not be deleted automatically without the confirmation. Let's think that we have a MESG service that manages passwords. And this service is used by many people. But after a while we decided to shutdown this service by deleting it. Still, passwords database should be kept to give a copy of them to their owners let's say via email. In a case like this, deleting a service not always means that data should be deleted as well.

Of course instead of removing the service, devs can stop it and keep the data but data is very sensitive thing so we should avoid any margin to errors. Since this is a UX thing and there is no one right answer, I'll not push on to have this delete optionally but this is what I'm thinking.

Second reason why we need this option is: In active development while using aliases, we'll want to keep data volumes. In order to do that within _dev_ command we should have this option to not delete data volumes. And for the devs who doesn't use _dev_ and just use the start command, when they update their source code they'll naturally use the delete command  directly to stop and delete the service -instead of stop-. So it shouldn't automatically delete the volume because it'll be needed during development.

NicolasMahe: Good point

NicolasMahe:
Why not deleting the volume by default?



> How about _-delete-data-volumes_?

I like `--delete-data`. From a user point of view, he doesn't know about docker volume. He only cares about data.

commands/service_delete.go
```diff
@@ -32,6 +33,7 @@ mesg-core service delete --all`,
 
 	c.cmd.Flags().BoolVar(&c.all, "all", c.all, "Delete all services")
 	c.cmd.Flags().BoolVarP(&c.force, "force", "f", c.force, "Force delete all services")
+	c.cmd.Flags().BoolVarP(&c.deleteData, "delete-data", "d", c.force, "Delete services' persistent data along the way")
```
NicolasMahe:
```suggestion
	c.cmd.Flags().BoolVarP(&c.deleteData, "delete-data", "d", c.deleteData, "Delete services' persistent data along the way")
```

commands/service_delete.go
```diff
@@ -54,14 +56,27 @@ func (c *serviceDeleteCmd) preRunE(cmd *cobra.Command, args []string) error {
...
 	}
+
+	return c.askDeleteData()
+}
+
+func (c *serviceDeleteCmd) askDeleteData() error {
+	if !c.deleteData {
+		if err := survey.AskOne(&survey.Confirm{
+			Message: "Do you want to remove service(s)' persistent data as well?",
+			Default: true,
```
NicolasMahe:
```suggestion
			Default: false,
```

protobuf/coreapi/api.proto
```diff
@@ -297,7 +297,8 @@ message DeployServiceReply {
 // }
 // ```
 message DeleteServiceRequest {
-  string serviceID = 1; // The Service ID. Generated when using the [`DeployService` API](#deployservice).
+  string serviceID = 1;  // The Service ID. Generated when using the [`DeployService` API](#deployservice).
+  bool deleteData = 2; // When enabled, any persistent data that belongs to service and its dependencies will be also deleted.
```
NicolasMahe:
```suggestion
  bool deleteData = 2; // When enabled, any persistent data (volumes) that belongs to service and its dependencies will be also deleted.
```

ilgooz: How about _any volume_ to _any data volume_? Bu still, gRPC API is meant to be used by devs so they don't necessarily need to know about **volumes**. Keeping _persistent data_ might be  better.

NicolasMahe: Here, i will say a gRPC user is more luckily to also be a service developer. In service with speak about volumes.

```suggestion
  bool deleteData = 2; // When enabled, any persistent data (volumes) that belongs to service and its dependencies will be also deleted.
```

commands/service_delete.go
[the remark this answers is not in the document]
ilgooz: A scenario with _-y_:

```
$ mesg-core delete --all 
// > Are you sure to delete all services? (y/N)
```

```
$ mesg-core delete -y --all SID 
// < Deletes all services. No questions asked.
```

```
$ mesg-core delete --delete-data SID 
// > Are you sure to delete all data associated with the service? (y/N)
```

```
$ mesg-core delete -y --delete-data SID 
// < Deletes service and data volumes. No questions asked.
```

```
$ mesg-core delete --all --delete-data 
// > Are you sure to delete all services? (y/N)
// > Are you sure to delete all data associated with the services? (y/N)
```


```
$ mesg-core delete -y --all --delete-data 
// < Deletes services and data volumes. No questions asked.
```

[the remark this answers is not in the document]
ilgooz: So, with this pattern if a dev uses _-y_, this means dev answers every asked question yes. Which means dev should already know what questions will be prompted before. This way we can make sure that devs understand what they're really doing.

By name _-force_ does not imply anything, it's ambiguous. Also I don't agree on the following behavior:
```
NO --force and NO --delete-data, question to confirm volume deletion
```
If the default of _delete-data_ is false, we shouldn't be asking for deleting data when _--delete-data_ flag is not used.

And following are already have the same result:
```
--force and --delete-data, no confirmation, delete volumes
```
```
NO --force and --delete-data, delete volume without question
```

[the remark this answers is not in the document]
ilgooz: We decided to:
* Have _-y_ to answer all asked questions with yes. 
* Delete data volumes by default. Change _-delete-data_ with _-keep-data_. When keep data flag is used, no questions will be asked to delete services. Services will not be deleted.

api/delete.go
NicolasMahe: What about services that have the same alias? Like a update of a service.
If the user want to delete the previous version (with the service ID), in your case, it will delete the volume that are actually used by the latest version of the service.
We need to leave this option to the user.

antho1404: Otherwise we could consider that if it’s an I’d then we don’t delete, if it’s an alias we delete. The volume is related to an alias not an id so it makes sense. I’m just afraid that if even for us it looks complicated with the different flags it means it will be way too complicated for users.

ilgooz: I've pushed a new commit (implemented https://github.com/mesg-foundation/core/pull/613#discussion_r238206929). Please check and let's have a final decision.

ilgooz: * _-y_ answers every question as yes. With `mesg-core service -y --all`, all services and their data will be deleted with no question asked.

NicolasMahe:
Last version
```
mesg-core service SID
Are you sure you want to delete service SID? [y/N]
Do you want to delete the data of service SID? [y/N]
```
```
mesg-core service SID -y
No question. Delete service and its data
```
```
mesg-core service SID -keep-data
Are you sure you want to delete service SID? [y/N]
No question about data
Delete only service
```
```
mesg-core service SID -keep-data -y
No questions
Delete only service
```
```
mesg-core service -all
Are you sure you want to delete ALL services? [y/N]
Do you want to delete ALL the data of ALL services? [y/N]
Delete ALL services and ALL their data
```
```
mesg-core service -all -keep-data
Are you sure you want to delete ALL services? [y/N]
Delete only ALL services. KEEP their data
```
```
mesg-core service -all -keep-data -y
Delete only ALL services. KEEP their data
```
```
mesg-core service -all -y
Delete ALL services and ALL their data
```

commands/service_delete.go
```diff
@@ -10,17 +10,27 @@ import (
 	survey "gopkg.in/AlecAivazis/survey.v1"
 )
 
+var (
+	errConfirmationNeeded = errors.New("can't continue without confirmation")
+	errInputsNeededErr    = errors.New("at least one service id must be provided (or run with --all flag)")
```
krhubert: s/errInputsNeededErr/errNoID/

+ two times err (as prefix and suffix)

commands/service_delete.go
```diff
@@ -29,39 +39,83 @@ mesg-core service delete --all`,
 		PreRunE: c.preRunE,
 		RunE:    c.runE,
 	})
-
-	c.cmd.Flags().BoolVar(&c.all, "all", c.all, "Delete all services")
-	c.cmd.Flags().BoolVarP(&c.force, "force", "f", c.force, "Force delete all services")
+	c.cmd.Flags().BoolVarP(&c.yes, "yes", "y", c.yes, `Automatically say 'Yes' to all prompts and run non-interactively`)
+	c.cmd.Flags().BoolVar(&c.deleteAllServices, "all", c.deleteAllServices, "Delete all services")
```
krhubert: s/deleteAllServices/all at it was before, at it represents the flag not the action that is taken.

Same as you did with yes - you haven't name id, confirmAllQuestions, just simple yes

commands/service_delete.go
```diff
@@ -29,39 +39,83 @@ mesg-core service delete --all`,
...
+	c.cmd.Flags().BoolVar(&c.keepData, "keep-data", c.keepData, "Delete services' persistent data along the way")
 	return c
 }
 
 func (c *serviceDeleteCmd) preRunE(cmd *cobra.Command, args []string) error {
-	if len(args) == 0 && !c.all {
-		return errors.New("at least one service id must be provided (or run with --all flag)")
+	if err := c.validateInputs(args); err != nil {
```
krhubert: remove this method, why was created for one if?

commands/service_delete.go
```diff
@@ -29,39 +39,83 @@ mesg-core service delete --all`,
...
+		return errInputsNeededErr
+	}
+	return nil
+}
+
+// confirmServiceDelete prompts a confirmation dialog for deleting services.
+func (c *serviceDeleteCmd) confirmServiceDelete() error {
+	var (
+		confirmed bool
+		confirm   = &survey.Confirm{Default: false}
```
krhubert: Defautl: false is not needed as it's false when created.

commands/service_delete.go
```diff
@@ -29,39 +39,83 @@ mesg-core service delete --all`,
...
+		return errConfirmationNeeded
 	}
 	return nil
 }
 
+// confirmServiceDelete prompts a confirmation dialog for deleting services' data.
+func (c *serviceDeleteCmd) confirmDataDelete() error {
+	if err := c.survey.AskOne(&survey.Confirm{
+		Message: "Do you want to remove service(s)' persistent data as well?",
+		Default: false,
```
krhubert: Defautl: false is not needed as it's false when created.

commands/service_delete.go
```diff
@@ -29,39 +39,83 @@ mesg-core service delete --all`,
...
+	}, &c.keepData, nil); err != nil {
+		return err
+	}
+
+	c.keepData = !c.keepData
```
krhubert: why here is not?

commands/service_delete_test.go
```diff
@@ -3,18 +3,198 @@ package commands
...
+
+	me.On("ServiceDelete", false, serviceID).Once().Return(nil)
+
+	require.NoError(t, c.cmd.Execute())
+
+	sm.AssertExpectations(t)
+	me.AssertExpectations(t)
+}
+
+func TestServiceDeleteWithServiceIDAndYesFlag(t *testing.T) {
```
krhubert: Function name too long
 use table testing

commands/service_delete_test.go
```diff
@@ -3,18 +3,198 @@ package commands
...
+
+	me.On("ServiceDelete", true, serviceID).Once().Return(nil)
+
+	require.NoError(t, c.cmd.Execute())
+
+	sm.AssertExpectations(t)
+	me.AssertExpectations(t)
+}
+
+func TestServiceDeleteWithAllAndYesFlags(t *testing.T) {
```
krhubert: Function name too long

commands/commands.go
```diff
@@ -52,7 +53,7 @@ type Executor interface {
 
 // Build constructs root command and returns it.
 func Build(e Executor) *cobra.Command {
-	return newRootCmd(e, newSurvey()).cmd
```
ilgooz: we should have tests for survey. we already had some tests, why removing them within this commit?

commands/service_delete_test.go
```diff
@@ -3,198 +3,65 @@ package commands
...
-	sm.AssertExpectations(t)
-	me.AssertExpectations(t)
+func TestServiceDeletePreRunE(t *testing.T) {
+	c := newServiceDeleteCmd(nil)
+	c.discardOutput()
+	require.Equal(t, errNoID, c.preRunE(c.cmd, nil))
 }
 
-func TestServiceDeleteWithAllAndYesFlags(t *testing.T) {
+func TestServiceDeleteRunE(t *testing.T) {
```
ilgooz: This test doesn't cover the tests deleted within this commit. I don't understand why they're removed.

commands/survey.go
```diff
@@ -1,21 +0,0 @@
-package commands
-
-import survey "gopkg.in/AlecAivazis/survey.v1"
-
-// Survey is a command line survey.
-type Survey interface {
```
ilgooz: we should have tests for survey. we already had some tests, why removing them within this commit?

commands/service_delete.go
```diff
@@ -29,39 +35,74 @@ mesg-core service delete --all`,
 		PreRunE: c.preRunE,
 		RunE:    c.runE,
 	})
-
+	c.cmd.Flags().BoolVarP(&c.yes, "yes", "y", c.yes, `Automatic "yes" to all prompts and run non-interactively`)
 	c.cmd.Flags().BoolVar(&c.all, "all", c.all, "Delete all services")
-	c.cmd.Flags().BoolVarP(&c.force, "force", "f", c.force, "Force delete all services")
+	c.cmd.Flags().BoolVar(&c.keepData, "keep-data", c.keepData, "Do not delete services' persistent data")
```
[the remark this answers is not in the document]
krhubert: Yes and I've changed the message from `Delete services' persistent data along the way` this one, as it indicates if you set this flag then data will be deleted.

commands/service_delete_test.go
krhubert: It tests only for runE, the previous tests were for both preRunE and runE at the same time

commands/survey.go
krhubert: Because it was too complicated,

Again there was a new interface, new mock, mocks in the tests just to test `AskOne` method. We need to find a better way to test it. If we will mock everything that we can't tests we will end up in mock hell.

Also, those tests mix testing of validation (preRunE) and execution (runE) and good unit tests should test only one behavior/func at once (test one thing at a time).

NicolasMahe: On [survey's README](https://github.com/AlecAivazis/survey#testing), the guy suggest to use (go-expect)[https://github.com/Netflix/go-expect] to catch stuff in stdout

commands/service_dev.go
```diff
@@ -79,7 +79,7 @@ func (c *serviceDevCmd) runE(cmd *cobra.Command, args []string) error {
 	defer func() {
 		var err error
 		pretty.Progress("Removing the service...", func() {
-			err = c.e.ServiceDelete(id)
+			err = c.e.ServiceDelete(false, id)
```
NicolasMahe: What about adding the flag `--keep-data` to command dev to prevent the deletion of volume, and the default behavior will be to delete the volume?
Anyway, let's NOT do this change in this PR.

service/volume_test.go
```diff
@@ -0,0 +1,47 @@
+package service
```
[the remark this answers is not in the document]
NicolasMahe: Should be done in another PR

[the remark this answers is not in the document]
NicolasMahe: I add the test but it show a bug with `VolumesFrom`. It try to delete the volume 2 times.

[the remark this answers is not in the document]
NicolasMahe: Fix in 2718c05bd6c9bf7a47cd1f393f25e95c1d513c9e.
I separate `dep.extractVolumes` and `dep.extractVolumesFrom` in 2 functions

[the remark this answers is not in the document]
NicolasMahe: The current test env of circleci does not support volume
https://support.circleci.com/hc/en-us/articles/360007324514-How-can-I-mount-volumes-to-docker-containers-

we should try to run test with `runOnMachine`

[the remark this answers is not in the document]
NicolasMahe: Let's do it in another PR

commands/survey.go
ilgooz: I don't approve removing the tests for survey. Having one more mock doesn't lead to a mock hell. I had a look on suvery's testing utilities yesterday but it seemed complicated to me, mocks are way simple. If you want to delete the tests anyway please put their replacement otherwise we will lose some meaningful tests.

commands/service_delete.go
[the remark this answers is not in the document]
ilgooz: Correct, for a second I thought this change is about the confirmation dialog.

service/volume_test.go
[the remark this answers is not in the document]
NicolasMahe: https://github.com/mesg-foundation/core/issues/621

commands/survey.go
NicolasMahe: https://github.com/mesg-foundation/core/issues/622

krhubert: My motiviation to delete those tests was:
- mock hell what I mean by that if we swtich to passing interface to commands we will need: `daemon.Deamon`, `api.API`, `service.Service` (maybe) `CoreClient` mocks so it's already a lot. Adding one more mock seems good from short perspective, but in long run it's not so obious.

- so choosing mocking should be the last thing while testing. We alwasy need to try to find other solutions/design patterns that will free code of mocks. Why I think so. Look at go source code:
There is only one mock `net/dnsclient_unix_test.go:func` here. 

```
➜  moby git:(master) gogrep -i mock |grep struct |wc -l
36
```
So docker has 36 mocks and we probably haven't reached 2% of their codebase and we already have:
```
➜  core git:(feature/auto-remove-volume) ✗ gogrep -i mock |grep struct |wc -l
13
```

13 mocks.

- while mocking seems easy to go solution it is not in long run (compared to js where it's standard and it's super easy, but js are more flexible as it is dynamic programming language)

- survey suggested go-expect and vx10 for testing but vx10 is suspicious to me, but it's still better to use it as it doesn't require to introduce changes in the source code. 

- about using https://github.com/AlecAivazis/survey/blob/v1.7.0/confirm_test.go#L21 I was wrong. It can't be used.

So to summarize we should think of a better solution if possible, sleep with it, look from a different perspective, but introduce mocks because it easy should be avoided in general.

In was the same with `commands` package before `caputereStd` was created, there were no tests at all and in the meantime, we introduced them because we find a way to do so.

ilgooz: Again, this is something that I just don't agree. There is no logical reason to remove a test if it's not replaced by an equivalent one. I'm not even talking about using the mocks here, it's not relevant.

Survey pkg doesn't provide a nice way of testing as I already checked so, it's a proper option to use mocks here. I don't mind adopting a **better way**. Could you please provide one? It seems that we're seeing things in different perspectives.  

> mock hell what I mean by that if we swtich to passing interface to commands we will need: daemon.Deamon, api.API, service.Service (maybe) CoreClient mocks so it's already a lot. Adding one more mock seems good from short perspective, but in long run it's not so obvious.

This is totally an another topic. If we want to reduce possibility for programming errors we should have tests about how _daemon_ and _coreapi_ used in the _commands/*_ pkgs. And I'd like doing that with unit tests. And I remember that you said it's not necessary to test *wrapper* funcs (the default implementation between these pkgs). And we don't have tests for them. I also [suggested](https://forum.mesg.com/t/automatize-manual-testing-for-core-true-integration-testing) to have at least some integration tests for reducing manual tests that we always required to do. Otherwise we don't have a way of catching bugs inside wrappers, the place where _commands_ pkg does IO which is very important.

I too prefer using the native testing utilities that can be provided by the pkg itself. But testing how _coreapi_ is used in commands pkg is easier by mocking the client side funcs of _coreapi_ as I experimented. We had the same problem before with the _go-application_ and _go-service_. We didn't use mocks there and implemented a fake _net.Conn_, which is a bit harder to maintain when compared to mocking because starting a fake gRPC server causes some async behavior that needed be synced otherwise it introduces bug into tests.

For daemon, we already have its interface so the best option is is to create mocks on top of that.

krhubert: > Survey pkg doesn't provide a nice way of testing as I already checked so, it's a proper option to use mocks here.

This is the point that it isn't good to use mocks. Again Survey pkg option for testing is still better then modifying code in order to introduce the Survey interface. Maybe there is an option to write Survery bridge. I don't know.

>  I don't mind adopting a better way. Could you please provide one?

I don't mind either, but as I said there is have no one at this moment, but shouldn't be the root cause of introducing mocks, because maybe there is another way that needs to be discovered.

For the rest of things is a different discussion, but in general, having too many tests as bad as having too little. And about the other things we can talk in different thread/PR

PR #614, opened 2018-12-03
dev-core: rm hard deleting MESG services & containers

ilgooz:


krhubert:
Why this way? There was pbm with stoping the container in the past (and I think the pbm still exist)

ilgooz:
> Why this way? There was pbm with stoping the container in the past (and I think the pbm still exist)

I removed them for this reason: https://github.com/mesg-foundation/core/issues/536#issuecomment-443606899

If we see any errors while stopping services again, let's determine the actual problem and open an issue about it so we can solve it.

Issue #618, opened 2018-12-04
Issue on service deletion with alias


NicolasMahe:
Here is a scenario that doesn't make sense:

```
➜  core git:(dev) ./dev-cli service deploy /Users/nico/Developments/MESG/services/service-bitcoin
✔ Service context received with success
✔ Image built with success
✔ Service deployed with ID: d95a83b25b51be82c38ee793ed15614d08b5b51d
To start it, run the command:
        mesg-core service start d95a83b25b51be82c38ee793ed15614d08b5b51d
```
```
➜  core git:(dev) ./dev-cli service list
ID                                          ALIAS                                        NAME       STATUS
d95a83b25b51be82c38ee793ed15614d08b5b51d    bitcoin                                      Bitcoin    stopped
```

I change something on the bitcoin service, then deploy again.
```
➜  core git:(dev) ./dev-cli service deploy /Users/nico/Developments/MESG/services/service-bitcoin
✔ Service context received with success
✔ Image built with success
✔ Service deployed with ID: 06e0d1c14c142b58de2b1ded3bad63bd4fe02948
To start it, run the command:
        mesg-core service start 06e0d1c14c142b58de2b1ded3bad63bd4fe02948
```
```
➜  core git:(dev) ./dev-cli service list
ID                                          ALIAS                                        NAME       STATUS
d95a83b25b51be82c38ee793ed15614d08b5b51d    bitcoin                                      Bitcoin    stopped
06e0d1c14c142b58de2b1ded3bad63bd4fe02948    bitcoin                                      Bitcoin    stopped
```
```
➜  core git:(dev) ./dev-cli service delete bitcoin
✔ service "bitcoin" deleted
```
```
➜  core git:(dev) ./dev-cli service list
ID                                          ALIAS                                        NAME       STATUS
d95a83b25b51be82c38ee793ed15614d08b5b51d    bitcoin                                      Bitcoin    stopped
```
```
➜  core git:(dev) ./dev-cli service delete bitcoin
⨯ can't delete service "bitcoin": rpc error: code = Unknown desc = database: service bitcoin not found
there was a problem with deleting some services
exit status 1
```

----

Alias should be unique.
Few ideas on how to solve it:
- 1
When deploying a new service, we should generate a new alias for the service that have the same alias. The new problem is the old service (now with a different alias) cannot access its volumes (it will have new volumes).
- 2
When deploying a service where its alias already exist in the db, we should remove the old service (and of course ask user to confirm).

[...]

PR #619, opened 2018-12-04
make service alias unique

ilgooz:
a part of https://github.com/mesg-foundation/core/issues/618


database/service_db.go
```diff
@@ -169,16 +169,36 @@ func (d *LevelDBServiceDB) Save(s *service.Service) error {
...
+	tx, err := d.db.OpenTransaction()
+	if err != nil {
+		return err
+	}
+
+	// check if there is a service with the same alias.
+	if _, err := d.get(tx, s.Alias); err == nil {
+		tx.Discard()
+		return &ErrSameAlias{s.Alias}
+	} else if _, ok := err.(*ErrNotFound); !ok {
```
NicolasMahe:
```suggestion
	} else if !IsErrNotFound(err) {
```

database/service_db_test.go
```diff
@@ -48,6 +48,21 @@ func TestServiceDBSave(t *testing.T) {
...
+func TestServiceDBSaveWithSameAlias(t *testing.T) {
+	db, closer := openServiceDB(t)
+	defer closer()
+
+	s := &service.Service{ID: "01", Alias: "a", Name: "test-service"}
+	require.NoError(t, db.Save(s))
+	s = &service.Service{ID: "02", Alias: "a", Name: "test-service"}
+	require.Equal(t, &ErrSameAlias{"a"}, db.Save(s))
+
+	_, err := db.Get("1")
```
krhubert: Why to get `"1"` and `"2"` (as they are unrelated to this test), probably you wanted `"01"` and `"02"`

ilgooz:
I'm freezing this PR for now. [Discussion](https://forum.mesg.com/t/brainstorming-services-alias-sid-versions-and-data-volumes/129/6) needs to be finalized first.

NicolasMahe:
Anthony suggest to put the logic if the service alias already exist in the `api.deploy` function.
In this function, we have access to the service def, and the database. We also have a 2 way stream, so we can send the replace confirmation to the cli and wait for confirmation.
The deploy should not rely on database error but should call the database.get function to check if the alias already exist

NicolasMahe:
This PR is only about adding a new behavior to the database package.
When saving a service with an alias that already exist in the db, existing service with same alias get deleted.
Result: only 1 service per alias can exist in the db

database/service_db_test.go
```diff
@@ -30,13 +30,28 @@ func TestServiceDBSave(t *testing.T) {
...
+	s2 := &service.Service{ID: "01", Alias: "1", Name: "test-service"}
+	require.NoError(t, db.Save(s2))
+	_, err := db.Get(s1.ID)
+	require.IsType(t, &ErrNotFound{}, err)
+
+	// different id, different alias. should not replace anything
+	s3 := &service.Service{ID: "02", Alias: "2", Name: "test-service"}
+	require.NoError(t, db.Save(s3))
+	ss, _ = db.All()
+	require.Len(t, ss, 2)
```
ilgooz: it'd be better also to compare received services here with Equal so we can make sure that we really overwrite the service with new one(s2) correctly.

ilgooz:
LGTM

PR #623, opened 2018-12-06
Ask user's confirmation when service will replace an other service

NicolasMahe:


ilgooz:
- [ ] fix `./mesg-tools ./scripts/build.proto`
- [ ] https://github.com/mesg-foundation/core/pull/623/files#diff-cc307140c73bc10debbf2d78fb4eb4b3R17

interface/grpc/core/deploy.go
```diff
@@ -29,15 +34,53 @@ func (s *Server) DeployService(stream coreapi.Core_DeployServiceServer) error {
...
+
+	deployOptions := []api.DeployServiceOption{
+		api.DeployServiceStatusOption(statuses),
+	}
+
+	// receive confirmation for a force deploy.
+	// if it's not provided deployment will start anyway.
+	if sr.Confirmation != nil {
+		confirmation := sr.Confirmation.GetValue()
+		deployOptions = append(deployOptions, api.DeployServiceConfirmationOption(func(alias string) bool {
```
ilgooz: merge this line with 54

Gopkg.toml
```diff
@@ -24,12 +24,15 @@
...
 required = [
+  # dependencies of ./scripts.
   "github.com/golang/protobuf/protoc-gen-go",
   "github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc",
   "github.com/go-bindata/go-bindata/go-bindata",
-  "github.com/vektra/mockery/cmd/mockery"
+  "github.com/vektra/mockery/cmd/mockery",
+
+  # used by protobuf pkg.
+  "github.com/golang/protobuf/ptypes/wrappers",
```
krhubert: This should be handled by dep itself, as it's direct dependency in source code

commands/service_dev.go
```diff
@@ -62,7 +63,17 @@ func (c *serviceDevCmd) runE(cmd *cobra.Command, args []string) error {
 		printDeployStatuses(statuses)
 	}()
 
-	id, validationError, err := c.e.ServiceDeploy(c.path, statuses)
+	id, validationError, err := c.e.ServiceDeploy(c.path, nil, func(alias string) (bool, error) {
+		pretty.DestroySpinner()
+		var confirm bool
+		if err := survey.AskOne(&survey.Confirm{
+			Message: fmt.Sprintf("A service already exist with the same alias %q. Do you confirm to replace it?", alias),
```
krhubert: `A service already exist with the same alias %q. Do you confirm to replace it?`
to
`A service with the same alias %q already exists. Do you want to replace it?`

commands/service_deploy.go
```diff
@@ -32,6 +35,7 @@ To get more information, see the [deploy page from the documentation](https://do
 		RunE:    c.runE,
 		Args:    cobra.MaximumNArgs(1),
 	})
+	c.cmd.Flags().BoolVarP(&c.force, "force", "f", c.force, "Force deploy the service by overwriting existing service with the same alias if there is")
```
krhubert: In one command we have `yes` , here we have `force` so it's starting to mix,

maybe change to overwrite?

Also,

`Force deploy the service by overwriting existing service with the same alias if there is` to
`Deploying overwrites existing service with the same alias`

One more thing related to `all` (it's not about this PR but in general) in service delete we have:
`	c.cmd.Flags().BoolVarP(&c.yes, "yes", "y", c.yes, `Automatic "yes" to all prompts and run non-interactively`)`
But there is no information for the user what question will be asked during the process and it isn't possible to answer to one question yes and others no by using the flags.

interface/grpc/core/deploy.go
```diff
@@ -1,8 +1,11 @@
 package core
 
 import (
+	"io"
 	"sync"
 
+	"github.com/golang/protobuf/ptypes/wrappers"
+
```
krhubert: empty line

interface/grpc/core/deploy.go
```diff
@@ -29,15 +34,53 @@ func (s *Server) DeployService(stream coreapi.Core_DeployServiceServer) error {
...
+	}
+
+	if sr.Confirmation != nil {
+		if err := sr.RecvMessage(); err != nil {
+			return err
+		}
+	}
+
+	if sr.URL != "" {
+		service, validationError, err = s.api.DeployServiceFromURL(sr.URL, deployOptions...)
```
krhubert: This functions is hard to read for other developers, as its not clear what's going on here

interface/grpc/core/deploy.go
```diff
@@ -29,15 +34,53 @@ func (s *Server) DeployService(stream coreapi.Core_DeployServiceServer) error {
...
 		return err
 	}
-	if url != "" {
-		service, validationError, err = s.api.DeployServiceFromURL(url, api.DeployServiceStatusOption(statuses))
+
+	deployOptions := []api.DeployServiceOption{
+		api.DeployServiceStatusOption(statuses),
+	}
+
+	// receive confirmation for a force deploy.
```
krhubert: no tests for new behavior.

commands/service_dev.go
ilgooz: done

commands/service_deploy.go
ilgooz: * 'Force' makes a good sense to me, it's also widely used. It has a meaning about 'force doing' things that aren't meant to be doable when it's not used. I'm not against to overwrite as well but  force is a bit better to me and we can use it in other commands so it's more convenient. 

* 'yes' and 'force' has totally meanings and results.

>But there is no information for the user what question will be asked during the process and it isn't possible to answer to one question yes and others no by using the flags.

* This is why 'yes' command is good. Devs shouldn't use it if they didn't experienced the questions before. So have to use the command without 'yes' and once they get how it's work they can switch to 'yes' to skip interactive.

interface/grpc/core/deploy.go
ilgooz: done. isn't there a lint rule for this?

interface/grpc/core/deploy.go
ilgooz: what do you mean?

interface/grpc/core/deploy.go
ilgooz: we don't have a way of faking _api_ pkg yet. feel free to push some integration tests.

interface/grpc/core/deploy.go
krhubert: Unfortunately not :(

interface/grpc/core/deploy.go
krhubert: But I'm not taking about api, I'm talking about grpc only - L47-L70 - you don't need api to test this

interface/grpc/core/deploy.go
krhubert: I don't know how exactly explain `it's hard to read`. When you first looking on the function (and you're not an author of it) you don't have an idea what's going on inside without further reading.

ilgooz: I don't feel the same. Feel free to put some comments if you wish but I think var/func naming is very good here so it self-explains.

interface/grpc/core/deploy.go
ilgooz: Could you please provide an example, I wonder how it can be tested without faking _api_ calls.

ilgooz: I don't think that would be a proper way of testing the DeployService. Feel free to create an another PR based on this one to implement the tests. So, we can compare and merge it to this PR if it's convenient enough.

interface/grpc/core/deploy.go
ilgooz: current implementation better for readability

ilgooz:
I've tried to solve the first bullet here https://github.com/mesg-foundation/core/pull/623#issuecomment-444884744 but could fix yet. Please create an another issue about it before merging this PR.

NicolasMahe:
@ilgooz please pause development on this PR because of the following conversation:
https://forum.mesg.com/t/improve-relation-between-sid-and-hashes/141

ilgooz:
There is no need for this PR anymore because of the forum topic posted by @NicolasMahe.
Improvements featured in this PR moved to an another one: https://github.com/mesg-foundation/core/pull/642

Issue #625, opened 2018-12-06
Rename service ID and Alias


NicolasMahe:
Rename:
Id -> hash
Alias -> SID

NicolasMahe:
Done

PR #630, opened 2018-12-07
Rename service id to hash

NicolasMahe:
Second half of https://github.com/mesg-foundation/core/issues/625

Not compatible with https://github.com/mesg-foundation/core/pull/631


database/service_db.go
```diff
@@ -136,39 +136,39 @@ func (d *LevelDBServiceDB) Get(idOrSID string) (*service.Service, error) {
 }
 
 // get retrives service from database by using r reader.
-func (d *LevelDBServiceDB) get(r leveldb.Reader, idOrSID string) (*service.Service, error) {
-	id := idOrSID
+func (d *LevelDBServiceDB) get(r leveldb.Reader, hashOrSID string) (*service.Service, error) {
+	id := hashOrSID
```
ilgooz: id to hash?

database/service_db.go
```diff
@@ -136,39 +136,39 @@ func (d *LevelDBServiceDB) Get(idOrSID string) (*service.Service, error) {
...
 
 	// check if key is a sid, if yes then get id.
```
ilgooz:
```suggestion
	// check if key is a sid, if yes then get hash.
```

ilgooz:
```suggestion
	// check if key is a sid, if yes then get hash.
```

PR #642, opened 2018-12-13
interface/grpc/core>deploy: minor improvements

ilgooz:


ilgooz:
needs review

krhubert:
This PR makes conflict with https://github.com/mesg-foundation/core/pull/641 and #641 has greater priority than this. Also, I've already taken some comments and pushed into #641. So let's focus on features first.

antho1404:
@krhubert maybe it would make more sense to do the changes on this PR and we merge this one when it's about the stream for the deploy.
Let's try to keep PR as small as possible and #641 should be only about the feature of #641

NicolasMahe:
The implementation of the reader from the gRPC stream done by @krhubert on #641 is actually simpler and doesn't involve any other parameter (eg: URL).
The good part is that the reader is only reading chunk from the stream, so if we add more parameter to the stream, the reader doesn't need to change.

https://github.com/mesg-foundation/core/blob/c8775575b921c22b09030d7f4d2c3495391dd522/interface/grpc/core/deploy.go#L97-L119x

I think this PR is not necessary anymore.

ilgooz:
> The implementation of the reader from the gRPC stream done by @krhubert on #641 is actually simpler and doesn't involve any other parameter (eg: URL).
> The good part is that the reader is only reading chunk from the stream, so if we add more parameter to the stream, the reader doesn't need to change.
> 
> https://github.com/mesg-foundation/core/blob/c8775575b921c22b09030d7f4d2c3495391dd522/interface/grpc/core/deploy.go#L97-L119x
> 
> I think this PR is not necessary anymore.

I don't agree with it because putting a simple _RecvMessage()_ func to stream reader and re-using it in the _io.Reader_ implementation and in the deploy gRPC handler(_DeployService_) was my first intention while creating the stream reader and to me it's way easier to use. You can check this [link](https://github.com/mesg-foundation/core/blob/53d6586877fb26a5532d459c697da81e58b980f7/interface/grpc/core/deploy.go) to compare.  Feel free to close this PR if you don't agree.

ilgooz:
this PR is now a part of another one https://github.com/mesg-foundation/core/pull/663

PR #645, opened 2018-12-15
MESG token integration

antho1404:


config/key.go
```diff
@@ -0,0 +1,45 @@
...
+	yaml "gopkg.in/yaml.v2"
+)
+
+// Key is the structure that stores all informations related to the key system
+type Key struct {
+	Address string
+}
+
+// GetKey returns the core's key
+func (c *Config) GetKey() (*Key, error) {
```
krhubert: This one could be loaded when whole config is load (at once)

interface/grpc/core/create_workflow.go
```diff
@@ -8,6 +8,9 @@ import (
 
 // CreateWorkflow creates and runs a new workflow.
 func (s *Server) CreateWorkflow(ctx context.Context, request *coreapi.CreateWorkflowRequest) (*coreapi.CreateWorkflowReply, error) {
+	if err := s.requireStake(); err != nil {
```
krhubert: for requireStake grpc-middelware would be great

interface/grpc/core/require_stake.go
```diff
@@ -0,0 +1,16 @@
+package core
+
+import "github.com/mesg-foundation/core/config"
+
+// ListenEvent listens events matches with eventFilter on serviceID.
+func (s *Server) requireStake() error {
+	c, err := config.Global()
```
krhubert: Do not use config.Global, pass it as argument

interface/grpc/core/require_stake.go
```diff
@@ -0,0 +1,16 @@
...
+	if err != nil {
+		return err
+	}
+	k, err := c.GetKey()
```
krhubert: Key will be loaded from disk every time someone call api.

NicolasMahe:
Let's keep it open for now. I will close it when i don't need it.

NicolasMahe:
Closing this PR as it's not relevant anymore

PR #646, opened 2018-12-16
parameters> support nested and repeated types

### LINKED ISSUE #564: service: parameterValidator should support all types

ilgooz:
This PR has no breaking changes but `Object` type should be deprecated in future. Please perform your own manual tests to report any possible breaking changes.

You can use [service-articles ](https://github.com/ilgooz/service-articles/blob/master/mesg.yml) to manually test this PR.
- [x] Introduce `Any` type _(Object will be deprecated in future, Object acts like Any for now.)_
- [x] Support arrays(repeated types) in parameters. 
- [x] Support nested parameters.
- [x] Fixes #564.
- [ ] See if we can implement yml Unmarshaler for Type field in the definition.

* Update output of `service detail` command to support printing nested parameters. will be solved by https://github.com/mesg-foundation/core/issues/647.


protobuf/coreapi/api.proto
```diff
@@ -514,11 +514,13 @@ message Output {
...
-  string key = 8;         // Parameter's key.
-  string name = 1;        // Parameter's name.
-  string description = 2; // Parameter's description.
-  string type = 3;        // Parameter's type: `String`, `Number`, `Boolean` or `Object`.
-  bool optional = 4;      // Set the parameter as optional.
+  string key = 8;                     // Parameter's key.
+  string name = 1;                    // Parameter's name.
+  string description = 2;             // Parameter's description.
+  string type = 3;                    // Parameter's type: `String`, `Number`, `Boolean` or `Any`. It's filled when there are no child parameters.
+  repeated Parameter parameters = 9;  // Parameter's child parameters for nested data.
```
NicolasMahe: `type` and `parameters` cannot be set a the same time, no?
So here a `oneOf` will fit perfectly.

service/importer/assets/schema.json
```diff
@@ -90,14 +90,26 @@
...
-                  ],
-                  "default": "String"
+                  "oneOf": [
+                    {
+                      "type": "string",
+                      "enum": [
+                        "String",
+                        "Number",
+                        "Boolean",
+                        "Object",
```
NicolasMahe:
```suggestion
```

I think Object should be removed.

service/validation.go
```diff
@@ -1,97 +1,117 @@
...
+		case "Number":
+			if v.Type != vtree.Number {
+				p.addWarning(param, "not a number")
+			}
+		case "Boolean":
+			if v.Type != vtree.Bool {
+				p.addWarning(param, "not a boolean")
+			}
+		// TODO(ilgooz) Deprecate Object in future.
+		case "Any", "Object":
```
NicolasMahe: I suggest to make a breaking change and directly force to remove Object from service definition.

NicolasMahe:
@ilgooz I wonder if we could find a simpler way to parse the yaml into the go struct without using vtree or mapstructure.

Maybe we could dig into the yaml package to implement our functionalities without re-implementing something similar ourself.

service/validation.go
```diff
@@ -1,97 +1,117 @@
...
-func validateParametersSchema(parameters []*Parameter, data map[string]interface{}) []*ParameterWarning {
-	warnings := make([]*ParameterWarning, 0, len(parameters))
-
-	for _, param := range parameters {
-		warning := newParameterValidator(param).Validate(data[param.Key])
-		if warning != nil {
-			warnings = append(warnings, warning)
-		}
+func (p *parameterValidator) Validate(params []*Parameter, data interface{}) ([]*ParameterWarning, error) {
+	v := vtree.Analyze(data)
```
NicolasMahe: Doing validation by using reflection is not efficient.
Reflection should not be use in high demanded function.
Here, for every event, result, task that the code receive, it will do reflection.

The goal is to do the reflection once to populate a structure that can do the validation efficiently.

service/importer/assets/schema.json
ilgooz: I'll be a breaking change for existing service, are you sure?

service/validation.go
ilgooz: The reason why we use reflection is to support struct types and not just maps. We can always change the implementation of _vtree_ pkg to do type assertions instead of using reflection. But that way we're no longer be able to support structs.

_About why we need to support structs:_

In system services it's easier to pass existent struct types as input data. In the [example](https://github.com/mesg-foundation/core/blob/dev/systemservices/workflow/workflow.go#L20) below we don't have this flexibility.

We do a small hack here by using json marshal & unmarshal to convert struct type to a map in order to satisfy old implementation of the parameter validation func because it can only accept map types. Since json pkg heaviliy depends on reflection pkg to analyze given data type and set values it got from json string to given type this operation is very heavy. Wherareas _vtree_ pkg only analyzes the data type, it doesn't do further reflection like json.Unmarshal does to set values and we only call Analyze method once. So this can be considered as a lightweight operation and a small giveaway in favor of supporting structs.
```go

func CreateInputs(definition workflowparser.WorkflowDefinition, name string) (map[string]interface{}, error) {
	// TODO: this hack is not something that we should do but
	// it's needed because *parameterValidator is not able to identify
	// structs for now.
	definitionData, err := json.Marshal(definition)
	if err != nil {
		return nil, err
	}
	var data interface{}
	if err := json.Unmarshal(definitionData, &data); err != nil {
		return nil, err
	}
	return map[string]interface{}{
		"definition": data,
		"name":       name,
	}, nil
}
``` 

So with the power of _vtree_ pkg this code becomes to (should x3-x4 faster because less reflection made):

```go
func CreateInputs(definition workflowparser.WorkflowDefinition, name string) (map[string]interface{}, error) {
    return map[string]interface{}{
        "definition": definition,
        "name":       name,
    }, nil
}
```

We can also change the implementation of _vtree_ to drop struct support and reflection and only do type assertion but in that case we need to manually convert struct types to maps and this will cause to long, hart to read code and specially if the struct consist of more nested structs. It can quickly become a nightmare and it's hard to maintain.

ilgooz: One other option is to change the implementation of [ExecuteTask](https://github.com/mesg-foundation/core/blob/2cd3fa0/api/execute.go#L31) to accept json encoded byte data instead of a map type. This way we still going to be able to use all kind of Go types including structs in the system services but we need to encode them to json before passing to the _ExecuteTask_.

Then _ExecuteTask_ can unmarshal it to a map and we can to type assertion on top of that inside the _vtree_ pkg.

Pros:
* No need for doing reflection ourselves within the _vtree_.

Cons:
* We limit us using json as the encoding type by requiring to pass byte representation of it to the _ExecuteTask_ func. Maybe not a big cons.

* For system services we'll need to marshal to json and unmarshal it again inside the _ExecuteTask_ for no real reason which lowers the performance x2 time where we could just use the data type as it is and stick with _vtree_ where it'll do a small amount of reflection.

ilgooz: here is a sample [PR](https://github.com/mesg-foundation/core/pull/657) when assertion is adopted within _vtree_.

antho1404:
This PR is really big, I think it could be good to separate it in multiple PR based on the TODO list of this PR:
- Support Any
- Support Array
- Support nested
etc...

ilgooz:
> This PR is really big, I think it could be good to separate it in multiple PR based on the TODO list of this PR:
> 
> * Support Any
> * Support Array
> * Support nested
>   etc...

I agree that it's a bit a long PR but mostly due to some tests and I updated the vendor folder. We could have create smaller PRs but I wanted to implement the features all in once because they're strongly related with each other and I had to create a lib that needs to make recurse calls. Splitting this PR will not make a big change at this moment.

antho1404:
should we close this PR ?

ilgooz:
> should we close this PR ?

We can close this one and the other PR that related to this. In future, if we want to make some refactoring & allow more types for parameter validations we can turn back and use some code from here.

Issue #647, opened 2018-12-16
commands>service>detail: support printing nested parameters


ilgooz:
broken by: https://github.com/mesg-foundation/core/pull/646

ilgooz:
solved by https://github.com/mesg-foundation/core/pull/648

PR #648, opened 2018-12-17
service>detail: print whole service definition

ilgooz:
closes https://github.com/mesg-foundation/core/issues/647


NicolasMahe:
Done here: https://github.com/mesg-foundation/core/pull/672

PR #649, opened 2018-12-17
protobuf/coreapi: lowercase sid in JSON encoding

ilgooz:


antho1404:
Is there any good reason for that change ?

ilgooz:
> Is there any good reason for that change ?

Yes, if we don't lowercase the _SID_ in json encoding, it'll stay as uppercase where all the other fields are lowercased which is not convenient. Also there is no reason to keep _SID_ uppercase in json format, we may also have it as _sID_ since it consists of two words but it kinda doesn't look good.

antho1404:
Hum I guess you are talking about the display in the detail command. That's the only place I see we use json. In that case I might be ok but I don't want to introduce a special flag.
If it's for consistency which is great it has to be consistent on the api level and this implies some other modifications and so maybe a lot of modifications for a small problem

NicolasMahe:
@mesg-foundation/core
We need to have both lowercase and first letter uppercase style.
For now, both are SID and that's weird.

I suggest to go with "sid" everywhere and "Sid" in go struct when sid need to be public.

antho1404:
I'm ok to have everything lowercase but it's weird to have in the api SID with json name sid.
I would simply do:
```
string sid = 12
```

PR #652, opened 2018-12-19
Add db transaction mechanism to database execution

NicolasMahe:
I added 3 new methods to `ExecutionDB`:
- `OpenTransaction() (*leveldb.Transaction, error)`,
- `FindWithTx(tx leveldb.Reader, executionID string) (*execution.Execution, error)`
- `SaveWithTx(tx *leveldb.Transaction, execution *execution.Execution) error`

Do you think we should merge them with the classic `Find` and `Close`?


krhubert:
> We're locking whole database for a single execution. This is not efficient. We can use a locking mechanism to only lock by execution id. Locking whole database for a single execution has no benefit and slows done other executions happening at the same time specially when json.Unmarshal(reflection) is used.

You will have to lock it anyway because this is how transaction works. [docs](https://godoc.org/github.com/syndtr/goleveldb/leveldb#DB.OpenTransaction)
"Only one transaction can be opened at a time."
There is no need to lock per execution id, because you won't open >=2 transaction anyway.

What you can do to reduce locking time:

1) `json.Unmarshal` before open transaction
2) check exec status right after find (before running service.FromService and exec.Complete)

ilgooz:
> > We're locking whole database for a single execution. This is not efficient. We can use a locking mechanism to only lock by execution id. Locking whole database for a single execution has no benefit and slows done other executions happening at the same time specially when json.Unmarshal(reflection) is used.
> 
> You will have to lock it anyway because this is how transaction works. [docs](https://godoc.org/github.com/syndtr/goleveldb/leveldb#DB.OpenTransaction)
> "Only one transaction can be opened at a time."
> There is no need to lock per execution id, because you won't open >=2 transaction anyway.
> 
> What you can do to reduce locking time:
> 
> 1. `json.Unmarshal` before open transaction
> 2. check exec status right after find (before running service.FromService and exec.Complete)

There is a really high possibility of opening multiple transactions because multiple task results can be submitted through the api at the same time. Infact, 1000s or more of them might be opened which will slow down the whole process without any reason. This is why using leveldb transactions does not provide an efficient locking mechanism for this case.

krhubert:
> I agree with @ilgooz, the goal is to have 1000s of different execution (different execution ID) at the same time. So locking per execution ID is very efficient and allow the db to be used by other go routine.

But there is no point for locking because you have one lock on database. Even if two execution could be run in the same time, it won't because you won't open 2 transaction. So it would be like this.

10 diffrent exeuctions id lock mutex and then they still run one by one because they have to wait for lock from database. Locking per execution id is pointless...

NicolasMahe:
The lock on the database should be replace by the lock on execution ID. Two locks doesn’t make sense.

We could implement an execution manager that take care of loading and saving execution and locking it based on the execution ID.

krhubert:
> The lock on the database should be replace by the lock on execution ID. Two locks doesn’t make sense.

Yes, but you can't remove lock from db.

ilgooz:
@krhubert locking whole execution database for dealing with one execution is not efficient. Because we only need to lock operations that will effect the same execution. I don't understand what is your argument here to use level db transactions. There can be 1000s of different executions will operate at the same time, locking every one of them to wait each other to complete is too much.

krhubert:
Ok, so make a benchmark and show me the results. I'm happy to see them. If the version with locking per execution id will be significantly faster/lower in case of cpu/time to lock/memory I don't see any reason why no to go that way.

PR #658, opened 2018-12-20
commands>deploy: add tests with env vars

ilgooz:


commands/service_deploy_test.go
```diff
@@ -13,3 +18,37 @@ func TestServiceDeployCmdFlags(t *testing.T) {
 	flags.Set("env", "a=1,b=2")
 	require.Equal(t, map[string]string{"a": "1", "b": "2"}, c.env)
 }
+
+func TestServiceDeploy(t *testing.T) {
```
krhubert: All tests of command inside one test:
- env
- statuses
- output

Should be spleted

commands/commands_test.go
```diff
@@ -73,3 +74,9 @@ func TestBaseCommandCmd(t *testing.T) {
 	// DO NOT REMOVE this test
 	_ = baseCmd{}.cmd
 }
+
+func readAll(t *testing.T, r io.Reader) []byte {
```
krhubert: Remove readAll and check an error in test.

PR #659, opened 2018-12-20
x/xos>env: rm unneeded cap

ilgooz:


antho1404:
If there is a real really valuable reason to keep that then it should be explicit in the code with a comment, if it's some micro optimization that bring confusion I would suggest to remove it in order to have code that is not confusing

PR #663, opened 2018-12-20
simplify interface/grpc/core>deploy & adopt options in protobuf/coreapi>deploy

ilgooz:


krhubert:
>  in fact moving is better for lowering memory & data usage

The memory sent over the wire is the same when env is null and is outside `oneOf`. In fact, there is even worst because you have to send 2 packets (1st with env and 2nd one with url) to a server so you need to transmit more tcp packets which mean more bytes on the wire. For such small amount it doesn't count in fact, but again this implementation has more memory footprint if you want to use memory as an argument.

> to simplify the implementation on both client & server side. You should compare the previous and current implementations in diffs.

```
diff
+334 lines added
−219 lines remove
```

It seems like you add over 100 line of code to "simplify". Hum that interesting...

> client sends the options first as the first message in the stream but it's also optional to send it. As second, url or chunks should be sent.

In what we have right now, the client sends env with first message + url/chunk (env is optional). What part is difficult here and "unclean"?

Ok you have hard pieces of evidence with memory+diff (btw you can measure the number of packets send via tcpdump/wireshark). You have used memory and diff as your arguments - try to defend them. I want to see the data proving that.

antho1404:
I would suggest to close this issue, I don't really the value that this bring and for me it adds some complexity at the api level

PR #665, opened 2018-12-23
Add favicon and og to docs

NicolasMahe:
Need to be merge in both master and dev.


docs/.vuepress/config.yml
```diff
@@ -1,10 +1,40 @@
-title: MESG
+title: MESG Docs
 description: Build applications connecting any technologies
 ga: UA-117583862-3
 # serviceWorker: true
 markdown:
   toc:
     includeLevel: [2]
+head:
+  - - link
```
antho1404: this seems complicated I'm not even sure I understand what is it, it looks like an array of array with one string and one object. 
Don'e we have a better way to have that ?

antho1404: also linter is failing

NicolasMahe:
Yeah it’s complicated by this is the way with vuepress to provide more tag
in the header.
I had some trouble to convert from the example in JSON they provide on the
vuepress documentation to this yml version.
But it’s working well.

I’m not sure why the linter is failing for this PR. It’s a problem on a .go
file not related to this PR.

Issue #353, opened 2018-08-09
fix LICENCE by adding company info


ilgooz:

NicolasMahe:
I'm not even sure about the License to use. Apache? GNU?
Ethereum is using GNU

krhubert:
@NicolasMahe @antho1404 Have you decided what condition do we want to have in LICENSE. I think LICENSE is very important if someone starts to think about using mesg in their company.

antho1404:
On this topic, I saw most of go project have the license on every files probably because we can use just a single package and not the whole project so we might need to put an automatic task to include the license on every files (eg: https://github.com/ethereum/go-ethereum)

PR #673, opened 2018-12-29
Add license header

NicolasMahe:
Closes https://github.com/mesg-foundation/core/issues/353

I create a script to add the license header to go, js, ts and sh files but I guess it can be improved.
See: https://github.com/mesg-foundation/core/blob/c04cda8deca0936c62309c62512b5d99a0ae522b/scripts/add-license.sh


NicolasMahe:
Nice script @antho1404 👍 

Branch updated!

krhubert:
Just to give a discussion - are you sure we need apache license in every file? Is it mandatory?

If we could keep this global that would be much cleaner and developer (for mesg devs and outside contributors) don't need to think about putting this header for every file.

I've read few topics and both ways (include global and in every file) are ok:
- https://stackoverflow.com/questions/845895/putting-license-in-each-code-file
- https://softwareengineering.stackexchange.com/questions/125836/do-you-have-to-include-a-license-notice-with-every-source-file

I'm aware of [apache apply](https://www.apache.org/licenses/LICENSE-2.0.html#apply) but they don't say it's the only option (and mandatory), rather a guideline.

Also, you can find a bunch of projects including apache license globally
- https://github.com/moby/moby
- https://github.com/mholt/caddy

And per file:
- https://github.com/gohugoio/hugo
- https://github.com/kubernetes/kubernetes

So my preference goes to globally `LICENSE` file as it's way simpler.

antho1404:
For me because of the nature of go that everything is an independent package you can fetch only a part of the repo and in that case not see the license file so having the license everywhere make sure that this "cannot" happen.
After I don't know enough about licenses, maybe even in that case it's enough to have only the license file and in that case I agree we don't need it in every files

krhubert:
> For me because of the nature of go that everything is an independent package you can fetch only a part of the repo and in that case not see the license file so having the license everywhere make sure that this "cannot" happen

I think you can't just use some repo by saying I wasn't aware of the license so I used the source.

NicolasMahe:
Apache license recommends to add the header. I think it’s makes even more
sense with the package approach of the repo.
The script is really fast and we can get a ci task to make sure the files
have the header.

krhubert:
> Apache license recommends to add the header.

this is what I'm saying.- they do not recommend it anywhere, is just one of options to use. Also you languages/files where they don't have comment (for example json/jpg/etc....) and you have to be able to license them. So why to choose the "hard" way with go files.

> The script is really fast and we can get a ci task to make sure the files
have the header.

Script, ci task, etc compared to One file in repo and nothing more is touched :)

antho1404:
i agree that if only the license file is valid then we go for that, if there is no legal reason to have it everywhere so let's not have unnecessary work.

PR #677, opened 2019-01-05
Stream acknowledgement

NicolasMahe:
Implementation of the stream acknowledgement system discussed here:
https://forum.mesg.com/t/acknowledgement-system-on-stream/172


antho1404:
Suggestion: 

I know it might sound bad at the beginning but can't we just update the the `coreClient` listen functions like that we don't need to duplicate this all the time.

I know this is a generated file but we could quite easily inject this line of code in the file when we run the script that generates it and like that we make sure that streams are always waiting for the headers because that the behavior that we want to enforce anyway.

interface/grpc/utils/metadata.go
```diff
@@ -0,0 +1,6 @@
+package utils
```
krhubert: This should be in grpc package. Also:

[Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains.](https://blog.golang.org/package-names)

interface/grpc/utils/metadata.go
```diff
@@ -0,0 +1,6 @@
...
+
+import "google.golang.org/grpc/metadata"
+
+// StatusReady is a metadata to notify a stream is ready.
+var StatusReady = metadata.Pairs("mesg-status", "ready")
```
krhubert:
```suggestion
var MDStatusReady = metadata.Pairs("status", "ready")
```

commands/provider/service_provider.go
```diff
@@ -103,6 +103,12 @@ func (p *ServiceProvider) ServiceListenEvents(id, eventFilter string) (chan *cor
 			}
 		}
 	}()
+
+	// wait for server to notify the stream is ready.
+	if _, err := stream.Header(); err != nil {
```
krhubert: There is not check if we recive metadata with status ready. It's error prone if someone add some metadata header in the future (send before this md), he will break this part.

NicolasMahe:
> Suggestion:
> 
> I know it might sound bad at the beginning but can't we just update the the `coreClient` listen functions like that we don't need to duplicate this all the time.
> 
> I know this is a generated file but we could quite easily inject this line of code in the file when we run the script that generates it and like that we make sure that streams are always waiting for the headers because that the behavior that we want to enforce anyway.

Yeah bad idea.. We need more flexibility to implement this feature. It will take more time to change the generated files (not only go but every other language) than just adding a few lines of code in clients.

commands/provider/service_provider.go
NicolasMahe: I wanted to keep the logic on the client side very simple.

If it need to check the header and its value, then it also need to get the right header (so adding a for around?).

interface/grpc/utils/metadata.go
NicolasMahe: It cannot be put in the grpc package because of import cycling issue..

NicolasMahe:
Yeah that’s a good idea. I was working on a similar version but I didn’t
know where to put the files. Putting it in the same package as the
generated proto is good.

interface/grpc/utils/metadata.go
krhubert: oh, ok, so at least let's change the name of package, maybe just `header`

grpc/header - list of headers + function to manipulate header (for example check values, status, cerating etc..)

commands/provider/service_provider.go
krhubert: But for now it even dosen't check values, so there is a branch you have created and I thought about similar funciton, maybe it could be placed inside the `header` package.

interface/grpc/utils/metadata.go
NicolasMahe: Moved to coreapi

commands/provider/service_provider.go
NicolasMahe: Done

NicolasMahe:
I move the logic to the `coreapi` package.
But because it's in `coreapi`, i didn't implement it for the stream `listenTask` in `serviceapi `. I don't want to duplicate the logic again.
Should i create a new `util` or `stream` or `acknowledgement` package that hold the logic and expose the functions in both `coreapi` and serviceapi`?

krhubert:
> Should i create a new util or stream or acknowledgement package that hold the logic and expose the functions in both coreapi and serviceapi?

> grpc/header - list of headers + function to manipulate header (for example check values, status, cerating etc..)

Could be named stream or ack, depends of current/future usage

protobuf/coreapi/stream.go
```diff
@@ -0,0 +1,38 @@
...
+	if err != nil {
+		return err
+	}
+	// header received. check status.
+	statuses := header.Get(statusKey)
+	statusesLen := len(statuses)
+	if statusesLen == 0 {
+		return fmt.Errorf("stream header does not contain any status")
+	}
+	lastStatus := statuses[statusesLen-1]
```
ilgooz: is there a reason why we have multiple statuses and getting the only last one here?

NicolasMahe:
> > Should i create a new util or stream or acknowledgement package that hold the logic and expose the functions in both coreapi and serviceapi?
> 
> > grpc/header - list of headers + function to manipulate header (for example check values, status, cerating etc..)
> 
> Could be named stream or ack, depends of current/future usage

I create the package `protobuf/acknowledgement` to hold the logic.
The reason I don't want to put it in `interface/grpc` package is this logic should be reusable in other application and service outside of the core. Those app and service have to import the `protobuf/xxx` package to get access to the protobuf generate files, so they can also import `protobuf/acknowledgement` to implement the acknowledgement system.

PR #678, opened 2019-01-05
Repeated parameters

antho1404:
Add support for repeated parameters + validation of the repeated parameters

Still use some reflection but only when it's repeated. If anyone have a simpler way to do it that would be great.


service/validation.go
```diff
@@ -48,7 +49,20 @@ func (v *parameterValidator) Validate(value interface{}) *ParameterWarning {
 		}
 		return v.newParameterWarning("required")
 	}
-
+	if v.parameter.Repeated {
+		// Check if the value is a slice
+		if reflect.TypeOf(value).Kind() != reflect.Slice {
```
krhubert: Shoudn't we have here check for `reflect.Array`?

service/validation.go
```diff
@@ -48,7 +49,20 @@ func (v *parameterValidator) Validate(value interface{}) *ParameterWarning {
...
+			return v.newParameterWarning("not an array")
+		}
+		array := reflect.ValueOf(value)
+		for i := 0; i < array.Len(); i++ {
```
krhubert: Tell me if I'm wrong, but here we can only repeated basic types, we can't repeat whole objects?

service/validation.go
antho1404: I couldn't make it work with `reflect.Array` but I guess we could add both
```go
if reflect.TypeOf(value).Kind() != reflect.Slice && reflect.TypeOf(value).Kind() != reflect.Array {
```

service/validation.go
antho1404: there is no object structure for the moment, this will be another feature. So yes now only basic types because that the only one we have.

But when we manage the nested objects we can just make a recursive call.
`validateType` calls `validate`

NicolasMahe:
I mistakenly merge this PR from a mobile application...
I open a new PR here: https://github.com/mesg-foundation/core/pull/679
Anyway, the branch was based on master and not dev, it's fixed in the new PR.

service/validation.go
NicolasMahe: Why not casting to `[]interface{}`?

NicolasMahe: See https://github.com/mesg-foundation/core/pull/680

PR #679, opened 2019-01-06
Repeated parameters

NicolasMahe:
Same PR as https://github.com/mesg-foundation/core/pull/678 that I merged by mistake.

---

Add support for repeated parameters + validation of the repeated parameters

Still use some reflection but only when it's repeated. If anyone have a simpler way to do it that would be great.


NicolasMahe:
I created a other PR that improve this one: https://github.com/mesg-foundation/core/pull/680
We can use array casting (to `[]interface{}`) rather than type casting.
I think, the problem was coming from test that are using `[]string` types to test array.
In https://github.com/mesg-foundation/core/pull/680 I also change validation test to be based on JSON data rather than Go struct. That way, we don't choose any type, but let `json.Unmarshal` decide.

NicolasMahe:
I just merged https://github.com/mesg-foundation/core/pull/680 into this PR.
Please review and merge!

PR #680, opened 2019-01-06
Repeated parameters improvements

NicolasMahe:
Improvement of https://github.com/mesg-foundation/core/pull/679

- Use type casting instead of reflection
- Update array test types from `[]string` to `[]interface{}`
- Update test to use data from JSON rather than go struct. That way, we don't choose any type, but let json.Unmarshal decide.

### DIFF
--- a/service/validation.go
+++ b/service/validation.go
@@ -2,7 +2,6 @@ package service
 
 import (
 	"fmt"
-	"reflect"
 )
 
 // ParameterWarning contains a specific warning related to a parameter.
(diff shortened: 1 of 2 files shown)


antho1404:
For now we need only to support `[]interface{}`, let's just support that and see if one day we need to support something else we can improve that. Here we just focus on validation of these data that are always from json.
We can add a comment to make sure that there is a different way to do it but for now we don't need it for this feature

PR #682, opened 2019-01-07
Add support of type Any

antho1404:
As described here https://forum.mesg.com/t/add-more-detailed-data-in-the-mesg-yml/126 a new type `Any` has been added that will accept any kind of data.

Question:
Do we still keep the documentation about `Object` (that have validation as a map of string and anything) or should we replace with `Any` to remove confusion ? I replaced from now with `Any`


service/validation_test.go
```diff
@@ -80,6 +84,18 @@ func TestObject(t *testing.T) {
 	require.False(t, validateParameterData("object", 42))
 }
 
+func TestAny(t *testing.T) {
+	require.True(t, validateParameterData("any", map[string]interface{}{
+		"foo": "bar",
+	}))
+	require.True(t, validateParameterData("any", []interface{}{
+		"foo",
+		"bar",
```
krhubert:
```suggestion
		0,
```

service/validation_test.go
```diff
@@ -100,6 +116,7 @@ func TestValidateParameters(t *testing.T) {
 				"object": {
 					"foo": "bar"
 				},
+				"any": 123,
```
krhubert:
```suggestion
				"any": 0,
```

service/validation_test.go
```diff
@@ -113,6 +130,7 @@ func TestValidateParameters(t *testing.T) {
...
```
krhubert:
```suggestion
				"any": 0,
```

service/validation_test.go
```diff
@@ -128,6 +146,7 @@ func TestValidateParameters(t *testing.T) {
 				"number": "string",
 				"boolean": 42,
 				"object": false,
+				"any": 123,
```
krhubert:
```suggestion
				"any": 0,
```

PR #685, opened 2019-01-08
Use go build cache in dev-cli script

krhubert:


NicolasMahe:
I’d like to know the reason as well

krhubert:
```go
package main

import (
	"fmt"
	"os"
)

func main() {
	fmt.Println(os.Args)
}
```

Although there is a cache, `go run` compiles and then removes the binary within each call

```bash
➜  ~ go run main.go
[/tmp/go-build350317863/b001/exe/main]
➜  ~ go run main.go
[/tmp/go-build332451827/b001/exe/main]
```

PR #687, opened 2019-01-09
Fix/execution

krhubert:


database/execution_db.go
```diff
@@ -53,6 +54,15 @@ func (db *LevelDBExecutionDB) Save(execution *execution.Execution) error {
 	if err != nil {
 		return err
 	}
+
+	has, err := db.db.Has([]byte(execution.ID), nil)
+	if err != nil {
+		return err
+	}
+	if has {
+		return fmt.Errorf("database: execution with id %s already exist", execution.ID)
```
NicolasMahe: This is not ok.
Execution needs to be saved multiple times for updating it.

execution/execution.go
```diff
@@ -38,12 +38,12 @@ func (s Status) String() (r string) {
 // Execution stores all informations about executions.
 type Execution struct {
 	ID                string                 `hash:"-"`
-	EventID           string                 `hash:"eventID"`
+	EventID           string                 `hash:"name:eventID"`
```
NicolasMahe: @krhubert can you add a simple test that create 2 execution with different data and compare their hash?

execution/execution_test.go
```diff
@@ -255,3 +257,25 @@ func TestStatus(t *testing.T) {
...
+
+		id := xstructhash.Hash(e, 1)
+		if ids[id] {
+			return false
+		}
+		ids[id] = true
+		return true
+	}
+
+	require.NoError(t, quick.Check(f, nil))
```
NicolasMahe: Nice "hack"!

PR #690, opened 2019-01-12
Support deployments from tarball urls

krhubert:
Need to test , but ready to review


api/deploy.go
```diff
@@ -83,8 +84,17 @@ func newServiceDeployer(api *API, env map[string]string, options ...DeployServic
 	return d
 }
 
-// FromGitURL deploys a service hosted at a Git url.
-func (d *serviceDeployer) FromGitURL(url string) (*service.Service, *importer.ValidationError, error) {
+// FromURL deploys a service from git or tarball url.
+func (d *serviceDeployer) FromURL(url string) (*service.Service, *importer.ValidationError, error) {
+	if !xgit.IsGitURL(url) {
```
antho1404: I would just change with 
```
if xgit.IsGitURL(url) {
  return d.FromGit(url)
}
// normal behavior with tar
```
not blocking just feel easier to understand

antho1404:
@NicolasMahe should we totally remove the git support on the deployment ?

api/deploy.go
```diff
@@ -83,31 +84,41 @@ func newServiceDeployer(api *API, env map[string]string, options ...DeployServic
...
-	d.sendStatus("Service downloaded with success", DonePositive)
-	r, err := xarchive.GzippedTar(path, nil)
+	// if not git repo then it should be tarball
+	resp, err := http.Get(url)
 	if err != nil {
 		return nil, nil, err
 	}
-	return d.deploy(r)
+	defer resp.Body.Close()
+	return d.FromGzippedTar(resp.Body)
```
ilgooz: nothing blocking but for convenience, we can use _deploy()_ here and move line 90 _(defer d.closeStatus())_ to 89.

api/deploy.go
```diff
@@ -83,31 +84,41 @@ func newServiceDeployer(api *API, env map[string]string, options ...DeployServic
...
-	if err := os.RemoveAll(filepath.Join(path, ".git")); err != nil {
-		return nil, nil, err
+		// XXX: remove .git folder from repo.
+		// It makes docker build iamge id same between repo clones.
+		if err := os.RemoveAll(filepath.Join(path, ".git")); err != nil {
+			return nil, nil, err
+		}
+
+		d.sendStatus("Service downloaded with success", DonePositive)
+		r, err := xarchive.GzippedTar(path, nil)
```
ilgooz: we can _Close()_ this reader with defer as well

x/xgit/repository_test.go
```diff
@@ -8,6 +8,22 @@ import (
 	"gopkg.in/src-d/go-git.v4/plumbing"
 )
 
+func TestIsGitURL(t *testing.T) {
```
ilgooz: if we can add some test cases for non-git urls that would be nice

krhubert: But it would require access to the network and that won't be good.

NicolasMahe:
> @NicolasMahe should we totally remove the git support on the deployment ?

If both can be supported without to much work, let's keep git. No breaking change this way ;)

x/xgit/repository.go
```diff
@@ -3,11 +3,44 @@ package xgit
...
+	}
+	c, err := client.NewClient(ep)
+	if err != nil {
+		return nil, err
+	}
+	s, err := c.NewUploadPackSession(ep, nil)
+	if err != nil {
+		return nil, err
+	}
+	return s.AdvertisedReferences()
```
NicolasMahe: Is it the only way to test if the url is a git repository?

NicolasMahe: Is there some more generic version? Like with `content-accept` header?

NicolasMahe: It will be better if the system and generic and we could be compatible with lots of sources by just implementing a interface.

NicolasMahe: Is it what the code is doing?

```
git ls-remote --exit-code -h "$REPO_URL"
```

NicolasMahe:
Deployment from git with branch doesn't work.
```
➜  core git:(feature/deploy-from-tarball-url) ✗ ./dev-cli service deploy https://github.com/mesg-foundation/service-ethereum-erc20#ropsten
✔ Service context received with success
rpc error: code = Unknown desc = archive/tar: invalid tar header
```

NicolasMahe:
> * this PR was about support tar.gz which was achive, let create topic and discuss if we want to support more archive extension (if yes which one etc....)

tar.gz doesn't work as well...

```
➜  core git:(feature/deploy-from-tarball-url) ✗ ./dev-cli service deploy https://api.github.com/repos/mesg-foundation/service-ethereum-erc20/tarball/ropsten
✔ Service context received with success
Service is not valid
to get more information, run: mesg-core service validate
```

antho1404:
Looks like tarball doesn't work because when untar the file it create another folder in the temp folder
eg: `service-webhook.tar.gz` will create `tmpfolder/service-webhook` and we are looking for the `mesg.yml` inside `tmpfolder` directly.

dirty hack is updating the tempPath when saving the context, something like that:
```go
// saveContext downloads service context to a temp dir.
func (s *Service) saveContext(r io.Reader) error {
	var err error
	s.tempPath, err = s.createTempDir()
	if err != nil {
		return err
	}

	s.sendStatus("Receiving service context...", DRunning)
	defer s.sendStatus("Service context received with success", DDonePositive)

	if err := archive.Untar(r, s.tempPath, &archive.TarOptions{
		Compression: archive.Gzip,
		NoLchown:    true,
	}); err != nil {
		return err
	}
	x, err := ioutil.ReadDir(s.tempPath)
	s.tempPath = filepath.Join(s.tempPath, x[0].Name())
	return nil
}
```

Like that `tar.gz` works great

krhubert:
Yes, but you assume that someone pack a direcotry into tar, what if someone don't? do we require tarball inside direcotry ?

antho1404:
I just debug based on the one from github after I don't remember any tar that I downloaded that was not in a directory so I might be ok to require tarball inside a directory.
Maybe we can look if the name of the file is `filename.tag.gz` then we look for the folder `filename`, if present it means it's a tarball inside a directory otherwise it's not.

service/service.go
```diff
@@ -178,26 +178,30 @@ func (s *Service) computeHash() string {
...
-		Compression: archive.Gzip,
-		NoLchown:    true,
-	})
-}
-
-func (s *Service) createTempDir() (path string, err error) {
-	return ioutil.TempDir("", "mesg-"+uuid.NewV4().String())
-}
+	if err := archive.Untar(r, s.tempPath, &archive.TarOptions{
+		NoLchown: true,
```
antho1404: why removing the `Compression: archive.Gzip,` ? Can we still support the `tar.gz` without ?

krhubert: Because Compression is unused while Untar (only works for `archive.Tar(WithOptions)`.

Untar detects compression so any of [uncompressed, bzip2, gzip, xz] can be used (all of them are supported by docker)

NicolasMahe:
TODO
- [ ] add test (-36% path)
- [ ] Service's hash is different from downloaded tarball and git. If I download the tarball manually and deploy it, the hash is the same as git....

PR #691, opened 2019-01-12
Add db transaction mechanism to database execution

krhubert:


database/execution_db.go
```diff
@@ -12,7 +13,16 @@ import (
 type ExecutionDB interface {
 	Find(executionID string) (*execution.Execution, error)
```
NicolasMahe: What about removing Find and Save function from `ExecutionDB` to keep only the one in `ExecutionTransaction`?
It means to update codebase, but I think it's only 1 or 2 files max, and the `ExecutionDB` will work only one way.

krhubert: I thought about this, but in that way we will have to open transaction every time and keeping find, save without opening transaction is faster and more "clean" outside database package

PR #693, opened 2019-01-12
Remove vendor

krhubert:


antho1404:
It looks like we need to force the modules with go 
```
go get: warning: modules disabled by GO111MODULE=auto in GOPATH/src;
	ignoring go.mod;
	see 'go help modules'
```

PR #694, opened 2019-01-13
Support Object type

krhubert:
TODO: 

- write tests (manual tests passes even with nested array objects)

```
{
  "message": ["a", "b"],
  "myobj": {
    "x1": "string",
    "x2": 1 
  },
  "myobjarr": [{
    "x1": {
      "x1": "string",
      "x2": 1 
    },
    "x2": 1 
  }]
}
```

- @NicolasMahe can you help with schema.json (how to tell that object type must be `definitions/parameters` - `"$ref": "#/definitions/parameters"` ?

- figure out better name for defParametersMapToService

- `Object map[string]*Parameter hash:"name:7"`  - i don't like that hash with number `7` is inside type, move object to last position in struct

- better error messages for nested objects in parameter warning.


krhubert:
Close in favor of #696, as it's more complete

PR #716, opened 2019-01-16
Remove system service resolver

NicolasMahe:
and ALL its vendor!


systemservices/systemservices.go
```diff
@@ -10,15 +10,11 @@ import (
 
 // list of system services.
 // these names are also relative paths of system services in the filesystem.
-const (
-	ResolverService = "resolver"
-)
+const ()
```
antho1404: I'm surprised this compiles

PR #717, opened 2019-01-16
Create stream autoreconnect

krhubert:


protobuf/coreapi/client_safe.go
```diff
@@ -0,0 +1,344 @@
...
+		ctx:  ctx,
+		in:   in,
+		opts: opts,
+	}
+	go c.recvLoop()
+	return c
+}
+
+// recvLoop recives ListenEvent response in loop and reconnect in on error.
+func (s *core_ListenEventClient) recvLoop() {
```
NicolasMahe: Is there any way to not have this function duplicated over and over?

Maybe with a generic version (using `interface{}`) and specific ones that cast the generic?

protobuf/coreapi/client_safe.go
```diff
@@ -0,0 +1,344 @@
...
+			s.c <- &core_ListenEventClientResponse{nil, err}
+			continue
+		}
+		if err := acknowledgement.WaitForStreamToBeReady(s.Core_ListenEventClient); err != nil {
+			s.c <- &core_ListenEventClientResponse{nil, err}
+			continue
+		}
+
+		// buffered channel, because it might happen that ctx.Done
+		// will be notifed first before stream.Recv in for loop.
```
NicolasMahe:
```suggestion
		// will be notified first before stream.Recv in for loop.
```

protobuf/coreapi/client_safe.go
```diff
@@ -0,0 +1,344 @@
...
+	for {
+		// connect
+		s.Core_ListenResultClient, err = s.client.listenResult(s.ctx, s.in, s.opts...)
+		if err != nil {
+			s.c <- &core_ListenResultClientResponse{nil, err}
+			continue
+		}
+
+		// buffered channel, because it might happen that ctx.Done
+		// will be notifed first before stream.Recv in for loop.
```
NicolasMahe:
```suggestion
		// will be notified first before stream.Recv in for loop.
```

protobuf/coreapi/client_safe.go
```diff
@@ -0,0 +1,344 @@
...
+			s.c <- &core_ServiceLogsClientResponse{nil, err}
+			continue
+		}
+		if err := acknowledgement.WaitForStreamToBeReady(s.Core_ServiceLogsClient); err != nil {
+			s.c <- &core_ServiceLogsClientResponse{nil, err}
+			continue
+		}
+
+		// buffered channel, because it might happen that ctx.Done
+		// will be notifed first before stream.Recv in for loop.
```
NicolasMahe:
```suggestion
		// will be notified first before stream.Recv in for loop.
```

protobuf/serviceapi/client_safe.go
```diff
@@ -0,0 +1,135 @@
...
+	for {
+		// connect
+		s.Service_ListenTaskClient, err = s.client.listenTask(s.ctx, s.in, s.opts...)
+		if err != nil {
+			s.c <- &core_ListenTaskClientResponse{nil, err}
+			continue
+		}
+
+		// buffered channel, because it might happen that ctx.Done
+		// will be notifed first before stream.Recv in for loop.
```
NicolasMahe:
```suggestion
		// will be notified first before stream.Recv in for loop.
```

protobuf/coreapi/client_safe.go
krhubert: I think this might be over complicated, for generic version we need to provide diffrent:

- api call (`ListenEvent`)
- api call request (`ListenEventRequest`)
- api call response (`ListenEventResponse`)
- data wrapper (`core_ListenEventClientResponse`)
- client (`Core_ListenEventClient`)

So there is a log of different values for 40 line function. I do not like copies of this func as well, but generic is way more complicated there. The best solution will be autogenerate code from protobuf :D

NicolasMahe:
@krhubert please provide code example to reproduce the gRPC error and test this PR properly!

ilgooz:
I think this SafeClient introduces too much complexity and it requires to be maintained. I'd go with re-creating dead streams directly in the code maybe with recursive calls. That way it's way simpler and we don't need to introduce a complicated logic like this and maybe it'll introduce some side effects.

For example in the _commands/provider/service_provider.go_ we can do:
```go
func (p *ServiceProvider) listenResults(errC, resultC) {
  stream, err := p.client.ListenResult()
  if err != nil {
    return err
  }
  go func(){
    for {
      res, err := stream.Recv()
      if err != nil {
        if err == io.EOF {
          p.listenResult(errC, resultC)
          return
        }
        errC <- err
      }
      resultC <- res
  }()
  if err := acknowledgement.WaitForStreamToBeReady(stream); err != nil {
    return nil, nil, err
  }
}

func (p *ServiceProvider) ServiceListenResults(id, taskFilter, outputFilter string, tagFilters []string) (chan *coreapi.ResultData, chan error, error) {
  resultC := make(chan *coreapi.ResultData)
  errC := make(chan error)
  if err := p.listenResult(id, taskFilter, outputFilter string, tagFilters []string, errC, resultC); err != nil {
    return nil, nil, err
  }
  return resultC, errC, nil
}
```

Or we could do this in the higher level for example [here](https://github.com/mesg-foundation/core/blob/047d33101171a1e67b72a45137851c13976eec71/commands/service_dev.go#L100) by calling this func recursively when stream is dead.

protobuf/coreapi/client_safe_test.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	s := grpc.NewServer()
+	RegisterCoreServer(s, m)
+	reflection.Register(s)
+	go func() { s.Serve(lis) }()
+	return s
+}
+
+func newCoreClinetSafe(t *testing.T) *CoreClientSafe {
```
NicolasMahe:
```suggestion
func newCoreClinetSafe(t *testing.T) *CoreClientSafe {
```
```suggestion
func newCoreClientSafe(t *testing.T) *CoreClientSafe {
```

antho1404:
Should we close this one now that we solved the disconnection issue ?

krhubert:
@antho1404 in fact nope, this is the client we should use because disconnection is different issue. Here we guarantee that we reconnect on any error which is important as well.

PR #722, opened 2019-01-20
Remove bind to systemservice local folder when starting the core

NicolasMahe:
Quick fix to master for removing the bind to systemservice local folder when starting the core.

Once merge we can release a bug fix version.

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -2,7 +2,6 @@ package daemon
 
 import (
 	"io"
-	"path/filepath"
 
 	"github.com/mesg-foundation/core/config"
 	"github.com/mesg-foundation/core/container"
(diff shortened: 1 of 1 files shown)


antho1404:
the install script has been updated https://github.com/mesg-foundation/mesg.com/pull/77
I don't see any reasons to touch at the core, also removing it to add it again next week and having to release a quick fix version for that i feel it's unnecessary especially that the install script is taking care of it and if we do a new version they will have to go to the install script again anyway

NicolasMahe:
> the install script has been updated [mesg-foundation/mesg.com#77](https://github.com/mesg-foundation/mesg.com/pull/77)
> I don't see any reasons to touch at the core, also removing it to add it again next week and having to release a quick fix version for that i feel it's unnecessary especially that the install script is taking care of it and if we do a new version they will have to go to the install script again anyway

We may not need this bind anyway to build system service. They can be deployed from an url.. I should have remove it as the same time as the other system service.

And for the marketplace system service, we will need to new version, so adding it (if necessary) later is fine.

antho1404:
I'm fine with this but it doesn't need a version for that, let's wait to have few other feedback and issue a version when it's really critical, let's not rush creating a version for one fix that is solved by the install script anyway

NicolasMahe:
I agree. Let's say we release it Tuesday.

PR #724, opened 2019-01-22
Remove system services

antho1404:
Removing the system services with the deployment associated.
The goal is to remove all system services and only be able to plug services when the core starts and that the core's codebase will have knowledge about.


.dockerignore
```diff
@@ -17,4 +17,4 @@ scripts/
 docker-images/
 
 # prevent go build cache to break inside Docker on changes to systemservices/sources.
```
NicolasMahe:
```suggestion
# prevent go build cache to break inside Docker on changes to systemservices.
```

ilgooz: removed in https://github.com/mesg-foundation/core/pull/727

antho1404: oups i just merged it in this one at the same time

PR #726, opened 2019-01-23
Add compatibility with log force colors option

NicolasMahe:
Add compatibility with log force colors option to get nice colored logs from the core!

image

- The command `start` accept a flag `--log-force-colors`
- The script `dev-core` read `ENV.MESG_LOG_FORCECOLORS` to start the core with the right config

So I advise you to add to you `.bashrc` / `.zshrc`:
```
export MESG_LOG_FORCECOLORS=true
```


dev-core
```diff
@@ -53,5 +54,5 @@ function onexit {
   ./bin/cli stop
 }
 
-./bin/cli start --log-level "$MESG_LOG_LEVEL"
+./bin/cli start --log-level "$MESG_LOG_LEVEL" --log-force-colors "$MESG_LOG_FORCECOLORS"
```
NicolasMahe: is it really necessary to transform from ENV to flag? seems useless

krhubert: Yes, because not everyone likes color and debug mode by default ;)

And for them this is option to change.

antho1404: This one will always set the flag as true (the value is ignored by the cli)

should do something like that

```bash
if [[ "$MESG_LOG_FORCECOLORS" != "" ]]; then 
  forceColor="--log-force-colors"
fi

./bin/cli start --log-level "$MESG_LOG_LEVEL" $forceColor
```

Not sure this works but something like that to only put the flag when the env variable is present otherwise we don't put it.

@krhubert you almost had colors in your terminal ;)

dev-core
```diff
@@ -4,8 +4,10 @@ set -e
 
 DIR=$(pwd)
 MESG_LOG_LEVEL=${MESG_LOG_LEVEL:-debug}
+MESG_LOG_FORCECOLORS=${MESG_LOG_FORCECOLORS:-true}
```
antho1404: Set default to false otherwise we have colors by default and we need explicitly to set MESG_LOG_FORCECOLORS=false to have the json version

```suggestion
MESG_LOG_FORCECOLORS=${MESG_LOG_FORCECOLORS:-false}
```

krhubert: Ok, I thought we want to have color by default ;)

antho1404: no only when we put the `FORCECOLORS` env variable. Basically we just add a new type of log

NicolasMahe:
@krhubert @ilgooz please review

PR #727, opened 2019-01-23
rm system services residuals

ilgooz:


.dockerignore
```diff
@@ -14,7 +14,4 @@ mesg-dev
 bin/cli
 docs/
 scripts/
-docker-images/
-
-# prevent go build cache to break inside Docker on changes to systemservices.
-systemservices
```
antho1404: I was thinking to still keep this directory to keep sources of the services. Just to store everything as monorepo but it's not needed anymore for the core to run so I think this can stays

NicolasMahe: Yeah let's keep it!

NicolasMahe: @ilgooz could you revert this modif?

ilgooz: done

PR #728, opened 2019-01-23
Add ethwallet source

NicolasMahe:
The build time of this service is quite long because it's building a crypto library.

I will try to create a docker image that already has the compiled crypto lib.

A folder `test-data` contains test payloads to easily test the service. Adapt their content accordingly.

## Create

```
mesg-core service execute ethwallet --task create --json ./test-data/create.json
```

## List

```
mesg-core service execute ethwallet --task list --json ./empty.json
```

## Delete

```
mesg-core service execute ethwallet --task delete --json ./test-data/delete.json
```

## Export

```
mesg-core service execute ethwallet --task export --json ./test-data/export.json
```

## Import

```
mesg-core service execute ethwallet --task import --json ./test-data/import.json
```

## Sign

```
mesg-core service execute ethwallet --task sign --json ./test-data/sign.json
```


antho1404:
Few feedback that are not blocking, the service works fine otherwise 👍 

### Output of delete
delete task returns `null` that would be better to return something we should always have output, here the address we just deleted could be helpful

### Rename task `new` into `create`
I feel create is easier to understand, we want to create a new address 

### Export and re-import
When we export an account we have an json `{"account": { ... }}` but this json is not accepted for the import.

Other than that everything is good

NicolasMahe:
> Few feedback that are not blocking, the service works fine otherwise 👍
> 
> ### Output of delete
> delete task returns `null` that would be better to return something we should always have output, here the address we just deleted could be helpful
> 
> ### Rename task `new` into `create`
> I feel create is easier to understand, we want to create a new address
> 
> ### Export and re-import
> When we export an account we have an json `{"account": { ... }}` but this json is not accepted for the import.
> 
> Other than that everything is good

Modification done 👍

antho1404:
Might need some optimization of the build, maybe with a specific pre built docker image to skip some steps in the build but this can be done in another PR

systemservices/ethwallet/ethwallet/ethwallet.go
```diff
@@ -0,0 +1,47 @@
...
+
+// New creates a new instance of EthWallet
+func New() (*Ethwallet, error) {
+	// mesg client
+	service, err := service.New()
+	if err != nil {
+		return nil, err
+	}
+
+	// keystore
```
krhubert:
```suggestion
```

systemservices/ethwallet/ethwallet/ethwallet.go
```diff
@@ -0,0 +1,47 @@
...
+
+// Ethwallet is a Ethereum Wallet
+type Ethwallet struct {
+	service  *service.Service
+	keystore *keystore.KeyStore
+}
+
+// New creates a new instance of EthWallet
+func New() (*Ethwallet, error) {
+	// mesg client
```
krhubert:
```suggestion
```

systemservices/ethwallet/ethwallet/sign.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type signInputs struct {
+	Address     string       `json:"address"`
+	Passphrase  string       `json:"passphrase"`
+	Transaction *transaction `json:"transaction"`
+	ChainID     int64        `json:"chainID"`
+}
+
+type transaction struct {
```
krhubert: https://github.com/ethereum/go-ethereum/blob/c1c43011211e70ffdd2a78fb1a467adb5db9f65e/core/types/transaction.go#L46

- Replace `Value` and `GasPrice with big.Int (it implemnts unmarshal json interface https://golang.org/pkg/math/big/#Int.UnmarshalJSON)
- rename `Data` json tag to `"input"`
- rename `Data` to either `Payload` or `Input` 
- rename `Value` To `Amount` (not mandatory)
- rename `To` -> `Recipient` (not mandatory)

systemservices/ethwallet/ethwallet/list.go
```diff
@@ -0,0 +1,20 @@
+package ethwallet
+
+import (
+	"github.com/mesg-foundation/core/client/service"
+)
+
+type listOutputSuccess struct {
+	Addresses []string `json:"addresses"`
```
krhubert: replace string with `accounts.Account`?

Or at least `common.Address` as this is standard thype for address that has already implemented json encoding and more

systemservices/ethwallet/ethwallet/list.go
```diff
@@ -0,0 +1,20 @@
...
+}
+
+func (s *Ethwallet) list(execution *service.Execution) (string, interface{}) {
+	addresses := make([]string, 0)
```
krhubert: Use `make([]string, len(s.keystore.Accounts())`

systemservices/ethwallet/ethwallet/sign.go
```diff
@@ -0,0 +1,63 @@
...
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/mesg-foundation/core/client/service"
+	"github.com/mesg-foundation/core/systemservices/ethwallet/x/xgo-ethereum/xaccounts"
+)
+
+type signInputs struct {
+	Address     string       `json:"address"`
```
krhubert: replace with `common.Address` or `accounts.Acount`

systemservices/ethwallet/ethwallet/sign.go
```diff
@@ -0,0 +1,63 @@
...
+	Passphrase  string       `json:"passphrase"`
+	Transaction *transaction `json:"transaction"`
+	ChainID     int64        `json:"chainID"`
```
krhubert: ChainID is *big.Int 
https://github.com/ethereum/go-ethereum/blob/ecb781297bfc891f4ff26bdf3fda362744bbb3e3/params/config.go#L148

Somewhere I saw uint64, but not int64

systemservices/ethwallet/x/xgo-ethereum/xaccounts/accounts.go
```diff
@@ -0,0 +1,23 @@
...
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/accounts/keystore"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// GetAccount retrieves an account from its address.
+func GetAccount(keystore *keystore.KeyStore, address string) (account accounts.Account, err error) {
+	err = fmt.Errorf("Account not found")
```
krhubert:
```go
func GetAccount(keystore *keystore.KeyStore, address common.Address) (accounts.Account, error) {
	for _, account := range keystore.Accounts() {
		if account.Address == hexaddress {
			return account, nil
		}
	}
	return nil, fmt.Errorf("Account not found")
}
```

Plus you can't use a `HexToAddress` here because it internally [drops](https://github.com/ethereum/go-ethereum/blob/f91312dbdbb9e04ef578946226e5d8069d5dfd5a/common/bytes.go#L97) all errors

systemservices/ethwallet/ethwallet/create.go
```diff
@@ -0,0 +1,29 @@
...
+type createInputs struct {
+	Passphrase string `json:"passphrase"`
+}
+
+type createOutputSuccess struct {
+	Address string `json:"address"`
+}
+
+func (s *Ethwallet) create(execution *service.Execution) (string, interface{}) {
+	var inputs createInputs
```
krhubert: Not a single method has validation for address

antho1404:
let's merge this one if everything is ok except the big number, this is another subject, we can implement this later if needed but let's move forward

systemservices/ethwallet/ethwallet/sign.go
NicolasMahe: Because of web3 transaction definition: https://web3js.readthedocs.io/en/1.0/web3-eth.html#eth-sendtransaction

krhubert: Ok so this is sign transaction, my bad.

https://github.com/ethereum/go-ethereum/blob/d3441ebb563439bac0837d70591f92e2c6080303/signer/core/types.go#L95:6

So there `GasLimit` Should be changed to `Gas` then, because it is gas in web3 and go-ethereum.

Also the field From is missing but I think it's irrelevant.

PR #730, opened 2019-01-24
Create new struct hash package

krhubert:
features:

- panic on an invalid tag (thanks to it one bug was found in Dependency)
- can use `omitempty` filed and extend structs
- more test coverage
- support complex type 🏆 
- tags `version` `method` and `lastversion` are not included because core doesn't' use it, but it could be easily extended


NicolasMahe:

utils/structhash/structhash.go
```diff
@@ -0,0 +1,199 @@
...
+)
+
+// Sha1 takes a data structure and returns its md5 hash as string.
+func Sha1(v interface{}) [sha1.Size]byte {
+	return sha1.Sum(serialize(v))
+}
+
+// Sha1 takes a data structure and returns its md5 hash as string.
+func Sha3(v interface{}) [64]byte {
+	return sha3.Sum512(serialize(v))
```
NicolasMahe: Wrong command and why sha3 is 512 bits?

NicolasMahe:
This PR is way less important since PR https://github.com/mesg-foundation/core/pull/731 has been merged.
The core still need to hash struct for event and execution but their lifetime is much smaller and thus having a different hash after an upgrade of the core doesn't impact.
It could only have impact with the network with nodes running different version of core. But anyway, the "omitempty" will not resolve the problem.

I suggest to keep the current hash package (https://github.com/cnf/structhash)

@krhubert any suggestion why we should merge this PR?

PR #482, opened 2018-09-20
service hash changes for every deploy (when fetching from git) #close 379

krhubert:
Fix for https://github.com/mesg-foundation/core/issues/379

PR #731, opened 2019-01-24
Deterministic hash

NicolasMahe:
Implementation of https://forum.mesg.com/t/services-hash-are-different-from-downloaded-tarball-and-remote-git/211/5

Docker hash are not deterministic.

This PR use the archive to create the service hash.
It very simple are works well:

From github archive
```
./dev-cli service deploy https://github.com/mesg-foundation/service-ethereum/archive/master.tar.gz
✔ Service deployed with hash: da39a3ee5e6b4b0d3255bfef95601890afd80709
```

From local git
```
./dev-cli service dev /Users/nico/Developments/MESG/services/service-ethereum
✔ Service deployed with ID: da39a3ee5e6b4b0d3255bfef95601890afd80709
```

From remote git
```
./dev-cli service dev https://github.com/mesg-foundation/service-ethereum
✔ Service deployed with ID: da39a3ee5e6b4b0d3255bfef95601890afd80709
```

Please try the remote ones on your computer and tell me if you have the same or different hash!


antho1404:
same hashes for me for the 3 deployment methods :)

service/service.go
```diff
@@ -103,6 +103,13 @@ func New(tarball io.Reader, env map[string]string, options ...Option) (*Service,
 	}
 	defer os.RemoveAll(s.tempPath)
 
+	// calculate hash
```
krhubert: It won't be deterministic here because tarball could be (un)compressed with diffrent algorithms.

Also tarball may have `.git` directory which will be removed when deploying.

antho1404: we are controlling the uncompression, if it's different then it's a different service. Same as the .git (but I agree it would be better to remove the .git when we create the tar)

ilgooz: Forcing on one compression & archive type will limit the flexibility and we'll not be able to accept for example _.zip_ formats but at the of the day I think **we need this limitation**.

Because, in the marketplace we'll store service's hash in smart contracts. So services will not be there with their source code & hash together. When devs want to do their hash comparasation once they download a service, they can simply tar.gz services, calculate a sha1 and compare with the one on the marketplace. So this is very useful on security side and doesn't require devs to have special tools.

NicolasMahe: The goal is to have the same hash when the service is well done (ignore .git, .DS_Store, dependency with tag or commit hash, etc...) and when the source is the same.
This is a good improvement over the version based on docker image, but it cannot be perfect because there are so many variable in place..

service/service.go
```diff
@@ -216,9 +214,14 @@ func (s *Service) deploy() error {
 	s.sendStatus("Image built with success", DDonePositive)
 
 	s.configuration().Image = imageHash
+	// TODO: FOR TEST ONLY
+	if s.Hash == "" {
+		panic("s.Hash is empty")
```
ilgooz: needs to be removed?

antho1404: @NicolasMahe why there is this panic here ? Does this code need to be removed ?

service/namespace.go
```diff
@@ -11,7 +14,12 @@ const resultChannel string = "Result"
 
 // namespace returns the namespace of the service.
 func (service *Service) namespace() []string {
-	return []string{service.Hash}
+	h, err := hex.DecodeString(service.Hash)
+	if err != nil {
+		panic(err)
+	}
+	sum := sha1.Sum([]byte(h))
```
NicolasMahe: Is it really necessary to `DecodeString`?
Isn't it possible to just `sum := sha1.Sum([]byte(service.Hash))`?

krhubert: No it's not, but it will be more constant in the future when we will change sevice.Hash to []byte

antho1404:
This is not constant anymore with git repositories.
It's working correctly with the archive but the git version doesn't generate the same hash and multiple deployment of the same git generate different hashes

```bash
➜  core git:(feature/determistic-hash) ./dev-cli service deploy https://github.com/mesg-foundation/service-ethereum
✔ Service downloaded with success
✔ Service context received with success
✔ Image built with success
✔ Service deployed with hash: 6151ac5d62e7aed8b4c7cb88dd75fb59334a56c8b3f1e0b203292e069cea7d38
To start it, run the command:
        mesg-core service start 6151ac5d62e7aed8b4c7cb88dd75fb59334a56c8b3f1e0b203292e069cea7d38
➜  core git:(feature/determistic-hash) ./dev-cli service deploy https://github.com/mesg-foundation/service-ethereum
✔ Service downloaded with success
✔ Service context received with success
✔ Image built with success
✔ Service deployed with hash: 474c94a881ac297e9b36a4c0e190798226bc1f43eaf92adf786d36a4b2762a1f
To start it, run the command:
        mesg-core service start 474c94a881ac297e9b36a4c0e190798226bc1f43eaf92adf786d36a4b2762a1f
```

NicolasMahe:
I got a panic when listing services. (i also merge dev into the branch, so maybe some bug appears)

```
panic: encoding/hex: invalid byte: U+00EF 'ï'

goroutine 91 [running]:
github.com/mesg-foundation/core/service.(*Service).namespace(0xc0003a3960, 0x0, 0x0, 0x0)
        /Users/nico/Developments/MESG/core/service/namespace.go:19 +0x12e
github.com/mesg-foundation/core/service.(*Dependency).namespace(0xc0003a51e0, 0x0, 0x8779, 0xa96800000000)
        /Users/nico/Developments/MESG/core/service/namespace.go:27 +0x36
github.com/mesg-foundation/core/service.(*Dependency).Status(0xc0003a51e0, 0x86f2, 0x0, 0x9da400000000)
        /Users/nico/Developments/MESG/core/service/status.go:68 +0x2b
github.com/mesg-foundation/core/service.(*Service).Status(0xc0003a3960, 0xd417e0, 0xc00041a600, 0xa56200000000)
        /Users/nico/Developments/MESG/core/service/status.go:46 +0x106
github.com/mesg-foundation/core/interface/grpc/core.(*Server).ListServices.func1(0xc00041a600, 0xc00041e660, 0xc00041a5f0, 0xc00041c960, 0xc0003a3960)
        /Users/nico/Developments/MESG/core/interface/grpc/core/list_services.go:32 +0x5c
created by github.com/mesg-foundation/core/interface/grpc/core.(*Server).ListServices
```

krhubert:
@antho1404 should be fixed now, can you confirm?

antho1404:
This is still not working, deploying from the tar is different than deploying from the git (deploying 2 times the git generate the same hash and deploying 2 times the tar generate the same hash, just `tar != git`)

There is also one other bug but regression from the `dev` I will create a separated issue

```
➜  core git:(feature/determistic-hash) ./dev-cli service list
HASH    SID    NAME    STATUS
➜  core git:(feature/determistic-hash) ./dev-cli service deploy https://github.com/mesg-foundation/service-ethereum && ./dev-cli service list | grep com.mesg.ethereum
✔ Service downloaded with success
✔ Service context received with success
✔ Image built with success
✔ Service deployed with id: com.mesg.ethereum
To start it, run the command:
        mesg-core service start com.mesg.ethereum
10cba92f00b8b1821bcda583e0ac0e90b48a4243f7b0bc04c45ee17b3cbf4e6b    com.mesg.ethereum    Ethereum    stopped
➜  core git:(feature/determistic-hash) ./dev-cli service deploy https://github.com/mesg-foundation/service-ethereum/archive/master.tar.gz && ./dev-cli service list | grep com.mesg.ethereum
✔ Service context received with success
✔ Image built with success
✔ Service deployed with id: com.mesg.ethereum
To start it, run the command:
        mesg-core service start com.mesg.ethereum
10cba92f00b8b1821bcda583e0ac0e90b48a4243f7b0bc04c45ee17b3cbf4e6b    com.mesg.ethereum    Ethereum    stopped
03d8dd8ed484c6605e8e0b1806dba9dcc13f450494d1edff5789f355de4e2df4    com.mesg.ethereum    Ethereum    stopped
```

service/service.go
```diff
@@ -98,16 +99,43 @@ func New(tarball io.Reader, env map[string]string, options ...Option) (*Service,
...
+	}); err != nil {
+		return nil, err
+	}
+
+	// tar it back without any compression to get constant hash
+	// despite used compression from tarball
+	t, err := archive.Tar(s.tempPath, archive.Gzip)
+	if err != nil {
+		return nil, err
+	}
```
NicolasMahe: Let's move this code from line 112 (remove .git) to line 131 (the creation of the tar) to `api/deploy.go`.
`api/deploy.go` should download AND clean the downloaded stuff (like it's already doing for git), and create a clean tarball to `Service.New`. In the future, we can pass a path between `api/deploy.go` and `Service.New` to avoir tar/untar again, but this should be done in a separate PR.

service/service.go
```diff
@@ -98,16 +99,43 @@ func New(tarball io.Reader, env map[string]string, options ...Option) (*Service,
...
 
+	// XXX: remove .git folder from repo.
+	// It makes docker build iamge id same between repo clones.
+	if err := os.RemoveAll(filepath.Join(s.tempPath, ".git")); err != nil {
+		return nil, err
+	}
+
+	// XXX: change the access and modification times of service to get constant hash
+	zerot := time.Time{}
+	if err := filepath.Walk(s.tempPath, func(path string, _ os.FileInfo, _ error) error {
```
NicolasMahe: I wonder if this is really useful. Git and tar concern the timestamp of file, don't they?

service/service.go
NicolasMahe: Also, the check for one folder in the root folder should also be moved to `api/deploy`:
https://github.com/mesg-foundation/core/blob/6700ccf4767899795308a5572d64bd57a0b45ff8/service/service.go#L220-L222

NicolasMahe: In the future for optimization reason, `Service.New` should accept a path instead of an `io.reader` to avoid unnecessary untar/tar.
This way, `api/deploy` downloads from wherever to a local folder, clean it, pass it to `Service.New`.
`Service.New` can create a tar if needed to calculate hash.

service/service.go
krhubert: Yes, the do. WIthout it ever git clone will be diffrent :/

antho1404: Why is it different, we had this problem before and removing the .git was the solution because we had a file from the logs in git that was different. You can look this PR here https://github.com/mesg-foundation/core/pull/482.

krhubert: It's different problem. Git was writing timestamp to the file. Here tar save mod/access/change time of the files.

You can check out [Header](https://golang.org/pkg/archive/tar/#Header) for tar archive file

NicolasMahe:

service/namespace.go
```diff
@@ -11,7 +14,12 @@ const resultChannel string = "Result"
 
 // namespace returns the namespace of the service.
 func (service *Service) namespace() []string {
-	return []string{service.Hash}
+	h, err := hex.DecodeString(service.Hash)
+	if err != nil {
+		panic(err)
```
antho1404: I don't really like the idea to have a panic here (or even the fatal)
```suggestion
		logrus.Fatal(err)
```

service/namespace.go
```diff
@@ -11,7 +14,12 @@ const resultChannel string = "Result"
 
 // namespace returns the namespace of the service.
 func (service *Service) namespace() []string {
-	return []string{service.Hash}
+	h, err := hex.DecodeString(service.Hash)
```
ilgooz: why not just using `h := []byte(service.Hash)` since hash is not actually a hex encoded word

ilgooz: why not just using `h := []byte(service.Hash)` since hash is not actually a hex encoded word

api/deploy.go
```diff
@@ -86,44 +86,58 @@ func newServiceDeployer(api *API, env map[string]string, options ...DeployServic
...
+	if err != nil {
+		return nil, nil, err
+	}
+	defer r.Close()
+	return d.deploy(r)
 }
 
 // FromArchive deploys a service from a gzipped tarball.
 func (d *serviceDeployer) FromArchive(r io.Reader) (*service.Service, *importer.ValidationError, error) {
+	d.sendStatus("Receiving service context...", Running)
```
ilgooz: I think it's better to move this _service_ pkg and also keep the _Service context received (done)_ message. This way we can show the status info about sending service context for both _FromArchive()_ & _FromURL()_

api/deploy.go
```diff
@@ -86,44 +86,58 @@ func newServiceDeployer(api *API, env map[string]string, options ...DeployServic
 
 // FromURL deploys a service from git or tarball url.
 func (d *serviceDeployer) FromURL(url string) (*service.Service, *importer.ValidationError, error) {
+	path, err := d.createTempDir()
+	if err != nil {
+		return nil, nil, err
+	}
+	d.sendStatus("Downloading service...", Running)
 	defer d.closeStatus()
```
antho1404: Can `defer os.RemoveAll(path)` here

service/service.go
```diff
@@ -167,43 +185,6 @@ func DeployStatusOption(statuses chan DeployStatus) Option {
...
-// computeHash computes a unique sha1 value for service.
-// changes on the names of constant configuration fields of mesg.yml will not
-// have effect on computation but extending or removing configurations or changing
-// values in mesg.yml will cause computeHash to generate a different value.
-func (s *Service) computeHash() string {
-	return xstructhash.Hash(s, 1)
-}
-
-// saveContext downloads service context to a temp dir.
-func (s *Service) saveContext(r io.Reader) error {
```
ilgooz: I prefer to keep this func as a wrapper so _New_ won't get too big

utils/tarsum/fileinfohashes.go
```diff
@@ -0,0 +1,61 @@
+package tarsum
+
+import (
+	"archive/tar"
+	"bytes"
+	"sort"
+)
+
+// FileInfoHash provides an struct for accessing file hash
```
ilgooz:
```suggestion
// FileInfoHash provides a struct for accessing file hash
```

service/service.go
```diff
@@ -91,23 +92,41 @@ type DeployStatus struct {
...
 		return nil, err
 	}
 
+	// create stream to calculate hash
+	decompressedStream, err := archive.Tar(s.tempPath, archive.Uncompressed)
+	if err != nil {
+		return nil, err
+	}
+	envbytes := []byte(xos.EnvMapToString(env))
+	hash, err := tarsum.New(tar.NewReader(decompressedStream)).Sum(envbytes)
```
antho1404: Can't we use the reader that we have in parameter already, I don't see the point of having to tar it again except the `Uncompressed` but why this is necessary ? Can we have some details on that.

Also I think there is function for the hash that accepts a reader so why do we need to create this `tarsum` package ?

x/xos/env.go
```diff
@@ -22,10 +22,21 @@ func EnvMapToSlice(values map[string]string) []string {
...
 		env = append(env, k+"="+v)
 	}
-	sort.Strings(env)
+	sort.Stable(sort.StringSlice(env))
 	return env
 }
 
+// EnvMapToString transform a map of key value to a string in the form "key=value;key1=value1".
+// Env vars are sorted by names to get an accurate order while testing.
+func EnvMapToString(values map[string]string) string {
```
ilgooz: we can just use _EnvMapToSlice()_ inside this func, can be done in an another PR too

service/namespace.go
krhubert: It's just temporary before refactor. It will be gone.

service/service.go
krhubert: > Can't we use the reader that we have in parameter already,
No because the reader can be only read once. Here we save context to extract the mesg.yaml, then calculate the tarsum.

> Also I think there is function for the hash that accepts a reader so why do we need to create this `tarsum` package ?

Can you point one you are talking about?

service/namespace.go
krhubert: Hash datatype will be change to `[]byte` so even usage of `hex.DecodeString` will gone, here is temporary.

service/service.go
krhubert: Logic from save context was moved to `api/deploy.go` so only one function Untar remain.

antho1404:
For some reasons I cannot test with some data. Does anyone have a similar issue ? I have the `./dev-core` running of course ;)

```
➜  core git:(dev) ✗ ./dev-cli service deploy https://api.github.com/repos/mesg-foundation/service-webhook/tarball/master
Cannot reach the Core
Please start the core by running: mesg-core start
```

service/service.go
antho1404: https://github.com/mesg-foundation/core/pull/731/commits/c65fadc878c01e1f4680a0e86bfedcf3e148ec2f#diff-eeade751807b5e894d1f9a5b21390878

What's wrong with this implementation ?
This one plus the `mapEnvToString` looks valid for me why not go with that ? Am I missing something.

krhubert: The implementation is wrong:

- the hash won't be deterministic if the compression is different
- tar has [headers](https://golang.org/pkg/archive/tar/#Header) that may affect the hash calculation - for example, the mod/access/change times for a file (although the context of the fill will be the same)

antho1404: That's the part I don't really understand.

> the hash won't be deterministic if the compression is different

Agree but we control the compression when we deploy it and/or github have the same algorithm . Also the archive itself is a version. As long as this archive is the same for every computer it's fine (if we regenerate one and it's a different it's fine)

> tar has [headers](https://golang.org/pkg/archive/tar/#Header) that may affect the hash calculation

Again this is in the archive as long as we have the archive it will be the same except if when we untar it then it changes but this shouldn't be the case.

krhubert: > Also the archive itself is a version.
Hum? What version? the only `version` we're using is hash.

> . As long as this archive is the same for every computer it's fine

What do you mean by "the same for every computer"

> (if we regenerate one and it's different it's fine)

So if we regenerate archive and the content will be the same but the hash different then we don't have deterministic hash and this was the purpose of this PR, wasn't it?

> Again this is in the archive as long as we have the archive it will be the same except if when we untar it then it changes but this shouldn't be the case.

But you want to support multiple releases. Archive on github will be different, then one put on CI Artifacts then one put on ipfs. And there is also git repo which we tar ourselves. Someone might archive the same files on different os, machine, and the headers might be different.

api/deploy.go
```diff
@@ -86,44 +86,59 @@ func newServiceDeployer(api *API, env map[string]string, options ...DeployServic
...
+	if len(dirs) == 1 && dirs[0].IsDir() {
+		path = filepath.Join(path, dirs[0].Name())
+	}
+
+	d.sendStatus("Service downloaded with success", DonePositive)
+	r, err := archive.Tar(path, archive.Uncompressed)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer r.Close()
```
NicolasMahe: This is the only place where a ReadCloser from a tarball is close.
Either we should close ALL of them, either we should not close.
I guess we should close ALL of them?

Eg: `decompressedStream` in `service/service.go`

service/service.go
```diff
@@ -91,23 +92,41 @@ type DeployStatus struct {
...
+		return nil, err
+	}
+	defer os.RemoveAll(s.tempPath)
 	def, err := importer.From(s.tempPath)
 	if err != nil {
 		return nil, err
 	}
 
+	// create stream to calculate hash
+	decompressedStream, err := archive.Tar(s.tempPath, archive.Uncompressed)
```
NicolasMahe: `decompressedStream` should be close in a defer

service/service.go
antho1404: can you show me how they are different, for me from my previous tests all the version generated were the same: git, tar from github, tar from local source so if the generation is the same then the goal is to have a similar hash.
I just really don't see the extra work on this one when this technic works fine https://github.com/mesg-foundation/core/commit/c65fadc878c01e1f4680a0e86bfedcf3e148ec2f#diff-eeade751807b5e894d1f9a5b21390878. After maybe it works fine only on osx and in that case i'm fine but I would like to see and understand really why there is all this `tarsum` package that based on my tests is not necessary

krhubert: > can you show me how they are different, for me from my previous tests all the version generated were the same: git, tar from github, tar from local source so if the generation is the same then the goal is to have a similar hash.

There were too few tests and it happens to be the same. Here you have a showcase what I'm talking about: https://play.golang.org/p/a_rKlE0jk5t.

The content of the files are the same but the hash is different. Again there are tar headers and someone might build the same source multiple times on different os/machine/timezones etc. This is why we need tarsum to have exact hash.

antho1404: of course these are different, in your example you are setting a different modification time for each files
```go
ModTime: time.Date(i, 0, 0, 0, 0, 0, 0, time.UTC),
```
where `i` is either `0` or `1`.

If you force it to the same (for example `0`) you have a consistent hash `8c60defcbc9ff6ce751b290dcf938c8c538e9028`

And in a real life application if the tar changes because of a modification on date of a file or the permission or whatever metadata/header of a file then it's a new version and a it's normal not to have the same hash.

I would love the feedback from @NicolasMahe and @ilgooz on that but your example exactly shows that with the same environment the hash is the same and yes if we change the environment the hash changes but that's exactly the idea.

So in that case all the tarsum package is irrelevant.

api/deploy.go
antho1404: the service should not be aware of the service deployer. I'm totally fine to keep it that way.

NicolasMahe:
> For some reasons I cannot test with some data. Does anyone have a similar issue ? I have the `./dev-core` running of course ;)
> 
> ```
> ➜  core git:(dev) ✗ ./dev-cli service deploy https://api.github.com/repos/mesg-foundation/service-webhook/tarball/master
> Cannot reach the Core
> Please start the core by running: mesg-core start
> ```

The core get stuck at:
```
INFO[0000] Deploying service "Ethwallet" from "http://core-service-server/ethwallet.tar.gz" 
```

I didn't try to debug yet.

service/service.go
krhubert: Yes and this is a real situation. You could have multiple ci pipelines for the same build that for example regenerates some scripts/source code and change the mod/access time etc and in that case, the version stays the same. Thinking that there will be only one tar for the version is an assumption that may not be true in some cases.

ilgooz: The old idea behind of announcing hash sums for tar files are actually for security. This is the pattern we see always. When program makers publish a tarball via a 3rd party, they also announce the hash sum for that tarball so people can recalculate it from the downloaded file and verify the truth of untouched tarball.

In the normal sense, users can simply verify a tarball that way. But this is only possible by creating tarball once and distributing copies from that one. This is a very simplistic way of validation or can be considered as a limitation on requiring to create the tarball only once, depending on how you look at it.

In our case, we're trying to support git, url tarballs and ipfs together. Which means we end up 3 different tarballs generated. This way we have the flexibility of supporting tarballs generated from different sources or even with different algorithms but it's not simplistic anymore on the validation side. And we have env variable values that needs to change the result of the calculations which brings even more complexity. 

@antho1404 as you mentioned, either we should threat this 3 tarballs as different services because their tarball calculations are different or we need to create our own calculation strategy like @krhubert did with this PR. Or another option is might be to give devs to possibility to provide multiple tarball sources at the marketplace publish time. So we can download and calculate hashes and put them into service manifest as aliases of each other. But still it's complicated to have different hashes for the same service.

Also when we bring env vars in to the table while calculating the hashes, it still complicates stuff because we're creating a custom hash calculation strategy anyway. This might be ok. But getting rid of that is also simple. And I'm a bit on that side as well. It's also odd to create a new **static** service hash depending on the configuration values like env vars. Because the env var values are generally application specific and this new hash(service) generated from the env var values shouldn't be published on the marketplace anyway. So there is no actual reason to apply env variables on the deploy time, I think.

We can apply them on start time. The only difference here is that we need to pass the env vars each time while starting a service and this is actually a good thing. No hidden configuration going on under the hood.

As a conclusion, I'd like to apply env variables to services at the start time and only respect to service hash that is in the service manifest saved to ipfs during the marketplace publish. Other deploy urls for the same service different than the one in the manifest should not be considered as legit nor shouldn't be profitable for the dev as in tokens.

NicolasMahe:
> > For some reasons I cannot test with some data. Does anyone have a similar issue ? I have the `./dev-core` running of course ;)
> > ```
> > ➜  core git:(dev) ✗ ./dev-cli service deploy https://api.github.com/repos/mesg-foundation/service-webhook/tarball/master
> > Cannot reach the Core
> > Please start the core by running: mesg-core start
> > ```
> 
> I also have an issue. The `./dev-core` get stuck at:
> 
> ```
> ✔ Core stopped
> ✔ Core started
> INFO[0000] Deploying service "Ethwallet" from "http://core-service-server/ethwallet.tar.gz"
> ```
> I didn't try to debug yet.

I guess my internet was shitty or something.. it's now working.

service/service.go
NicolasMahe: I think we are trying to solve 2 different problems with one solution that is complicated for the first problem and not "enough" for the second:

- The first problem to resolve is about verifying that the downloaded source code of a service is the one that the developer published.
This is a checksum system that could be **only** available on service downloaded from the marketplace. In this case, the source code is uploaded as an archive created once and its checksum will be always the same because there is no decompression (so modification of file headers) or whatever that come into the calculation of the checksum.
As the archive is currently uploaded to IPFS, its checksum is already created and verified by IPFS libs, so we might not even need to verify it again in the core.
This was the first problem and why the implemented in this PR is actually too complicated.

- The second problem to resolve is to guarantee that the exact same software built from the same source is actually the same.
This current implementation is not enough because it's not verifying ALL the dependencies and the version of the system (eg ubuntu vs debian) and the compiler version (node vX vs node vY, go vX vs go vY, etc..) and many other things...
In order to resolve this problem, the docker image should be built on the computer of the developer and fully uploaded to the marketplace (like Docker Hub). This way, not only the source code but also ALL the OS and the dependencies will be exactly the same.
That's why I said this implementation is not "enough" for the second problem.

----

I think the hash of a service should be the hash of the docker image and not the hash of the source files. So actually, **forgetting completely** about a deterministic hash that could solve both problems.

ilgooz: I agree on that same service sources does not mean that a docker image built from that source from different environments will be the same. Because a docker image might be using _from x:latest_ tag or downloading other softwares during the build process while not explicitly fixing their versions.

With this awareness, if we really want the services to be truly same, we can start using the hash of Docker image instead of the service's source code as @NicolasMahe suggested and publish the image to ipfs.

But I'm not sure to go with this way because I'd like to give full control to devs about building their Docker images from the "true" source code. Because Docker images kind of black boxes and you'll not be easily know what is inside once it's built. Also, it is hard for devs to verify Docker image's hash if they want to for the same reason about unfixed versions.

So I think, we should give this responsibility to service authors about fixing their versions inside the Dockerfile and keep publishing service's tar file to ipfs and use its hash. And move env variables to start time and accept only one tar file url for the service from a single source.

krhubert: > I think the hash of a service should be the hash of the docker image and not the hash of the source files. So actually, forgetting completely about a deterministic hash that could solve both problems.

The docker hash image is not deterministic, we've checked this.

antho1404: > I think we are trying to solve 2 different problems with one solution that is complicated for the first problem and not "enough" for the second:

For me the problem is that we want for the same archive, the one from the marketplace to make sure that this will generate always the same hash.

I don't think we should even discuss your second problem yet. As for using the docker image as hash like we saw this is not enough. For me here the only stuff that is valid whatever the machine is doing the shaX sum of the sources (the tarball from the marketplace) I would not even think about the one from git or anything (even if like the initial commit shows, it's the same).

@NicolasMahe I suggest to close this PR if we are not clear on the problem then discuss the problem and then implement the solution. From my understanding it's having a same hash for a given tarball (that will not change).

NicolasMahe:
As decided on the [forum](https://forum.mesg.com/t/services-hash/261/5), let's use a simple hash calculated from a tar of the service + env var.
I think for this PR, the modification that needs the be done is remove the package tarsum to use a simple hash calculation on the created tar in `service.go`.

krhubert:
It's already done in #799

antho1404:
@krhubert #799 is really big, I would like to have closure on this first before doing another huge PR

api/deploy_test.go
```diff
@@ -33,19 +33,14 @@ func TestDeployService(t *testing.T) {
 		service, validationError, err := a.DeployService(archive, nil, DeployServiceStatusOption(statuses))
 		require.Nil(t, validationError)
 		require.NoError(t, err)
-		require.Len(t, service.Hash, 40)
+		require.Len(t, service.Hash, 64)
```
antho1404:
```suggestion
		require.Len(t, service.Hash, 40)
```

api/deploy_test.go
```diff
@@ -117,7 +101,7 @@ func TestDeployServiceFromURL(t *testing.T) {
 		service, validationError, err := a.DeployServiceFromURL(url, nil, DeployServiceStatusOption(statuses))
 		require.Nil(t, validationError)
 		require.NoError(t, err)
-		require.Len(t, service.Hash, 40)
+		require.Len(t, service.Hash, 64)
```
antho1404:
```suggestion
		require.Len(t, service.Hash, 40)
```

service/service.go
```diff
@@ -283,6 +258,21 @@ func (s *Service) configuration() *Dependency {
 	return nil
 }
 
+func serviceHash(contextDir string, env map[string]string) ([]byte, error) {
+	// create stream to calculate hash
+	stream, err := archive.Tar(contextDir, archive.Uncompressed)
```
antho1404: Is this really necessary ? We already have the tarball in the `New` we could simply 
```go
        h := sha256.New()
	if _, err := io.Copy(h, tarball); err != nil {
		return nil, err
	}
	envbytes := []byte(xos.EnvMapToString(env))
	s.Hash := hex.EncodeToString(h.Sum(envbytes))
```

in the `New` with no need of creating another tar

Dockerfile
```diff
@@ -13,9 +13,10 @@ RUN go build -o mesg-core \
       -ldflags="-X 'github.com/mesg-foundation/core/version.Version=$version'" \
       core/main.go
 
+# NOTE: docker uses xz-utils for uncompress tar.xz files
 FROM ubuntu:18.04
 RUN apt-get update && \
-      apt-get install -y --no-install-recommends ca-certificates=20180409 && \
+      apt-get install -y --no-install-recommends ca-certificates=20180409 xz-utils && \
```
antho1404: is this necessary if we remote the creation of the tarball in the `service.New` function ?

service/service.go
antho1404: resolved here https://forum.mesg.com/t/services-hash/261/6?u=anthony

Dockerfile
krhubert: yes, it is necessary to extract tar.xz (in service.New we don't create tar with xz compression)

antho1404:
Did some fixes here https://github.com/mesg-foundation/core/pull/804 @mesg-foundation/core can you have a look

NicolasMahe:
TODO:
- [ ] use another untar function that preserve file permission (MOD)
- [ ] calculate the hash based on the files:
  - [ ] only headers `Name`, `Linkname`, `Mode` of dir/files should be taken into account to calculate the hash
  - [ ]  filetype all (reg, link, symlink, char, block dir, fifo, socket)

service/service.go
```diff
@@ -91,23 +91,38 @@ type DeployStatus struct {
...
-	if err := s.saveContext(tarball); err != nil {
+
+	// untar tarball to retrieve mesg.yml
+	s.tempPath, err = ioutil.TempDir("", "mesg-")
+	if err != nil {
 		return nil, err
 	}
 	defer os.RemoveAll(s.tempPath)
 
+	// calculate the hash while untaring the archive
```
```suggestion
```

service/service.go
```diff
@@ -216,9 +193,10 @@ func (s *Service) deploy() error {
 	s.sendStatus("Image built with success", DDonePositive)
 
 	s.configuration().Image = imageHash
+	// TODO: the following test should be moved in New function
 	if s.Sid == "" {
 		// make sure that sid doesn't have the same length with id.
-		s.Sid = "_" + s.computeHash()
+		s.Sid = "_" + s.Hash
```
NicolasMahe: The generated Sid is not valid in this case...
Let's fix this in another PR.

NicolasMahe:
I cannot with GitHub because i am the creator on the PR but I still approve :)

service/service.go
NicolasMahe: https://github.com/mesg-foundation/core/issues/806

NicolasMahe:
Oh yeah 🍾 🍾 🍾 🍾

antho1404:
finished my tests and everything is fine :)

PR #732, opened 2019-01-24
Multiple argument for service start/stop commands

antho1404:
It is now possible to call the commands `service start` and `service stop` with multiple arguments and not only one.

```bash
mesg-core service stop sid1 sid2 sid3...
```

### DIFF
--- a/commands/service_start.go
+++ b/commands/service_start.go
@@ -20,24 +20,24 @@ func newServiceStartCmd(e ServiceExecutor) *serviceStartCmd {
 		Short:   "Start a service",
 		Long:    "Start a service previously published services",
 		Example: `mesg-core service start SERVICE`,
-		Args:    cobra.ExactArgs(1),
 		RunE:    c.runE,
 	})
 	return c
 }
 
 func (c *serviceStartCmd) runE(cmd *cobra.Command, args []string) error {
-	var (
-		serviceID = args[0]
-		err       error
-	)
-	pretty.Progress("Starting service...", func() {
-		err = c.e.ServiceStart(serviceID)
-	})
-	if err != nil {
-		return err
+	for _, serviceID := range args {
+		var err error
+		pretty.Progress("Starting service...", func() {
+			err = c.e.ServiceStart(serviceID)
+		})
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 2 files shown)


commands/service_list.go
```diff
@@ -30,21 +30,23 @@ To have more details, see the [detail command](mesg-core_service_detail.md).`,
 		Args:    cobra.NoArgs,
 		RunE:    c.runE,
 	})
+	c.cmd.Flags().BoolVarP(&c.quiet, "quiet", "q", c.quiet, "Display only the service hashes")
```
krhubert: I would change `quiet` to `--format` as this one is what we want.

Quiet is grate flag to be implemented globally and suppress all progress/messages that are not relevant for final output

antho1404: i will remove it from this PR, it needs more thoughts and it should have been two different PR anyway

antho1404:
I removed the `-q` on the list command, we need to discuss more about that before any implementation. This PR is only about multi arguments for start and stop commands

commands/service_start.go
```diff
@@ -19,25 +19,29 @@ func newServiceStartCmd(e ServiceExecutor) *serviceStartCmd {
...
-		return err
+	for _, arg := range args {
+		var err error
+		// build function to avoid using arg inside progress
+		fn := func(serviceID string) func() {
+			return func() {
+				err = c.e.ServiceStart(serviceID)
+			}
+		}(arg)
+		pretty.Progress("Starting service...", fn)
```
NicolasMahe:
```suggestion
		pretty.Progress(fmt.Sprintf("Starting service %q...", arg), fn)
```

commands/service_stop.go
```diff
@@ -18,21 +18,26 @@ func newServiceStopCmd(e ServiceExecutor) *serviceStopCmd {
...
-		return err
+	for _, arg := range args {
+		var err error
+		// build function to avoid using arg inside progress
+		fn := func(serviceID string) func() {
+			return func() {
+				err = c.e.ServiceStop(serviceID)
+			}
+		}(arg)
+		pretty.Progress("Stopping service...", fn)
```
NicolasMahe:
```suggestion
		pretty.Progress(fmt.Sprintf("Stopping service %q...", arg), fn)
```

PR #733, opened 2019-01-25
Create stream autoreconnect simple version

NicolasMahe:
Simplification of #717

- No acknowledgement system in the safe clients
- No choice of reconnect strategy. Always reconnect

- Add acknowledgement system in `client/service`

- FIX BUG where function that create stream didn't wait for it to be actually created resulting in pointer error with the acknowledgement system

- Improve naming. Follow gRPC way to name stuff

### DIFF
--- a/client/service/service.go
+++ b/client/service/service.go
@@ -142,7 +142,7 @@ func (s *Service) setupServiceClient() error {
 	if err != nil {
 		return err
 	}
-	s.client = serviceapi.NewServiceClientSafe(s.conn, serviceapi.OnError)
+	s.client = serviceapi.NewServiceClientSafe(s.conn)
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


krhubert:
> - No acknowledgement system in the safe clients

why not to put there? it would be good to have all grpc related function in one place
note: the ack system was removed from coreapi but not from tests

> - No choice of reconnect strategy. Always reconnect

Ok, but not everybody may want it

> - FIX BUG where function that create stream didn't wait for it to be actually created resulting in pointer error with the acknowledgement system

I left it intentionally and it's not a bug, it's a feature! (Btw you have just introduced bug :))
https://github.com/mesg-foundation/core/pull/733/files#diff-abeb2b7f0d83d05c7817daee216cf544R66

Putting the streamCreated here is wrong because you might get an error and the stream still will be nil
putting it after error check might end up with an endless loop if the client can't connect

So the solution for this is .... put back set stream ready in client_safe code, and in case of accessing Core_listenEvent you need a mutex to sync access by different go routines

PR #734, opened 2019-01-25
Use service id

krhubert:


krhubert:
I agree with @ilgooz too, but this PR is about being consistent so when there is variable service id use sid. The next prs will add hash to api, database etc ... But here is the first simple step

PR #735, opened 2019-01-27
Add env in hash calculation

antho1404:
Based on this comment https://github.com/mesg-foundation/core/pull/731#issuecomment-457146812

This is the support for the hashing of the service based on the tarball + env variable given during deployment


antho1404:
tests are failing, still need some investigation before merge

krhubert:
I'm on it, I don't know why there is an empty value in namespace

PR #737, opened 2019-01-28
Add info api

antho1404:
This API returns the basic configurations of the core and the list of services that are started by the core to let any client consume them

depends on https://github.com/mesg-foundation/core/pull/725

### DIFF
--- a/interface/grpc/server.go
+++ b/interface/grpc/server.go
@@ -53,6 +53,13 @@ func (s *Server) listen() (net.Listener, error) {
 		return nil, err
 	}
 
+	// keep-alive is need to prevent Docker network to think that gRPC connection is
+	// idle (see issues/549). without keep-alive it'll close the connection when there is
+	// no activity after some time. in that case grpc pkg will reconnect but all open
+	// gRPC streams needed to be re-established manually.
+	//
+	// keep-alive needs to be lower than Docker network's timeout that gives the
+	// decions about idle connections.
 	keepaliveOpt := grpc.KeepaliveParams(keepalive.ServerParameters{
 		Time: 1 * time.Minute,
 	})


ilgooz:
Why clients needs to know what services Core  uses?

antho1404:
The CLI needs to know that in order to call the services related to the marketplace. It's a way to expand the API. We could have "private" services maybe later if needed but now it's to expose all accessible "plugins" of the core. With this the CLI or UI could access all informations needed only with the services and without the need to create an api for every single feature.

protobuf/coreapi/api.proto
```diff
@@ -402,6 +405,21 @@ message ServiceLogsRequest {
...
+// Information concerning the Core from the `info` API.
+message InfoReply {
+  message CoreService {
+    string sid = 1;
+    string hash = 2;
+    string url = 3;
+  }
+  repeated CoreService services = 1;
+  string Address = 2;
+	string Image = 3;
```
ilgooz:
```suggestion
  string Image = 3;
```

config/config.go
```diff
@@ -56,7 +58,7 @@ type Config struct {
 	}
 
 	Service struct {
-		Foo ServiceConfig
+		// Foo ServiceConfig
```
krhubert: remove Foo (use comment with `NOTE` to describe the intention

config/config.go
```diff
@@ -138,9 +140,9 @@ func (c *Config) Validate() error {
 }
 
 // Services returns all services that the configuration package is aware of
-func (c *Config) Services() []ServiceConfig {
-	return []ServiceConfig{
-		c.Service.Foo,
+func (c *Config) Services() []*ServiceConfig {
+	return []*ServiceConfig{
+		// &c.Service.Foo,
```
krhubert: Same remove comment

interface/grpc/core/info.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"context"
+
+	"github.com/mesg-foundation/core/config"
+	"github.com/mesg-foundation/core/protobuf/coreapi"
+)
+
+// Info returns all necessary informations from the core.
+func (s *Server) Info(ctx context.Context, request *coreapi.InfoRequest) (*coreapi.InfoReply, error) {
+	c, err := config.Global()
```
krhubert: Do not use config.Global - pass the config or needed data

protobuf/coreapi/api.proto
```diff
@@ -42,6 +42,9 @@ service Core {
 
   // ServiceLogs gives a stream for dependency logs of a service.
   rpc ServiceLogs (ServiceLogsRequest) returns (stream LogData) {}
+
+  // Info returns all necessary informations from the core.
```
krhubert: Better comment (all necessary info but for what?)

s/Informations/Information/

protobuf/coreapi/api.proto
```diff
@@ -42,6 +42,9 @@ service Core {
...
+  rpc Info(InfoRequest) returns (InfoReply) {}
```
krhubert: s/Info/CoreInfo/ but i'm not sure about this

antho1404: I think Info is fine, we are accessing the API of the Core already

interface/grpc/core/info.go
antho1404: What would be your recommendations ?
- Give the config as dependency of `Server` (we need to change the `New` because we give the address and this address comes from the config)
- Delegate to the API and give the config to the API.
- Give a struct with all the info (needed by the Info api) to the server when we create it
- Keep it like that, `config.Global` is a singleton so let's use a singleton

I'm not really happy with any of them so I'm open for this but anyway this might be done in another PR.

protobuf/coreapi/api.proto
```diff
@@ -402,6 +405,21 @@ message ServiceLogsRequest {
...
 
+// The request to fetch all informations of the Core
+message InfoRequest {}
+
+// Information concerning the Core from the `info` API.
+message InfoReply {
+  message CoreService {
+    string sid = 1;
+    string hash = 2;
+    string url = 3;
```
NicolasMahe: Hash is useless because of sid.
The service key / type is missing.
The goal is for the cli to get the sid from a hardcoded type.
Eg: Give me SID of service type 'marketplace'.

protobuf/coreapi/api.proto
```diff
@@ -402,6 +405,21 @@ message ServiceLogsRequest {
...
+  }
+  repeated CoreService services = 1;
+  string Address = 2;
+  string Image = 3;
```
NicolasMahe: Replace image by version

protobuf/coreapi/api.proto
```diff
@@ -402,6 +405,21 @@ message ServiceLogsRequest {
...
+message InfoRequest {}
+
+// Information concerning the Core from the `info` API.
+message InfoReply {
+  message CoreService {
+    string sid = 1;
+    string hash = 2;
+    string url = 3;
+  }
+  repeated CoreService services = 1;
```
NicolasMahe: This could be a map like the Service in config.

protobuf/coreapi/api.proto
antho1404: Ok for the "key" but the `hash` is absolutely needed, the `sid` can be deleted from that but the `hash` is the version we want to use, not the sid, sid is just an alias on the latest version deployed.
Sid is useless because of hash, not the other way around

protobuf/coreapi/api.proto
antho1404: why ? ok to add another field version but this is the docker image, it's not the version. Even if there is a link between the version and the docker image let's not mix them.

Issue #549, opened 2018-10-23
interface/grpc/core: ListenTask gets canceled 


ilgooz:
This happened to me with a service that uses _go-service_. And several times more in back with some other services that I don't remember _(maybe webhook service)_. I'm not sure if I faced with this issue with _mesg-js_. So this might be relavent with _go-service_ or can be caused from something totally else. The problem occurs a while after _(<=40mins)_ starting the service which causes service to not react further task executions.

```
✔ Core stopped
✔ Core started
time="2018-10-23T09:18:11Z" level=info msg="deploying (2) system services..."
time="2018-10-23T09:18:12Z" level=info msg="'resolver' system service deployed"
time="2018-10-23T09:18:19Z" level=info msg="'workflow' system service deployed"
time="2018-10-23T09:18:19Z" level=info msg="starting system services..."
time="2018-10-23T09:18:20Z" level=info msg="all system services started"
time="2018-10-23T09:18:20Z" level=info msg="starting MESG Core version local"
time="2018-10-23T09:18:20Z" level=info msg="server listens on [::]:50052"
time="2018-10-23T09:18:27Z" level=info msg="finished unary call with code OK" grpc.code=OK grpc.method=CreateWorkflow grpc.service=api.Core grpc.start_time="2018-10-23T09:18:26Z" grpc.time_ms=1019.393 span.kind=server system=grpc
time="2018-10-23T09:18:27Z" level=info msg="finished unary call with code OK" grpc.code=OK grpc.method=SubmitResult grpc.request.deadline="2018-10-23T09:18:37Z" grpc.service=api.Service grpc.start_time="2018-10-23T09:18:27Z" grpc.time_ms=0.07 span.kind=server system=grpc
time="2018-10-23T09:18:36Z" level=info msg="finished unary call with code OK" grpc.code=OK grpc.method=SubmitResult grpc.request.deadline="2018-10-23T09:18:46Z" grpc.service=api.Service grpc.start_time="2018-10-23T09:18:36Z" grpc.time_ms=0.087 span.kind=server system=grpc
time="2018-10-23T09:18:36Z" level=info msg="finished unary call with code OK" grpc.code=OK grpc.method=CreateWorkflow grpc.service=api.Core grpc.start_time="2018-10-23T09:18:35Z" grpc.time_ms=1012.981 span.kind=server system=grpc
[...]

PR #738, opened 2019-01-28
Add 1 min keepalive in grpc

krhubert:
fixes https://github.com/mesg-foundation/core/issues/549


ilgooz:
my tests are also good, it's working!

Issue #741, opened 2019-01-28
Implement `context.Done` and `context.Err`


NicolasMahe:
Client should implement `context.Done` and `context.Err` to be notify when the stream is done and when an error occurred. The error should be return to parent function / or panic if previous not possible.

- [x] cli commands
- [ ] `client/service`

PR #747, opened 2019-01-29
Implement context in cli

NicolasMahe:
Closes https://github.com/mesg-foundation/core/issues/741

- [x] Still need to implement it for `client/service`.


NicolasMahe:
@ilgooz @antho1404 This PR is ready for review and merge.

commands/provider/service_provider.go
```diff
@@ -94,6 +94,10 @@ func (p *ServiceProvider) ServiceListenEvents(id, eventFilter string) (chan *cor
 	resultC := make(chan *coreapi.EventData)
 	errC := make(chan error)
 
+	go func() {
+		<-stream.Context().Done()
+		errC <- stream.Context().Err()
```
antho1404: What if there is an error in the context but it doesn't close ? just want to make sure that it's not a problem here.

What about something like that ?

```go
go func() {
  select {
    case <- stream.Context().Done()
      errC <- fmt.Errorf("context closed")
    case err := <- stream.Context().Error()
      errC <- err
  }
}
```

NicolasMahe: `stream.Context().Error()` is not a channel.
The way is done is the right way to do.

antho1404: sorry yes my bad... that's friday night that's why...

NicolasMahe:
> I think, we don't need this PR since _Recv()_ will already stop blocking and return an error when stream is closed by the server or is got broken. Is there any other benefits of reading errors from context that I might be missing?

`Recv()` doesn't return error when the connection is broken.

All the network function are using `context`. The core and cli should implement it the right way.

PR #749, opened 2019-01-29
Autodeploy service for dev

antho1404:
With the new services deployed from tar and the new kind of system services that are just services https://github.com/mesg-foundation/core/pull/725 we need a way to easily work and deploy local services locally and for production a way to configure the endpoint of the services we will want for the release.

For that few things:
- **dev and prod** Configurable variable in the config package with the URL of the services with the format `ServiceURL`
- **dev** A server that contains all the services archives from the systemservices directory
- **dev** A script `dev-core` that inject the services tar endpoint from the server created

This way we now have a way to configure the address of the services needed by the core during the compilation (and can be integrated in the CI) and a way in development to work with the local directory without any hack on the code base, just on the dev tools.


NicolasMahe:
The function `deleteServiceServer` in `dev-core` script doesn't work if there is no service `core-service-server`.

I add some output to see what happen:
```bash
function deleteServiceServer {
  echo "will delete a"
  rm -rf tmp-systemservices
  echo "will delete b"
  serviceExists=$(docker service list | grep core-service-server)
  if [[ "$serviceExists" != "" ]]; then
    echo "Deleting service server"
    docker service rm $MESG_SERVICE_SERVER
    echo "Finish deleting service server"
  fi
  echo "Finish finish service server"
}
```
```
➜  core git:(feature/autodeploy-service-for-dev) ✗ ./dev-core
will delete a
will delete b
```
The I think the `if` return an error somehow..

dev-core
```diff
@@ -8,6 +8,36 @@ MESG_LOG_FORCECOLORS=${MESG_LOG_FORCECOLORS:-false}
 VERSION=local
 LDFLAGS="-X 'github.com/mesg-foundation/core/version.Version=$VERSION'"
 CLIFLAGS=""
+MESG_SERVICE_SERVER="core-service-server"
+
+function deleteServiceServer {
+  rm -rf tmp-systemservices
+  serviceExists=$(docker service list | grep core-service-server)
+  if [[ "$serviceExists" != "" ]]; then
```
NicolasMahe:
```suggestion
  if docker service list | grep -q core-service-server; then
```

dev-core
```diff
@@ -8,6 +8,36 @@ MESG_LOG_FORCECOLORS=${MESG_LOG_FORCECOLORS:-false}
 VERSION=local
 LDFLAGS="-X 'github.com/mesg-foundation/core/version.Version=$VERSION'"
 CLIFLAGS=""
+MESG_SERVICE_SERVER="core-service-server"
+
+function deleteServiceServer {
+  rm -rf tmp-systemservices
+  serviceExists=$(docker service list | grep core-service-server)
```
NicolasMahe:
```suggestion
```

dev-core
NicolasMahe: see https://github.com/mesg-foundation/core/pull/749#issuecomment-459661589

antho1404:
Can you please review @krhubert @ilgooz

config/services.go
```diff
@@ -0,0 +1,19 @@
...
+// Use the following format for the variable name: "[Service]URL" (where Service is the name of the service capitalized)
+// The service name should be the name of the directory inside `systemservices` but capitalized
+// example:
+// 		var (
+// 			FooURL string
+// 			BarURL string
+// 		)
+var ()
+
+func (c *Config) initServices() {
```
krhubert: What is the purpose of initServices?

antho1404: It's called from the config, it's to have all the logic about the services in the same folder instead of having to initialize it on the config.go but declare it here

PR #750, opened 2019-01-29
Parse float number to big.Float

### LINKED ISSUE #639: Parsing Number in mesg.yaml
Parse is done into `float64` and it should be into `big.Float`

krhubert:
Close #639

### DIFF
--- a/service/validation.go
+++ b/service/validation.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"fmt"
+	"math/big"
 )
 
 // ParameterWarning contains a specific warning related to a parameter.
@@ -71,13 +72,15 @@ func (v *parameterValidator) validateType(value interface{}) []*ParameterWarning
 		}
 
 	case "Number":
-		_, okFloat64 := value.(float64)
-		_, okFloat32 := value.(float32)
-		_, okInt := value.(int)
-		if !okInt && !okFloat64 && !okFloat32 {
+		switch value.(type) {
+		case
+			uint8, uint16, uint32, uint64, uint,
+			int8, int16, int32, int64, int,
+			float64, float32,
+			*big.Int, *big.Float:
+		default:
 			return []*ParameterWarning{v.newParameterWarning("not a number")}
 		}
-
 	case "Boolean":
 		if _, ok := value.(bool); !ok {
 			return []*ParameterWarning{v.newParameterWarning("not a boolean")}
--- a/utils/servicecasting/cast.go
+++ b/utils/servicecasting/cast.go
@@ -3,6 +3,7 @@ package casting
 import (
 	"encoding/json"
 	"fmt"
+	"math/big"
 	"strconv"
 	"strings"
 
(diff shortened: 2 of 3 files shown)


service/validation.go
```diff
@@ -71,13 +72,15 @@ func (v *parameterValidator) validateType(value interface{}) []*ParameterWarning
...
-		_, okFloat64 := value.(float64)
-		_, okFloat32 := value.(float32)
-		_, okInt := value.(int)
-		if !okInt && !okFloat64 && !okFloat32 {
+		switch value.(type) {
+		case
+			uint8, uint16, uint32, uint64, uint,
+			int8, int16, int32, int64, int,
+			float64, float32,
+			*big.Int, *big.Float:
```
NicolasMahe: Why having `*big.Int` and `*big.Float` here?
I'm not familiar with those type.
Can you provide a reason?

NicolasMahe: I don't see the point of being compatible with this when most JSON parser / encoder don't support number bigger than 64bits.

krhubert: > most JSON parser / encoder don't support number bigger than 64bits.

python:
```python
>>> json.loads('115792089237316195423570985008687907853269984665640564039457584007913129639936')
115792089237316195423570985008687907853269984665640564039457584007913129639936L
```

node:
```
> JSON.parse('115792089237316195423570985008687907853269984665640564039457584007913129639936')
1.157920892373162e+77
```

go support this with big, and probably every other language. 

We need to support all numbers if we are only using number type.

NicolasMahe: > node:
> 
> ```
> > JSON.parse('115792089237316195423570985008687907853269984665640564039457584007913129639936')
> 1.157920892373162e+77
> ```
> go support this with big, and probably every other language.

Here the number is not the same. The precision has been lost.
> JavaScript numbers are always stored as double precision floating point numbers, following the international IEEE 754 standard. 
> Integers (numbers without a period or exponent notation) are accurate up to 15 digits.
> https://www.w3schools.com/js/js_numbers.asp

krhubert: I know, but it was just to show that other languages support parsing big numbers (even if the precision is lost, but this is something nodejs developer needs to worries about, we need to guarantee the go code handles data input well.)

NicolasMahe:
```json
{
  "number": 115792089237316195423570985008687907853269984665640564039457584007913129639936,
  "string": "115792089237316195423570985008687907853269984665640564039457584007913129639936"
}
```
```js
const BigNumber = require('bignumber.js')
const test = require('./test.json')
console.log('number', new BigNumber(test.number).toString())
console.log('string', new BigNumber(test.string).toString())
```
```
➜  test node index.js
number 1.157920892373162e+77
string 1.15792089237316195423570985008687907853269984665640564039457584007913129639936e+77
```

JS cannot parse the number even using BigNumber.

krhubert:
I guess it can (http://derpturkey.com/untitled/) but you have used `require test.json` and it automatically convers number to javascript number which has less precision

But, even if javascript can't, go can (and other languages) and someone may want to send big numbers in go (and for javascript nothing changes with it)

NicolasMahe:
The goal is to have a seamless communication (so types) between any language.
Unfortunately we have to be compatible with the less capable language of the mains ones and JS is the most important language to be compatible with.

PR #751, opened 2019-01-29
Fix/docker mock

### LINKED ISSUE #701: Fix build mocks for docker client
With the removal of vendor, the mocks for docker client cannot be build like it was because the script was looking into vendor to get the interface:

```
mockery -name CommonAPIClient -dir ./vendor/github.com/docker/docker/client -output ./utils/docker/mocks
```

I see 3 solutions:
- Easy: Create a file that import the docker client interface and export it
- Complicated: Find the right version of the client in `go/pkg/dev`
- Ugly: execute `go mod vendor` for the mocks...

I recommend to go with solution 1.

File to fix:
https://github.com/mesg-foundation/core/blob/94e58541ca3d6ac4baf1f2cd8c191c2fce8548f1/scripts/build-mocks.sh

krhubert:
close #701


mocks.go
```diff
@@ -0,0 +1,11 @@
+package main
```
ilgooz: it makes the most sense to move this to _./utils/docker/_ path

krhubert: We could, but I took a different approach - like with [tools.go](https://github.com/mesg-foundation/core/blob/master/tools.go) - put every interface (from vendor) that we want to mock in one palce and not spread over different directories. Maybe it should be good to add comment why we have it here.

NicolasMahe:
```suggestion
package mocks
```

Why not naming it mocks in this case? Like tools?

PR #748, opened 2019-01-29
Add info api

antho1404:
This API returns the basic configurations of the core and the list of services that are started by the core to let any client consume them

Same as https://github.com/mesg-foundation/core/pull/737
This one is based on https://github.com/mesg-foundation/core/pull/743

PR #755, opened 2019-01-30
Feature/ipfs deployment

antho1404:
Add deployment command that deploy a service directory on IPFS:

Workflow:
- Create a tar of the folder
- Upload this tar on IPFS
- Create a definition file that includes IPFS link of the tar, service definition, readme, version
- Upload a json version of this definition file on IPFS

TODO in other PR:
- [ ] Connect to the ethereum wallet service to get the address of the developer (depends on https://github.com/mesg-foundation/core/pull/728)
- [ ] Connect to the marketplace service to send this definition file hash on the marketplace contract
- [ ] Add support of .mesgignore


ipfs/ipfs.go
```diff
@@ -0,0 +1,72 @@
...
+type Response struct {
+	Name string
+	Hash string
+	Size string
+}
+
+// New creates a new IPFS client method
+func New() *IPFS {
+	return &IPFS{
+		client:   &http.Client{},
```
ilgooz: it's good to add a timeout to http client, by default it's forever

ipfs/ipfs.go
```diff
@@ -0,0 +1,72 @@
...
+		endpoint: "https://ipfs.infura.io:5001/",
```
ilgooz: it's be good to move this to _config_ pkg

ipfs/ipfs.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+// Add data from a reader to the IPFS node
+func (ipfs *IPFS) Add(name string, reader io.Reader) (*Response, error) {
+	var b bytes.Buffer
```
ilgooz: this will memory a lot. instead we can just use [http.Post()](https://golang.org/pkg/net/http/#Client.Post). since it accepts an _io.Reader_, we can directly pass this _reader_ var to it. for the _content-type_, it'll be the same for every upload, so we can use a constant value

ipfs/ipfs.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	req.Header.Set("Content-Type", w.FormDataContentType())
+	res, err := ipfs.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("bad status: %s", res.Status)
+	}
+	body, err := ioutil.ReadAll(res.Body)
```
ilgooz: we can directly use _json.NewEncoder(body).Encode(&response)_

commands/service_publish.go
```diff
@@ -0,0 +1,105 @@
...
+	ipfs *ipfs.IPFS
+
+	e ServiceExecutor
+}
+
+type marketplaceData struct {
+	Version int `json:"version,omitempty"`
+	Service struct {
+		Deployment struct {
+			Source string `json:"source,omitempty"`
```
ilgooz: i think, we can reduce _omitempty_'s a bit. it doesn't have any effect where a field is always set(required), for example in this case(service, deployment and source fields), definition and version(?) fields.

commands/service_publish.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (c *servicePublishCmd) preRunE(cmd *cobra.Command, args []string) error {
+	c.path = getFirstOrDefault(args)
+	c.ipfs = ipfs.New()
+	return nil
+}
+
+func (c *servicePublishCmd) runE(cmd *cobra.Command, args []string) error {
+	tar, err := archive.TarWithOptions(c.path, &archive.TarOptions{
```
ilgooz: got to respect to _.dockerignore_

ipfs/ipfs.go
antho1404: This package should not be related to MESG, introducing the config here will strongly coupled this lib. Also this is a package only used in the CLI (that should not access to the config, if it needs to it should contact the Info api https://github.com/mesg-foundation/core/pull/748)

commands/service_publish.go
```diff
@@ -0,0 +1,105 @@
...
+	return c
+}
+
+func (c *servicePublishCmd) preRunE(cmd *cobra.Command, args []string) error {
+	c.path = getFirstOrDefault(args)
+	c.ipfs = ipfs.New()
+	return nil
+}
+
+func (c *servicePublishCmd) runE(cmd *cobra.Command, args []string) error {
```
krhubert: This whole function should be moved away from here. Commands pkg is only for input/output, so this should be rather in `commands/provider`

ipfs/ipfs.go
ilgooz: then it makes sense to move it under _commands_ pkg?

ipfs/ipfs.go
```diff
@@ -0,0 +1,72 @@
+package ipfs
```
krhubert: Is there a reason why you haven't used already written client like https://github.com/ipfs/go-ipfs-api but you go with own implementation? Does this client have some limitation that stops us from using it?

commands/service_publish.go
antho1404: This is the source that are deployed and not the source that are built.
Dockerignore is really to ignore when we deploy in docker and it's not what we are doing. We talked with nico about a `.mesgignore` that will be to ignore some files for deployment. This can be supported but I don't think dockerignore should be supported here.

commands/service_publish.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("https://gateway.ipfs.io/ipfs/" + definitionResponse.Hash)
+
+	return nil
+}
+
+func (c *servicePublishCmd) readReadme() string {
```
krhubert: Also, this function should be in a separate package, could be even something simple like `utils/readme`.

ipfs/ipfs.go
antho1404: no this is an independent package, it shouldn't be linked to the commands, i would be agree maybe to have it int the `x` packages but inside command no, the goal is to have it independent from the rest.

ipfs/ipfs.go
```diff
@@ -60,12 +59,8 @@ func (ipfs *IPFS) Add(name string, reader io.Reader) (*Response, error) {
 	if res.StatusCode != http.StatusOK {
 		return nil, fmt.Errorf("bad status: %s", res.Status)
 	}
-	body, err := ioutil.ReadAll(res.Body)
-	if err != nil {
-		return nil, err
-	}
 	var response Response
-	if err := json.Unmarshal(body, &response); err != nil {
+	if err := json.NewDecoder(res.Body).Decode(&response); err != nil {
```
ilgooz: no need for _if_ we can return this error directly

ilgooz: somethin like this:

```go
var response Response
err := json.NewDecoder(res.Body).Decode(&response)
return &response, err
```

commands/service_publish.go
antho1404: but this command for now has no need of the api and any core client so it doesn't make sense yet to move it into the serviceprovider for me

ipfs/ipfs.go
antho1404: But then between `err := json.NewDecoder(res.Body).Decode(&response)` and `return &response, err` if we add an instruction we might miss an error. I like to immediately check the error or return all the result directly

ipfs/ipfs.go
krhubert: hum ok, ill check this

commands/service_publish.go
NicolasMahe: Can you add support for the `.mesgignore`?

commands/service_publish.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func (c *servicePublishCmd) createDefinitionFile(tarballHash string) ([]byte, error) {
+	definition, err := importer.From(c.path)
+	if err != nil {
+		return nil, err
+	}
+	var data marketplaceData
+	data.Version = PublishVersion
+	data.Service.Deployment.Source = tarballHash
```
NicolasMahe: What is the value of `tarballHash`?
Just the hash? Does it contains an `ipfs://` prefix? Is it an http url?

commands/service_publish.go
antho1404: i wanted to do that in another pr this one has already too much discussion anyway

ipfs/ipfs.go
antho1404: this can be an improvement later

commands/service_publish.go
antho1404: just the hash

NicolasMahe: I think it should be an URL with the ipfs prefix in this case.
We don't want to have to guess that's a ipfs hash, we want to be sure.
And like this the marketplace can be compatible with any file storage network.

antho1404: I didn't find any information about an IPFS protocol. We could consider this as a protocol but i  don't like the idea of supporting something that doesn't exists.

We could have a deployment type attribute or put the full address of gateway ipfs, eg: https://gateway.ipfs.io/ipfs/xxxx

commands/service_publish.go
krhubert: It's not about API or Core client.

The package commands shouldn't do any operation expect input/output (doesn't matter if it need contact via api, http or any other service like ipfs).

This is why providers are, to avoid putting logic inside commands package (or to put as less logic as possible)

antho1404: done

antho1404:
This command needs to connect to the ethereum wallet and sign 2 transactions, I think we will have enough confirmations with that :)
I think `public: true` is too much, and for the confirmation, I agree but I prefer to see that after the whole command completed with signing and broadcasting of the transaction to see what makes sense.

commands/service_publish.go
NicolasMahe: Yeah right....
The best "official" answer:
> IPFS data is designed to be linked in canonical path (without schema) but support for it is not yet in browsers. In case of storage and internal representation I would use this format: `/ipfs/QmburSK2Cjav26X6oGsiXwwwDXnVUUjm9i65UErRD99Z9v`
> From https://github.com/ipfs/ipfs/issues/181

We need to have something that say that's an IPFS hash. Otherwise we are stuck with IPFS for link on the marketplace..

antho1404: or we can anyway add a type for this, in this case it will be type ipfs and we could have type url, type git... instead of having to do a detection on the url we can explicitly say it

antho1404: deployment type added

antho1404:
@NicolasMahe @krhubert @ilgooz can you review please :)

krhubert:
If you ever want to add tests for such clients (http) you can take a look at https://github.com/alphasoc/nfr/tree/master/client - there is a client with 100% test code coverage.

PR #757, opened 2019-01-30
Read dockerignore in dev and deploy commands

### LINKED ISSUE #736: Support dockerignore in command deploy
The `.dockerignore` file is not supported in the command deploy.

But it is supported in `container/build.go`.

Let's also check if other command / core are supporting dockerignore.

Also, all codebase should use `xarchive.GzippedTar` instead of `archive.TarWithOptions`.
`dockerignoreFiles` function should also be move the `xarchive`.

krhubert:
close #736


krhubert:
But what is the resaon behind it? Why do you want just subset of this funcionalities?

If we want to reduce complexity then we already did. We have used functions that are already writter and working. If we want to copy them, then we need to write, test, think what part cut off etc... For example why do you want to not handle symlink (or any other funcinality)?

I think we must be as most compatible with `docker build` as we can. Because if you are developer and you run `docker build` and `mesg service deploy` you expect the same results.

And about [dockerignoreFiles](https://github.com/mesg-foundation/core/blob/bfeb67d/container/build.go#L48) I wrote it before I knew it that there is already such function. What we should do is replace it with the one from `build` package to avoid dedup function.

commands/provider/service_deployer.go
```diff
@@ -73,8 +75,27 @@ func (p *ServiceProvider) ServiceDeploy(path string, env map[string]string, stat
...
+
+	if err := build.ValidateContextDirectory(contextDir, excludes); err != nil {
+		return fmt.Errorf("error checking context: %s", err)
+	}
+
+	// And canonicalize dockerfile name to a platform-independent one
+	relDockerfile = archive.CanonicalTarNameForPath(relDockerfile)
+	excludes = build.TrimBuildFilesFromExcludes(excludes, relDockerfile, false)
+
 	archive, err := archive.TarWithOptions(path, &archive.TarOptions{
```
NicolasMahe: What about `xarchive` package?

Should we refacto most of the content of `deployServiceSendServiceContext` or simply delete `xarchive` package?

krhubert: Some other packages like service (or container) can use it, but it needs refactoring first. And here is small PR to add this to dev/deploy command

NicolasMahe:
> Why putting too much logic on the cli. A lot of these validations are great but would make more sense on the core itself.
> 
> I would go with something simple and maybe move the validation on the backend
> 
> ```go
> excludes, err := build.ReadDockerignore(path)
> archive, err := xarchive.GzippedTar(path, excludes)
> ```

I agree on at least try to refactor this function. But I also agree to merge this one now if @krhubert works directly on the refactor version.

krhubert:
Those functions won't work on server side and all **must** be used on cli where the directory exists.

`build.ValidateContextDirectory(contextDir, excludes)`
ValidateContextDirectory checks if all the contents of the directory can be read and returns an error if some files can't be read symlinks which point to non-existing files don't trigger an error

`archive.CanonicalTarNameForPath`
It fixes the filename for diffrent os (for now we don't support windows but when we will already start with bug)

`build.TrimBuildFilesFromExcludes`
TrimBuildFilesFromExcludes removes the named Dockerfile and .dockerignore from the list of excluded files.  (otherwise Dockerfile and .dockerignore won't be send to server)

antho1404:
I don't understand why they will not work on the server side, we are getting the tar and extracting the content so we will have the same informations that the cli has.

But anyway, same this can be done in another PR, this one can be merged

NicolasMahe:
> I don't understand why they will not work on the server side, we are getting the tar and extracting the content so we will have the same informations that the cli has.
> 
> But anyway, same this can be done in another PR, this one can be merged

@antho1404 can you approve in this case

PR #762, opened 2019-01-31
Fix improve guide start in readme

NicolasMahe:


NicolasMahe:
> I'm ok to have some verbosity but this is a quick start and should be quick, not perfect, I would remove the handling of errors here, on the document we need it, on the quick start not really we don't want to show a huge code that they might be afraid of.
> 
> I feel this code is 3 times bigger, that's not attracting

It's bigger not because of error handling but because of code style.

The only error added is on the return of the executeTaskAndWaitResult func:
```js
      if (result.outputKey !== 'success') {
        const message = JSON.parse(result.outputData).message
        console.error('an error occurred while sending the invitation: ', message)
        return
      }
```

I think most developer don't check much the actual documentation, so having verbosity here will help them to understand what's going on.

README.md
```diff
@@ -91,45 +93,73 @@ npm init && npm install --save mesg-js
...
+Now let's give this super simple application a try.
 
 Let's trigger the webhook with the following command:
 
 ```bash
 curl -XPOST http://localhost:3000/webhook
 ```
 
-:tada: You should have received an email in your inbox with your precious invitation to our Discord.
+:tada: You should have received an email in your inbox with your invitation to our Discord. Come join to our community.
```
NicolasMahe:
```suggestion
:tada: You should have received an email in your inbox with your invitation to our Discord. Come join our community.
```

PR #764, opened 2019-01-31
Do not remove previous sid from database

### LINKED ISSUE #718: Cannot start any services
I run a service that is present in my services list but this service cannot start because of an error when trying to connect to the core.

I just updated with the v0.7, I had this service running and so the core stopped during the update and start again and now all my previously running services cannot run anymore, even the one that were not running before.

I don't know if this is related to the v0.7 or just because the restart of the core was maybe brutal.

screen shot 2019-01-16 at 10 38 48 pm

krhubert:
close #718

### DIFF
--- a/database/service_db.go
+++ b/database/service_db.go
@@ -178,12 +178,6 @@ func (d *LevelDBServiceDB) Save(s *service.Service) error {
 		return err
 	}
 
-	// delete existent service that has the same sid.
-	if err := d.delete(tx, s.Sid); err != nil && !IsErrNotFound(err) {
-		tx.Discard()
-		return err
-	}
-
 	// encode service
 	b, err := d.marshal(s)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


antho1404:
I think there is a problem here, it's good that we keep the hash so we can always access but we are loosing the information that this hash is linked to a sid.

Maybe that will be done in another PR but we should have an array of hashes for the sid

Something like that (of course with error management ;))
```go
tx.Put([]byte(sidKeyPrefix+s.Sid), []byte(s.Hash), nil)
```
into
```go
tx.Put([]byte(sidKeyPrefix+s.Sid), append(tx.Get(s.Sid), []byte(s.Hash)), nil)
```

krhubert:
> I think there is a problem here, it's good that we keep the hash so we can always access but we are loosing the information that this hash is linked to a sid.

In that case, we don't lose it because we have it in the `service` struct, what we lose here is the order of added hashes (but we never had :)) so this will be done in next PR.

Also, I was thinking (we should have it i guess) about adding for example `*` in command `service list` to show the last sid.

PR #766, opened 2019-02-02
commands>service>publish: confirm before publishing to ipfs

ilgooz:


commands/service_publish.go
```diff
@@ -56,6 +60,18 @@ func (c *servicePublishCmd) preRunE(cmd *cobra.Command, args []string) error {
 }
 
 func (c *servicePublishCmd) runE(cmd *cobra.Command, args []string) error {
+	if !c.yes {
+		var confirmed bool
+		if err := survey.AskOne(&survey.Confirm{
+			Message: fmt.Sprintf("Are you sure to publish path? (%s)", c.path),
```
ilgooz:
```suggestion
			Message: fmt.Sprintf("Are you sure to publish from path? (%s)", c.path),
```

antho1404:
I would recommend to wait the full feature finished before thinking about ux, for now this features needs a lot of additional work

ilgooz:
> I would recommend to wait the full feature finished before thinking about ux, for now this features needs a lot of additional work

to me this is not for ux. it is to secure personal files against uploading them to ipfs by mistake. for example, even us can give a wrong path to publish cmd like home folders, documents folder etc. and they'll be published to ipfs which we cannot take it back

NicolasMahe:
> > I would recommend to wait the full feature finished before thinking about ux, for now this features needs a lot of additional work
> 
> to me this is not for ux. it is to secure personal files against uploading them to ipfs by mistake. for example, even us can give a wrong path to publish cmd like home folders, documents folder etc. and they'll be published to ipfs which we cannot take it back

I agree with Anthony.

The current implementation of the publish function doesn't integrate the marketplace on ethereum. A lot's of confirmation will be required already..
Let's wait for marketplace integration ;)

antho1404:
this command anyway for now is in hidden so people will not do that mistake if we never document this command

ilgooz:
I'm closing this PR for now

ilgooz:
related to https://github.com/mesg-foundation/core/pull/755#pullrequestreview-199345577

PR #769, opened 2019-02-03
Improve and simplify a test in client service 

NicolasMahe:
I simplify the test `TestClose` because CircleCI had a lots of problem to run it.. It worked on my computer but not most of the time on CircleCi:
https://circleci.com/gh/mesg-foundation/core/tree/feature%2Fclient-context

### DIFF
--- a/client/service/examples/logger/logger/logger_test.go
+++ b/client/service/examples/logger/logger/logger_test.go
@@ -60,7 +60,7 @@ func TestListenError(t *testing.T) {
 
 	var resp errorResponse
 	require.Nil(t, execution.Data(&resp))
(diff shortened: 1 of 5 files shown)


client/service/examples/logger/logger/logger_test.go
```diff
@@ -71,21 +70,12 @@ func TestClose(t *testing.T) {
 	logger := New(s)
 
 	go server.Start()
-
-	var wg sync.WaitGroup
-
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		assert.Nil(t, logger.Start())
```
ilgooz: I think this check should stay because we're testing if _Start()_ will return an err or not

Issue #760, opened 2019-01-30
grpc goaway


ilgooz:
happened after ~2 days run of marketplace service on DO droplet

```
root@marketplace-service:~# docker service logs core-5ecd2e15230cff43bb7529942c221505605f9e52-service
core-5ecd2e15230cff43bb7529942c221505605f9e52-service.1.k9sjv9cpmjop@marketplace-service    | 
core-5ecd2e15230cff43bb7529942c221505605f9e52-service.1.k9sjv9cpmjop@marketplace-service    | > service-ethereum@1.0.0 start /app
core-5ecd2e15230cff43bb7529942c221505605f9e52-service.1.k9sjv9cpmjop@marketplace-service    | > ts-node src/index.ts
core-5ecd2e15230cff43bb7529942c221505605f9e52-service.1.k9sjv9cpmjop@marketplace-service    | 
core-5ecd2e15230cff43bb7529942c221505605f9e52-service.1.ym2qexlzf1hz@marketplace-service    | 
core-5ecd2e15230cff43bb7529942c221505605f9e52-service.1.ym2qexlzf1hz@marketplace-service    | > service-ethereum@1.0.0 start /app
core-5ecd2e15230cff43bb7529942c221505605f9e52-service.1.ym2qexlzf1hz@marketplace-service    | > ts-node src/index.ts
core-5ecd2e15230cff43bb7529942c221505605f9e52-service.1.ym2qexlzf1hz@marketplace-service    | 
core-5ecd2e15230cff43bb7529942c221505605f9e52-service.1.ym2qexlzf1hz@marketplace-service    | catch listenTask { Error: 13 INTERNAL: GOAWAY received
core-5ecd2e15230cff43bb7529942c221505605f9e52-service.1.ym2qexlzf1hz@marketplace-service    |     at Object.exports.createStatusError (/app/node_modules/grpc/src/common.js:91:15)
core-5ecd2e15230cff43bb7529942c221505605f9e52-service.1.ym2qexlzf1hz@marketplace-service    |     at ClientReadableStream._emitStatusIfDone (/app/node_modules/grpc/src/client.js:233:26)
core-5ecd2e15230cff43bb7529942c221505605f9e52-service.1.ym2qexlzf1hz@marketplace-service    |     at ClientReadableStream._receiveStatus (/app/node_modules/grpc/src/client.js:211:8)
[...]

Issue #759, opened 2019-01-30
Transport is closing after ~10mn


antho1404:
This might be related to the keep alive development.
After few minutes (~10mn) of a service inactivity I get the following error.

<img width="604" alt="screen shot 2019-01-30 at 10 24 59 pm" src="https://user-images.githubusercontent.com/1783126/51991552-e8959400-24dd-11e9-9333-750d117b58f9.png">

It seems ok when the service is emitting events.

NicolasMahe:
No problem on my side with services that are running in background with command `service start` after 30min.

NicolasMahe:
@antho1404 did you experience a similar issue? Did the demo experiences it again?

antho1404:
Can still reproduce after 10mn, this is not happening on the demo but during development.

Workflow:
- git checkout dev (commit cf19a76cc9f4912f0eca37d6f32010ebf0da3ec8)
- ./dev-core (this runs the ethwallet, please ignore)
- update ethwallet with another sid (ethwallet2)
- ./dev-cli service dev systemservices/ethwallet
- set timer to 10mn
- come back after 10mn
<img width="1671" alt="screen shot 2019-02-04 at 9 49 59 pm" src="https://user-images.githubusercontent.com/1783126/52215553-da24ef00-28c6-11e9-9e52-c1bcf65bf29f.png">

NicolasMahe:
PR https://github.com/mesg-foundation/core/pull/774 fix this issue.

PR #770, opened 2019-02-04
Update keep alive of client to 5min to prevent spamming the server

NicolasMahe:
Related to https://github.com/mesg-foundation/core/issues/760 and https://github.com/mesg-foundation/core/issues/759

Client should not ping the server more often than the `EnforcementPolicy.MinTime` set on the server. By default it's 5min, so I set the client to ping the server every 5min.

### DIFF
--- a/client/service/service.go
+++ b/client/service/service.go
@@ -78,7 +78,7 @@ type Option func(*Service)
 // New starts a new Service with options.
 func New(options ...Option) (*Service, error) {
 	dialKeepaliveOpt := grpc.WithKeepaliveParams(keepalive.ClientParameters{
-		Time: 1 * time.Minute,
+		Time: 5 * time.Minute,
 	})
 	s := &Service{
 		endpoint:     os.Getenv(endpointEnv),
--- a/interface/cli/main.go
+++ b/interface/cli/main.go
@@ -26,7 +26,7 @@ func main() {
 	}
 
 	dialKeepaliveOpt := grpc.WithKeepaliveParams(keepalive.ClientParameters{
-		Time: 1 * time.Minute,
+		Time: 5 * time.Minute,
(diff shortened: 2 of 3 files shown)


client/service/service.go
```diff
@@ -78,7 +78,7 @@ type Option func(*Service)
 // New starts a new Service with options.
 func New(options ...Option) (*Service, error) {
 	dialKeepaliveOpt := grpc.WithKeepaliveParams(keepalive.ClientParameters{
-		Time: 1 * time.Minute,
+		Time: 5 * time.Minute,
```
krhubert: If `EnforcementPolicy.MinTime` is set to 5 min maybe we should set this a little bit higher, or decrease MinTime on the server?

NicolasMahe: @krhubert What did we say about client time again?

antho1404: also using this variable to explain the why 5 could be good

`Time: EnforcementPolicy.MinTime + Time.epsilon`

I don't know if epsilon exists in go but something like that to add a small delay

NicolasMahe: I think `EnforcementPolicy.MinTime` is enforce for at least 3 ping in this period. So even if `ClientParameters.Time` = `EnforcementPolicy.MinTime` it's find.

krhubert: EnforcementPolicy.MinTime - 1 min
Time for server - 1 min
Time for client - 2 min

NicolasMahe:
Closing in favor of https://github.com/mesg-foundation/core/pull/774

client/service/service.go
NicolasMahe: See https://forum.mesg.com/t/reduce-keepalive-values/247

PR #771, opened 2019-02-04
Don't return error when stream header doesn't contains any status

NicolasMahe:
The acknowledgement was returning an error when the stream header didn't contained `status`, thus overriding the error return by the stream..
The check are now eased to not return an error if header doesn't contains `status`.

It also prevent blocking the thread if the server doesn't implement acknowledgement header.

### DIFF
--- a/protobuf/acknowledgement/stream.go
+++ b/protobuf/acknowledgement/stream.go
@@ -23,7 +23,7 @@ func WaitForStreamToBeReady(stream grpc.ClientStream) error {
 	statuses := header.Get(statusKey)
 	statusesLen := len(statuses)
 	if statusesLen == 0 {
-		return fmt.Errorf("stream header does not contain any status")
+		return nil // Ignore headers with no status key
 	}
 	lastStatus := statuses[statusesLen-1]
 	if lastStatus != statusReady {


ilgooz:
by name _WaitForStreamToBeReady()_ should actually wait for status and return an error if it's not sent or not in 'ready'. status if we still need to remove that error with this PR, we should slightly refactor this func to maybe accept a 'header' type instead of 'stream' and check for existence of stream & 'ready' state. in this case we can rename this func with something like _CheckForStreamToBeReady_.

NicolasMahe:
Header are automatically send with the first message if not previous header has been sent.
So if the stream doesn't implement acknowledgement system, and send a message, it means the stream is actually ready. This PR allows this behavior.
When an error is returned on the stream before the acknowledgement header or any message, it also means that the stream is actually ready but contains an error the client needs to manage. This PR allows this behavior and moreover doesn't return an error (`stream header does not contain any status`) that is preventing the client to catch the actual error.

NicolasMahe:
@krhubert @ilgooz @antho1404 can you review or give feedback please

protobuf/acknowledgement/stream.go
```diff
@@ -22,12 +22,11 @@ func WaitForStreamToBeReady(stream grpc.ClientStream) error {
 	// header received. check status.
 	statuses := header.Get(statusKey)
 	statusesLen := len(statuses)
-	if statusesLen == 0 {
-		return fmt.Errorf("stream header does not contain any status")
-	}
-	lastStatus := statuses[statusesLen-1]
-	if lastStatus != statusReady {
-		return fmt.Errorf("stream header status is different than ready. Got %q", lastStatus)
+	if statusesLen > 0 {
```
antho1404:
```go
if statusesLen == 0 {
	return nil // Ignore headers with no status key
}
lastStatus := statuses[statusesLen-1]
if lastStatus != statusReady {
	return fmt.Errorf("stream header status is different than ready. Got %q", lastStatus)
}
```

NicolasMahe: Done

PR #772, opened 2019-02-04
Improve command logs errors

NicolasMahe:
Dependant of https://github.com/mesg-foundation/core/pull/771

The command logs was not returning error from the stream and the command was not even terminated.
So any error on the logs (like the service exists but doesn't run) was not displaying an nothing show up.

The err channel was not returned to the user and just discarded:
```go
go func() {	
  <-errC	
  closer()	
}()
```

I also add to change something in chunker: https://github.com/mesg-foundation/core/pull/772/files#diff-f59c295c71a834c1b8615304341180e4
The function `Close` was blocking..


antho1404:
Agree to merge this one when #771 is validated

PR #774, opened 2019-02-05
Update keep alive of client to 5min to prevent spamming the server

NicolasMahe:
Related to https://github.com/mesg-foundation/core/issues/760, https://github.com/mesg-foundation/core/issues/759 and https://github.com/mesg-foundation/core/pull/770

Client should not ping the server more often than the `EnforcementPolicy.MinTime` set on the server otherwise they get a GOAWAY error.
By default it's 5min, so I set the client to ping the server every 5min.

Also, by setting the client ping > to server ping, it also resolves the problem when both client and server were pigging themselve in both way. Now the client actually never has to ping the server if the connection is still open and the server pings the client.

This is a **critical fix** and needs to be released as soon as possible.


krhubert:
This fix worked for me

Issue #695, opened 2019-01-13
Variable `y` in mesg.yml triggers an error


antho1404:
When using a key `y` like a parameter this key is interpreted as `boolean` and mess with the parsing.
This is probably related to the parsing library that we use.

Workaround is to use a different name for the variable.

`n` also have the same unexpected behavior.

Example of file with issue
```yml
name: "myservice"
sid: "test"
tasks:
  taskX:
    inputs:
      y:
        type: Object
    outputs:
      success:
        data:
          message:
            type: String
```

krhubert:
Here is the pbm: https://github.com/mesg-foundation/core/blob/master/service/importer/service_file.go#L20

Yaml has several [representation](https://yaml.org/type/bool.html) of true/false.
So when yaml.Unmarshal see `y` it treats this as boolean. If we pass interface{} into unmarshal then the lib first recognizes it as boolean and put inside map[boolean]interface{}. 

Solutions:
- quick - replace `y` with `"y"` to make sure yaml sees it as string
- read mesg.yaml only with the defined struct (`impoter.ServiceDefinition`) and validate in go source it without converting to json.

Below is an example that demonstrates this behavior.

```go
package main

import (
	"fmt"

	"github.com/davecgh/go-spew/spew"
	yaml "gopkg.in/yaml.v2"
)

var data = []byte(`
y:
 x: foo
`)

func main() {
	var b interface{}
	var s = make(map[string]interface{})
	yaml.Unmarshal(data, &b)
	yaml.Unmarshal(data, &s)

	fmt.Print("interface{}: ")
	spew.Dump(b)
	fmt.Print("map[string]interface{}: ")
	spew.Dump(s)
}

```

output:

```
interface{}: (map[interface {}]interface {}) (len=1) {
 (bool) true: (map[interface {}]interface {}) (len=1) {
  (string) (len=1) "x": (string) (len=3) "foo"
 }
}
map[string]interface{}: (map[string]interface {}) (len=1) {
 (string) (len=1) "y": (map[interface {}]interface {}) (len=1) {
  (string) (len=1) "x": (string) (len=3) "foo"
 }
}
```

NicolasMahe:
Done in #781

PR #781, opened 2019-02-10
Replace json schema with go validation

antho1404:
Remove the json schema to replace it with a validation in go with https://github.com/go-playground/validator

It will be easier to plug some specific validation if we want too and it remove some extra complexity.


service/importer/definition.go
```diff
@@ -1,108 +1,111 @@
...
+
 // ServiceDefinition represents MESG services configurations.
 type ServiceDefinition struct {
 	// Name is the service name.
-	Name string `yaml:"name" json:"name,omitempty"`
+	Name string `yaml:"name" json:"name,omitempty" validate:"required,printascii,min=1"`
 
 	// Sid is the service id. It must be unique.
-	Sid string `yaml:"sid" json:"sid,omitempty"`
+	Sid string `yaml:"sid" json:"sid,omitempty" validate:"printascii,max=39"`
```
krhubert:
```suggestion
	Sid string `yaml:"sid" json:"sid,omitempty" validate:"printascii,max=32"`
```

service/importer/definition.go
```diff
@@ -1,108 +1,111 @@
...
-	Volumes []string `yaml:"volumes" json:"volumes,omitempty"`
+	Volumes []string `yaml:"volumes" json:"volumes,omitempty" validate:"unique,dive,printascii"`
 
 	// VolumesFrom are the docker volumes-from from.
-	VolumesFrom []string `yaml:"volumesfrom" json:"volumefrom,omitempty"`
+	VolumesFrom []string `yaml:"volumesfrom" json:"volumefrom,omitempty" validate:"unique,dive,printascii"`
 
 	// Ports holds ports configuration for container.
-	Ports []string `yaml:"ports" json:"ports,omitempty"`
+	Ports []string `yaml:"ports" json:"ports,omitempty" validate:"unique,dive,numeric"`
```
krhubert: need a validate range 0-65535 (but it cannot be done with validate it has to be done in func)

service/importer/definition.go
```diff
@@ -1,108 +1,111 @@
...
 	// Outputs are the definition of the execution results of task.
-	Outputs map[string]*Output `yaml:"outputs" json:"outputs,omitempty"`
+	Outputs map[string]*Output `yaml:"outputs" json:"outputs,omitempty" validate:"required,dive,keys,printascii,endkeys,required"`
 }
 
 // Output describes task output.
 type Output struct {
 	// Name is the name of task output.
-	Name string `yaml:"name" json:"name,omitempty"`
+	Name string `yaml:"name" json:"name,omitempty" validate:"printascii"`
```
krhubert:
```suggestion
	Name string `yaml:"name" json:"name,omitempty" validate:"printascii,min=1"`
```

service/importer/definition.go
```diff
@@ -1,108 +1,111 @@
...
 	// Env is a slice of environment variables in key=value format.
-	Env []string `yaml:"env" json:"env,omitempty"`
+	Env []string `yaml:"env" json:"env,omitempty" validate:"unique,dive,printascii"`
 }
 
 // Task describes a service task.
 type Task struct {
 	// Name is the name of task.
-	Name string `yaml:"name" json:"name,omitempty"`
+	Name string `yaml:"name" json:"name,omitempty" validate:"printascii"`
```
krhubert:
```suggestion
	Name string `yaml:"name" json:"name,omitempty" validate:"printascii,min=1"`
```

service/importer/definition.go
```diff
@@ -1,108 +1,111 @@
...
-	Data map[string]*Parameter `yaml:"data" json:"data,omitempty"`
+	Data map[string]*Parameter `yaml:"data" json:"data,omitempty" validate:"required,dive,keys,printascii,endkeys,required"`
 }
 
 // Parameter describes task input parameters, output parameters of a task
 // output and input parameters of an event.
 type Parameter struct {
 	// Name is the name of parameter.
-	Name string `yaml:"name" json:"name,omitempty"`
+	Name string `yaml:"name" json:"name,omitempty" validate:"printascii"`
```
krhubert:
```suggestion
	Name string `yaml:"name" json:"name,omitempty" validate:"printascii,min=1"`
```

service/importer/definition.go
```diff
@@ -1,108 +1,111 @@
...
 	// a Git host.
-	Repository string `yaml:"repository" json:"repository,omitempty"`
+	Repository string `yaml:"repository" json:"repository,omitempty" validate:"omitempty,uri"`
 }
 
 // Event describes a service task.
 type Event struct {
 	// Name is the name of event.
-	Name string `yaml:"name" json:"name,omitempty"`
+	Name string `yaml:"name" json:"name,omitempty" validate:"printascii"`
```
krhubert:
```suggestion
	Name string `yaml:"name" json:"name,omitempty" validate:"printascii,min=1"`
```

service/importer/definition.go
```diff
@@ -1,108 +1,111 @@
...
-	Data map[string]*Parameter `yaml:"data" json:"data,omitempty"`
+	Data map[string]*Parameter `yaml:"data" json:"data,omitempty" validate:"required,dive,keys,printascii,endkeys,required"`
 }
 
 // Dependency represents a Docker container and it holds instructions about
 // how it should run.
 type Dependency struct {
 	// Image is the Docker image.
-	Image string `yaml:"image" json:"image,omitempty"`
+	Image string `yaml:"image" json:"image,omitempty" validate:"printascii"`
```
krhubert:
```suggestion
	Image string `yaml:"image" json:"image,omitempty" validate:"printascii,min=1"`
```

service/importer/definition.go
```diff
@@ -1,108 +1,111 @@
...
-	Description string `yaml:"description" json:"description,omitempty"`
+	Description string `yaml:"description" json:"description,omitempty" validate:"printascii"`
 
 	// Inputs are the definition of the execution inputs of task.
-	Inputs map[string]*Parameter `yaml:"inputs" json:"inputs,omitempty"`
+	Inputs map[string]*Parameter `yaml:"inputs" json:"inputs,omitempty" validate:"dive,keys,printascii,endkeys,required"`
 
 	// Outputs are the definition of the execution results of task.
-	Outputs map[string]*Output `yaml:"outputs" json:"outputs,omitempty"`
+	Outputs map[string]*Output `yaml:"outputs" json:"outputs,omitempty" validate:"required,dive,keys,printascii,endkeys,required"`
```
krhubert:
```suggestion
	Outputs map[string]*Output `yaml:"outputs" json:"outputs,omitempty" validate:"required,dive,keys,printascii,endkeys"`
```

service/importer/definition.go
```diff
@@ -1,108 +1,111 @@
...
-	Name string `yaml:"name" json:"name,omitempty"`
+	Name string `yaml:"name" json:"name,omitempty" validate:"printascii"`
 
 	// Description is the description of task.
-	Description string `yaml:"description" json:"description,omitempty"`
+	Description string `yaml:"description" json:"description,omitempty" validate:"printascii"`
 
 	// Inputs are the definition of the execution inputs of task.
-	Inputs map[string]*Parameter `yaml:"inputs" json:"inputs,omitempty"`
+	Inputs map[string]*Parameter `yaml:"inputs" json:"inputs,omitempty" validate:"dive,keys,printascii,endkeys,required"`
```
krhubert:
```suggestion
	Inputs map[string]*Parameter `yaml:"inputs" json:"inputs,omitempty" validate:"required,dive,keys,printascii,endkeys"`
```

service/importer/definition.go
```diff
@@ -1,108 +1,111 @@
...
-	Name string `yaml:"name" json:"name,omitempty"`
+	Name string `yaml:"name" json:"name,omitempty" validate:"printascii"`
 
 	// Description is the description of task output.
-	Description string `yaml:"description" json:"description,omitempty"`
+	Description string `yaml:"description" json:"description,omitempty" validate:"printascii"`
 
 	// Data holds the output parameters of a task output.
-	Data map[string]*Parameter `yaml:"data" json:"data,omitempty"`
+	Data map[string]*Parameter `yaml:"data" json:"data,omitempty" validate:"required,dive,keys,printascii,endkeys,required"`
```
krhubert:
```suggestion
	Data map[string]*Parameter `yaml:"data" json:"data,omitempty" validate:"required,dive,keys,printascii,endkeys"`
```

service/importer/definition.go
```diff
@@ -111,5 +114,5 @@ type Parameter struct {
 	Repeated bool `yaml:"repeated" json:"repeated,omitempty"`
 
 	// Definition of the structure of the object when the type is object
-	Object map[string]*Parameter `yaml:"object" json:"object,omitempty"`
+	Object map[string]*Parameter `yaml:"object" json:"object,omitempty" validate:"dive,keys,printascii,endkeys,required"`
```
krhubert:
```suggestion
	Object map[string]*Parameter `yaml:"object" json:"object,omitempty" validate:"required,dive,keys,printascii,endkeys"`
```

service/importer/service_file_test.go
```diff
@@ -21,60 +20,34 @@ func TestReadServiceFileDoesNotExist(t *testing.T) {
...
 	require.Equal(t, 1, len(warnings))
 }
 
 func TestValidateServiceFileWithMultipleErrors(t *testing.T) {
 	data, _ := readServiceFile("./tests/service-multiple-errors")
-	warnings, err := validateServiceFile(data)
-	require.NoError(t, err)
-	require.Equal(t, 2, len(warnings))
+	warnings := validateServiceFile(data)
+	require.Equal(t, 1, len(warnings))
```
ilgooz: there is only 1 error where this test is about multiple errors

service/importer/definition.go
ilgooz: Ince case when a task doesn't necessarily need to send an output then we might allow empty outputs. But on the other hand, sending outputs _always_ can be considered as a good practice because there is no another way of confirming if a task is successfully executed by the service or not.

service/importer/definition.go
antho1404: Same here, `Name` is not required, the key in the map is

service/importer/definition.go
antho1404: same

service/importer/definition.go
antho1404: same

service/importer/service_file_test.go
antho1404: Good point, now it's one parsing error instead of multiple validation like before, we could get rid of this test or rename it as parsing error test

antho1404:
> require at least one of the `events` or `tasks` sections 
Is it really necessary, I agree that it would be almost required but it's not something totally useless to have services without any tasks/event. We could have a service that basically just have one or multiple dependencies like a `docker-compose`. Maybe we can keep it simple for now, I'm open to both solutions

service/importer/definition.go
krhubert: So it's from different vliadator package, function is port looks like :

```
// IsPort checks if a string represents a valid port
func IsPort(str string) bool {
	if i, err := strconv.Atoi(str); err == nil && i > 0 && i < 65536 {
		return true
	}
	return false
}
```

And we could add this as our custom validator to package that we are using

service/importer/definition.go
krhubert: Hum, so it makes things bit complicated. In that case we should validate only dependencies if image is not empty and here required must be dropped

service/importer/definition.go
```diff
@@ -1,108 +1,111 @@
...
 	// Events are the list of events that service can emit.
-	Events map[string]*Event `yaml:"events" json:"events,omitempty"`
+	Events map[string]*Event `yaml:"events" json:"events,omitempty" validate:"dive,keys,printascii,endkeys,required"`
 
 	// Dependencies are the Docker containers that service can depend on.
-	Dependencies map[string]*Dependency `yaml:"dependencies" json:"dependencies,omitempty"`
+	Dependencies map[string]*Dependency `yaml:"dependencies" json:"dependencies,omitempty" validate:"dive,keys,printascii,ne=service,endkeys,required"`
 
 	// Configuration is the Docker container that service runs inside.
 	Configuration *Dependency `yaml:"configuration" json:"configuration,omitempty"`
```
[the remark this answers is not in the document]
antho1404: not sure about this one, by default the `Dependency` should be checked and this is optional field

krhubert:
The question is - if someone will need a service without `event` or `task`? what will be usecase for service which you can't communicate with? If user has such use case then we even must allow both events and tasks to be empty. If there is no use case then importer should validate it and in the future we can take off this restriction if we find usecase for service without events or tasks.

antho1404:
I just worked on an IPFS this weekend that runs in a MESG service but that actually didn't expose anything except the dependency port to the IPFS node. I agree that this is not the best use of MESG (and I'm actually trying to merge it in the core directly) but this is a use case, just accessing a server from a dependency.

But again both are fine to me

krhubert:
If there is a use-case (event as you describe not the best) then we should allow such configuration 💪

service/importer/definition.go
krhubert: Ok I have update on this:

- based on [rfc2181](https://tools.ietf.org/html/rfc2181#section-11) and [stackoverflow answer](https://stackoverflow.com/questions/2180465/can-domain-name-subdomains-have-an-underscore-in-it)
dns names can have under score.

- there is implementation in golang [net](https://github.com/golang/go/blob/master/src/net/dnsclient.go#L63) package, which we can use as is already working (we need only copy it to our src because it isn't exported)

NicolasMahe: @krhubert can you show the new regex that needs to be use?

krhubert: There is no regexp, look at the net package and the `for` loop. Regexp is not needed

krhubert:
Also it will close #695

krhubert:
I will replace gopkg.in/go-playground/validator.v9 with github.com/asaskevich/govalidator

motiviationŁ
- there is one big issue with validator.v9 - it dosen't export functions to validate so we can only use it inside structs. On the other side govalidator exposes all functions and we are using it in commands/provider and we might want use this in the future.

service/importer/definition.go
```diff
@@ -1,108 +1,110 @@
...
+
 // ServiceDefinition represents MESG services configurations.
 type ServiceDefinition struct {
 	// Name is the service name.
-	Name string `yaml:"name" json:"name,omitempty"`
+	Name string `yaml:"name" json:"name,omitempty" validate:"required,printascii,min=1"`
 
 	// Sid is the service id. It must be unique.
-	Sid string `yaml:"sid" json:"sid,omitempty"`
+	Sid string `yaml:"sid" json:"sid,omitempty" validate:"printascii,max=63,domain"`
```
NicolasMahe: The Sid should not be required.

```
        * ServiceDefinition.Sid with value "" is invalid: domain 
```

x/xvalidator/validator.go
```diff
@@ -0,0 +1,29 @@
...
+func IsPort(fl validator.FieldLevel) bool {
+	ports := strings.Split(fl.Field().String(), ":")
+	for _, port := range ports {
+		i, err := strconv.Atoi(port)
+		if !(err == nil && 0 < i && i <= 65535) {
+			return false
+		}
+	}
+	return true
+}
```
NicolasMahe: IsPort is now compatible with port notation like xx:yy as the service.start function is also compatible with

service/importer/service_file.go
```diff
@@ -4,70 +4,84 @@ import (
...
+		for _, e := range errs {
+			// Remove the name of the struct and the field from namespace
+			trimmedNamespace := strings.TrimPrefix(e.Namespace(), namespacePrefix)
+			trimmedNamespace = strings.TrimSuffix(trimmedNamespace, e.Field())
+			// Only use it when in-cascade field
+			namespace := ""
+			if e.Field() != trimmedNamespace {
+				namespace = trimmedNamespace
+			}
+			warnings = append(warnings, fmt.Sprintf("%s%s", namespace, e.Translate(trans)))
```
NicolasMahe: I change from custom error output to use the `Translate` function available in govalidator that format nicely the error.

I add namespace to have specific "path" to the error, eg:
```
        * Tasks[decodeLog].Inputs[logIndex].Type must be one of [String Number Boolean Object Any]
```
instead of default:
```
        * Type must be one of [String Number Boolean Object Any]
```

service/importer/service_file.go
```diff
@@ -4,70 +4,84 @@ import (
...
+		t, _ := ut.T("port", fe.Field(), namespacePrefix)
+		return t
+	})
+	validate.RegisterValidation("domain", xvalidator.IsDomainName)
+	validate.RegisterTranslation("domain", trans, func(ut ut.Translator) error {
+		return ut.Add("domain", "{0} must respect domain-style notation. Eg: author.name", false)
+	}, func(ut ut.Translator, fe validator.FieldError) string {
+		t, _ := ut.T("domain", fe.Field())
+		return t
+	})
```
NicolasMahe: Custom validation can have custom translation to output the error really nicely

service/importer/service_file.go
```diff
@@ -4,70 +4,84 @@ import (
...
+	validate.RegisterValidation("port", xvalidator.IsPort)
+	validate.RegisterTranslation("port", trans, func(ut ut.Translator) error {
+		return ut.Add("port", "{0} must be a valid port", false)
+	}, func(ut ut.Translator, fe validator.FieldError) string {
+		t, _ := ut.T("port", fe.Field(), namespacePrefix)
+		return t
+	})
+	validate.RegisterValidation("domain", xvalidator.IsDomainName)
+	validate.RegisterTranslation("domain", trans, func(ut ut.Translator) error {
+		return ut.Add("domain", "{0} must respect domain-style notation. Eg: author.name", false)
```
NicolasMahe: If somebody has a better error message than:
```
must respect domain-style notation. Eg: author.name
```
this is for sid.

NicolasMahe:

commands/provider/service_deployer.go
```diff
@@ -56,7 +56,7 @@ func (p *ServiceProvider) ServiceDeploy(path string, env map[string]string, stat
 	}()
 	go readDeployReply(stream, deployment, statuses)
 
-	if govalidator.IsURL(path) {
+	if _, err := url.Parse(path); err != nil {
```
ilgooz: I think, `url.Parse()` is not enough, it doesn't require a url schema and won't return an error.

https://play.golang.org/p/g_5OQhzyWv7

ilgooz: I think, `url.Parse()` is not enough, it doesn't require a url schema and won't return an error.

https://play.golang.org/p/g_5OQhzyWv7

x/xvalidator/validator.go
krhubert: It has a bug, you can write port in format `80:80:80` and it will be ok.

Also it should be renamed to `IsPortMapping` and register under `port_map(ing?)` tag

krhubert: Also I used value like 65535 and it will be good to change them to constant.

We can think of adding `/tcp` `/udp` to validation but it's not required in this PR

service/importer/service_file.go
krhubert: Nice! ❤️ 

Is would be nice to have fields in loware case (like we do in mesg.yaml) so 

`task[XXX].inputs[YYY].type bla bla bla`

service/importer/service_file.go
```diff
@@ -4,70 +4,84 @@ import (
...
 
-func validateServiceFileDependencyKey(data interface{}) (warning string) {
-	s, _ := data.(map[string]interface{})
-	dep, _ := s["dependencies"].(map[string]interface{})
-	if dep[ConfigurationDependencyKey] != nil {
-		return fmt.Sprintf("cannot use %q as dependency key", ConfigurationDependencyKey)
-	}
-	return ""
-}
+func validateServiceStruct(service *ServiceDefinition) []string {
```
krhubert: Validation is done only for struct fields, but it need some few more validations (Volumes and dependencies etc...)

You can take a look 
https://github.com/mesg-foundation/core/blob/service-refactor/service/definition.go#L63

But it could be added with refactoring PR

commands/provider/service_deployer.go
krhubert: So maybe replace it with os.Stat() && IsDir then check if is url?

service/importer/service_file.go
NicolasMahe: Yes. But I don't know what is it uppercased.
For instance, `e.Field()` is containing the field declare in the json label, as opposite as `StructField` that return the Field name of the Go struct.

x/xvalidator/validator.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/mesg-foundation/core/x/xnet"
+	validator "gopkg.in/go-playground/validator.v9"
+)
+
+// IsDomainName validates if given field is valid domain name.
+func IsDomainName(fl validator.FieldLevel) bool {
+	if fl.Field().String() == "" {
+		return true
+	}
+	return xnet.IsDomainName(fl.Field().String())
```
NicolasMahe: @krhubert why using a custom domain name verification?

This https://godoc.org/gopkg.in/go-playground/validator.v9#hdr-Hostname_RFC_952 or https://godoc.org/gopkg.in/go-playground/validator.v9#hdr-Hostname_RFC_1123 are not enough?

x/xvalidator/validator.go
NicolasMahe: Done. Let's do the `/tcp` and `/udp` in another PR with also the implementation in `service.Start`

commands/provider/service_deployer.go
NicolasMahe: Fixed by using `go-playground/validator.v9`

service/importer/service_file.go
NicolasMahe: Done

NicolasMahe:
@mesg-foundation/core modification done. please review.

x/xvalidator/validator.go
krhubert: Hostname != domainname

https://github.com/golang/go/blob/master/src/net/dnsclient.go#L63

// See RFC 1035, RFC 3696.

service/importer/service_file.go
krhubert: I mean e.Fields because all fields in mesg.yml are lowerace

NicolasMahe: @krhubert the modif you did, didn't work.
The output is:
```
        * NameName is a required field
```

`e.Translate(trans)` is returning `Type must be one of [String Number Boolean Object Any]` (uppercased).
The system I did is only adding the beginning of the sentence. I don't want to spend hours fixing this problem. It's fine ;)

NicolasMahe: If you try to lowercase all the string, then the user's camelcase key will also be lowercase. I don't see easy solution for this very small problem.

NicolasMahe:

service/importer/service_file.go
```diff
@@ -4,70 +4,121 @@ import (
...
 
-func validateServiceFileSchema(data interface{}, schemaPath string) (*gojsonschema.Result, error) {
-	schemaData, err := assets.Asset(schemaPath)
-	if err != nil {
-		return nil, err
+	for key, dep := range service.Dependencies {
+		if dep == nil {
+			continue
+		}
+		if dep.Image == "" {
```
ilgooz: why don't we check this with govalidator's 'required' tag?

antho1404: because it depends on the context, if in a dependency then the image is required, if in the configuration no so it's more complicated to have this with govalidator than just code it

x/xvalidator/validator.go
```diff
@@ -8,23 +8,28 @@ import (
...
 
+const (
+	minPort = 1
+	maxPort = 65535
+
+	portSeparator = ":"
+)
+
 // IsDomainName validates if given field is valid domain name.
 func IsDomainName(fl validator.FieldLevel) bool {
```
NicolasMahe: @krhubert The following code was necessary. Please don't remove if you don't test.
```
	if fl.Field().String() == "" {	
		return true	
	}
```

ilgooz:
@NicolasMahe have you tested each rule manually? Or I need to do that because we don't have tests. Code seems fine.

NicolasMahe:
> @NicolasMahe have you tested each rule manually? Or I need to do that because we don't have tests. Code seems fine.

I test a lot manually.

ilgooz:
> > @NicolasMahe have you tested each rule manually? Or I need to do that because we don't have tests. Code seems fine.
> 
> I tested a lot manually.

Ok, let me do one more in case if we miss something and then I'll merge.

x/xvalidator/validator.go
krhubert: The following code is wrong and should not be here. It's not valid domain.

You can check the default validators here - https://github.com/go-playground/validator/blob/v9.27.0/baked_in.go#L64

None of them check for empty strings. If here check appears then something else is wrong.

ilgooz:
@NicolasMahe I cannot merge because some commits are not signed.

x/xvalidator/validator.go
NicolasMahe: 👍 condition removed and tag `omitempty` added to sid

NicolasMahe:
i fixed the optional SID and added some test

please review

antho1404:
finished my tests and everything is fine :)

PR #784, opened 2019-02-12
System service marketplace

NicolasMahe:


systemservices/marketplace/src/index.ts
```diff
@@ -0,0 +1,112 @@
...
+    createServiceVersion: createServiceVersion(contract, createTransaction),
+    disableServiceOffer: disableServiceOffer(contract, createTransaction),
+    purchase: purchase(contract, createTransaction),
+    transferServiceOwnership: transferServiceOwnership(contract, createTransaction),
+    sendSignedTransaction: sendSignedTransaction(web3),
+    isAuthorized: isAuthorized(contract),
+  })
+  .on('error', error => console.error('catch listenTask', error))
+
+  // const newBlock = await newBlockEventEmitter(web3, blockConfirmations, null, pollingTime)
```
NicolasMahe: TODO

systemservices/marketplace/mesg.yml
```diff
@@ -0,0 +1,414 @@
...
+      success:
+        name: ""
+        description: ""
+        data: *service
+      error: *outputError
+  getServiceVersion:
+    name: ""
+    description: ""
+    inputs:
+      sid: *inputSid
```
antho1404: sid is not necessary here, we have the hash and that's all we need on this api

systemservices/marketplace/mesg.yml
```diff
@@ -0,0 +1,414 @@
...
+            name: ""
+            description: ""
+            type: Boolean
+      error: *outputError
+  createService:
+    name: ""
+    description: ""
+    inputs:
+      sid: *inputSid
+      from: &inputFrom
```
antho1404: should we rename into `author`, it makes more sense than `from`

systemservices/marketplace/mesg.yml
```diff
@@ -0,0 +1,414 @@
...
+            type: Number
+          gasPrice:
+            name: "Gas price"
+            description: "The gas price in Wei."
+            type: String
+          data:
+            name: "Data"
+            description: "The data of the transaction."
+            type: String
+  createServiceOffer:
```
antho1404: We could keep this simple and not manage the offers

systemservices/marketplace/mesg.yml
```diff
@@ -0,0 +1,414 @@
...
+        name: ""
+        description: ""
+        type: String
+      from: *inputFrom
+      gas: *inputGas
+      gasPrice: *inputGasPrice
+    outputs:
+      error: *outputError
+      success: *outputTransaction
+  disableServiceOffer:
```
antho1404: I suggest to remove this. This can be done directly on the markeplace website. Keep it simple :)

systemservices/marketplace/mesg.yml
```diff
@@ -0,0 +1,414 @@
...
+        name: ""
+        description: ""
+        type: String
+      from: *inputFrom
+      gas: *inputGas
+      gasPrice: *inputGasPrice
+    outputs:
+      error: *outputError
+      success: *outputTransaction
+  transferServiceOwnership:
```
antho1404: I suggest to remove this. This is a really advanced mode and can be done directly on etherscan or something like that

systemservices/marketplace/mesg.yml
```diff
@@ -0,0 +1,414 @@
...
+      error: *outputError
+      success:
+        name: ""
+        description: ""
+        data:
+          authorized:
+            name: ""
+            description: ""
+            type: Boolean
+  checkForDeployment:
```
antho1404: what's the purpose of this one ?

systemservices/marketplace/mesg.yml
```diff
@@ -0,0 +1,414 @@
...
+            name: ""
+            description: ""
+            type: String
+            optional: true
+          type:
+            name: ""
+            description: ""
+            type: String
+            optional: true
+events:
```
antho1404: Do we listen any events from the core or cli ?
For now not that I know so I wonder if it's not better to add them when we actually need them and keep this service as simple as possible and just add the stuff we actually need

systemservices/marketplace/mesg.yml
NicolasMahe: it's the `from` parameter for creating / signing the transaction, like all the other tasks that return a transaction.
It's not directly related to the `owner` of the service.

systemservices/marketplace/mesg.yml
NicolasMahe: We have the manage offer. It's a basic functionality.

systemservices/marketplace/mesg.yml
NicolasMahe: This task is called by the CLI or Core when a service need to be deployed.
It check if one of the account of the wallet has is authorize (has paid) the service.

systemservices/marketplace/mesg.yml
NicolasMahe: The marketplace website backend should listen event and refresh the data accordingly. So yes it's useful.

systemservices/marketplace/mesg.yml
antho1404: I understand that but this api is not related to a transaction on ethereum it's about creating a service. What does it means a from to create a service.
Otherwise we should rename this api with `createServiceTransaction` where the outputs would make more sense and here the from is making sense too.

systemservices/marketplace/mesg.yml
antho1404: this can also be done in the marketplace website. They put their services to the marketplace and then we give them a link on the marketplace and they have to add the offer on top of their service with a nice ui

systemservices/marketplace/mesg.yml
antho1404: so it needs to be renamed. If this function is checking the authorization to access a service maybe `checkauthorisation` here the name sounds like we check if we can deploy and we are on the service marketplace so there is no notion of deployment on the core just deployment on the marketplace

antho1404:
```suggestion
  checkAuthorization:
```

antho1404:
simplified version here https://github.com/mesg-foundation/core/compare/dev...ss/marketplace-simplified

PR #786, opened 2019-02-13
cli marketplace commands

NicolasMahe:
Depend on branch `ss/marketplace` and `feature/ethwallet-cli`.

The service marketplace needs to be start manually for now.


commands/provider/marketplace_provider.go
```diff
@@ -0,0 +1,428 @@
...
+		return "", "", err
+	}
+	if result.Error != "" {
+		return "", "", errors.New(result.Error)
+	}
+	return result.OutputKey, result.OutputData, nil
+}
+
+// listenEvents returns a channel with event data streaming.
+func (p *MarketplaceProvider) listenEvents(id, eventFilter string) (chan *coreapi.EventData, chan error, error) {
```
antho1404: All these functions should be extracted later in a cli helper package for example

commands/wallet_delete.go
```diff
@@ -0,0 +1,57 @@
...
+	c.cmd = newCommand(&cobra.Command{
+		Use:     "delete",
+		Short:   "Delete an account",
+		Long:    "Delete an account from the wallet",
+		Example: "mesg-core wallet delete 0x0",
+		Args:    cobra.ExactArgs(1),
+		RunE:    c.runE,
+	})
+
+	c.cmd.Flags().BoolVarP(&c.noPassphrase, "no-passphrase", "n", c.noPassphrase, "Leave passphrase empty")
```
antho1404: why is there a noPassphrase ? we should always have one no ?

krhubert: @NicolasMahe said that is ok to do not have passpharse, so I added the flag to avoid creating wallet without pass by mistake

PR #787, opened 2019-02-13
Wallet + Marketplace commands

krhubert:
Depends on https://github.com/mesg-foundation/core/pull/784 for the marketplace service dev, and https://github.com/mesg-foundation/core/pull/787.

- Commands `wallet` done
- Command `marketplace publish` can publish a service with a version on the marketplace and ipfs.

The service marketplace still need to be started manually!
```
./dev-cli service dev ./systemservices/marketplace
```

Still a lot to do, and a lot of TODO in the code ;)


commands/provider/service_deployer.go
```diff
@@ -91,51 +93,65 @@ func (p *ServiceProvider) deployServiceFromMarketplace(u string, env map[string]
...
+
+	// manifest, protocol,
+	switch version.ManifestProtocol {
+	case "https":
+		return stream.Send(&coreapi.DeployServiceRequest{
+			Value: &coreapi.DeployServiceRequest_Url{Url: version.Manifest},
+			Env:   env,
+		})
+	case "ipfs":
+		tarball, err := &bytes.Buffer{}, error(nil) // TODO: merge ipfs first ipfs.Get(version.Manifest)
```
NicolasMahe: Why not just getting the URL of api gateway and forward it to the core?
Let's do the "real" implementation of IPFS, in the core, and later.

commands/provider/service_deployer.go
```diff
@@ -57,7 +61,12 @@ func (p *ServiceProvider) ServiceDeploy(path string, env map[string]string, stat
 	go readDeployReply(stream, deployment, statuses)
 
 	if govalidator.IsURL(path) {
```
NicolasMahe: The following string is not recognized as a URL:
`mesg.marketplace:<sid>:<hash>`

antho1404:
Can we extract only the wallet commands (that I would love to see renamed as account).
This PR is doing a lot and can be extracted into 3
- Wallet/Account commands
- Marketplace system service (https://github.com/mesg-foundation/core/pull/797)
- Commands for the marketplace system service

PR #788, opened 2019-02-14
Refactor ipfs client + add tests

krhubert:
There was also bug in previous ipfs - resp.Body wasn't close


krhubert:
I think we still have a bug in Add. With every add the content will be diffrent (so the output hash), but it should be the same (like with `ipfs add` cmd)

```
➜  multipart go run main.go
--a994c3e69377bf9cfcf39ed913aa9601c9a9fbf308227bb349c278f25f83
Content-Disposition: form-data; name="file"; filename="some-name"
Content-Type: application/octet-stream

x

--a994c3e69377bf9cfcf39ed913aa9601c9a9fbf308227bb349c278f25f83--

QmTP4tQZEG66zwaHPJVuuwaY3P1C3RedqeWjjfzkwdyne7 

➜  multipart go run main.go
--96ea20c35d7fbebd3c06e81b9484483b48349eaf27f1036012be51e1d563
Content-Disposition: form-data; name="file"; filename="some-name"
Content-Type: application/octet-stream

x

--96ea20c35d7fbebd3c06e81b9484483b48349eaf27f1036012be51e1d563--

Qmez8YfputXFtCGoFgTJWcrjRCbv18tta62VW4vUoUzjSy 
```

PR #789, opened 2019-02-17
Add URL function to ipfs package to simply return the url from the provider

NicolasMahe:
Needed for https://github.com/mesg-foundation/core/pull/787

### DIFF
--- a/ipfs/basic.go
+++ b/ipfs/basic.go
@@ -54,3 +54,9 @@ func (ipfs *IPFS) Get(hash string) (io.Reader, error) {
 	}
 	return resp.Body, nil
 }
+
+// URL retrives the file's url from the ipfs node.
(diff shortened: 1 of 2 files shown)


antho1404:
Is this really needed for #787 ? Can't this dev use the `DefaultEndpoint`, I'm afraid that this will not work if we move that in a container with our own node or and also what is the purpose of having the api ? I guess what you want is more the final endpoint (like with IPFS Gateway)

NicolasMahe:
> Is this really needed for #787 ? Can't this dev use the `DefaultEndpoint`, I'm afraid that this will not work if we move that in a container with our own node or and also what is the purpose of having the api ? I guess what you want is more the final endpoint (like with IPFS Gateway)

About #787, because the core cannot download the service, I will try to directly download the service in CLI. So this PR might not be useful.

PR #790, opened 2019-02-20
Stop services when core stop

antho1404:
Right now we rely on the CLI to stop all service before stoping the core. This is not really nice and create issues if we stop the core with another CLI (`docker service rm` for exemple).

When the Core stops, it will stop all the running services, close the api and then exit.

### DIFF
--- a/core/main.go
+++ b/core/main.go
@@ -8,7 +8,6 @@ import (
 	"github.com/mesg-foundation/core/database"
 	"github.com/mesg-foundation/core/interface/grpc"
 	"github.com/mesg-foundation/core/logger"
-	"github.com/mesg-foundation/core/service"
 	"github.com/mesg-foundation/core/version"
 	"github.com/mesg-foundation/core/x/xsignal"
 	"github.com/sirupsen/logrus"
@@ -80,14 +79,8 @@ func stopRunningServices(api *api.API) error {
 		return err
 	}
 	for _, s := range services {
-		status, err := s.Status()
-		if err != nil {
+		if err := api.StopService(s.Hash); err != nil {
 			return err
-		}
-		if status == service.STARTING || status == service.PARTIAL || status == service.RUNNING {
-			if err := api.StopService(s.Hash); err != nil {
-				return err
-			}
 		}
 	}
 	return nil
--- a/core/main.go
+++ b/core/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"log"
 	"path/filepath"
 
 	"github.com/mesg-foundation/core/api"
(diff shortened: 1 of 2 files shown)


antho1404:
not really necessary, services will anyway be stopped but that simplifies the cli for sure

antho1404:
@mesg-foundation/core can you please review

core/main.go
```diff
@@ -86,5 +120,8 @@ func main() {
 	}()
 
 	<-xsignal.WaitForInterrupt()
+	if err := stopRunningServices(dep.api); err != nil {
+		logrus.Fatalln(err)
+	}
 	server.Close()
```
ilgooz: closing gRPC server first is better otherwise some services can be started by devs during the stopping process

core/main.go
```diff
@@ -1,48 +1,62 @@
...
+		return &dep, err
+	}
+	dep.config = config
+
 	// init services db.
-	db, err := database.NewServiceDB(filepath.Join(c.Core.Path, c.Core.Database.ServiceRelativePath))
+	serviceDB, err := database.NewServiceDB(filepath.Join(dep.config.Core.Path, dep.config.Core.Database.ServiceRelativePath))
 	if err != nil {
-		return nil, err
+		return &dep, err
```
ilgooz: not blocking but we can follow returning _nil_ pattern instead of '&dep'

core/main.go
```diff
@@ -1,48 +1,62 @@
...
-func initGRPCServer(c *config.Config) (*grpc.Server, error) {
+type dependencies struct {
+	config      *config.Config
+	serviceDB   database.ServiceDB
+	executionDB database.ExecutionDB
+	api         *api.API
+}
+
+func initDependencies() (*dependencies, error) {
+	dep := dependencies{}
```
ilgooz: nothin blocking but creating pointer here is better `dep := &dependencies{}`

core/main.go
ilgooz: not blocking but we can follow returning _nil_ pattern instead of '&dep'

core/main.go
```diff
@@ -1,48 +1,62 @@
...
+	dep.config = config
+
 	// init services db.
-	db, err := database.NewServiceDB(filepath.Join(c.Core.Path, c.Core.Database.ServiceRelativePath))
+	serviceDB, err := database.NewServiceDB(filepath.Join(dep.config.Core.Path, dep.config.Core.Database.ServiceRelativePath))
 	if err != nil {
-		return nil, err
+		return &dep, err
 	}
+	dep.serviceDB = serviceDB
```
ilgooz: nothing blocking but I'd do this assignments once at the bottom of this func for all of the fields together, not one by one

core/main.go
antho1404: good point but at the same time these services rely on the api so to gracefully shutdown the api we need to close all the connections first. If we don't the `server.Close()` is waiting for it and never stops.

core/main.go
antho1404: done

core/main.go
antho1404: done

core/main.go
antho1404: done

core/main.go
```diff
@@ -61,22 +74,42 @@ func deployCoreServices(c *config.Config, api *api.API) error {
...
+	services, err := api.ListServices()
+	if err != nil {
+		return err
+	}
+	for _, s := range services {
+		status, err := s.Status()
+		if err != nil {
+			return err
+		}
+		if status == service.STARTING || status == service.PARTIAL || status == service.RUNNING {
```
NicolasMahe: This test is useless because it's done in `service.Stop()` already:
https://github.com/mesg-foundation/core/blob/348aa982644ff31e9571813b298919dd99ed1c85/service/stop.go#L10-L20

core/main.go
```diff
@@ -61,22 +74,42 @@ func deployCoreServices(c *config.Config, api *api.API) error {
...
 
 	// init logger.
-	logger.Init(c.Log.Format, c.Log.Level, c.Log.ForceColors)
+	logger.Init(dep.config.Log.Format, dep.config.Log.Level, dep.config.Log.ForceColors)
 
-	// init gRPC server.
-	server, err := initGRPCServer(c)
-	if err != nil {
+	// init system services.
+	if err := deployCoreServices(dep.config, dep.api); err != nil {
```
NicolasMahe: I think it make more sense to start the gRPC server before deploying the Core services:
The services will fail (and restart) until the gRPC server is up and running.

core/main.go
```diff
@@ -61,22 +74,42 @@ func deployCoreServices(c *config.Config, api *api.API) error {
 	return nil
 }
 
+func stopRunningServices(api *api.API) error {
+	services, err := api.ListServices()
+	if err != nil {
+		return err
+	}
+	for _, s := range services {
```
NicolasMahe: Could all the services be stop at the same time in parallel?

core/main.go
NicolasMahe: I agree with you @antho1404.
The perfect way will be do "deactivate" the api service start but I think it's too much for now.

NicolasMahe:
> lgtm, but we got to delete previous service stopping part from the commands provider #792

Modif of https://github.com/mesg-foundation/core/pull/792/files are not in this PR anymore. Maybe because of a force push.

core/main.go
antho1404: done

core/main.go
antho1404: It could I just wanted to make it simple for now. This can be done in another PR

core/main.go
antho1404: just kept the same behavior here as before just refactors it but the behavior is the same.
I agree it's not perfect, optimal solution would be:
- deploy service
- start service api
- start deployed services
- start core api

But this is not the purpose of this PR

antho1404:
modifications from https://github.com/mesg-foundation/core/pull/792 in the PR again.

core/main.go
ilgooz: How about closing the [listener](https://github.com/mesg-foundation/core/blob/62bf84c/interface/grpc/server.go#L51)? That way grpc pkg will not be able to _Accept()_ new connections but currently accepted ones still be active.

I moved this to a new [issue](https://github.com/mesg-foundation/core/issues/803), please resolve this conversation after you read.

ilgooz:
Manual tests are OK. I see weird bugs sometimes, ethwallet service seems running but it's Docker service is actually not exists. I'll open an issue if I can reproduce.

PR #799, opened 2019-03-07
Refactor

krhubert:


antho1404:
I had a quick look, I have few feedback for now: 

- API package should accept the service as parameter for the stop, start etc... instead of the sid/hash like that we can `api.StartService(api.GetService())` this will be easier to delegate to a system service

- Why reverting to events/task using map and not array

- I like the service manager but why not even delegating this directly to the API to simplify

krhubert:
> API package should accept the service as parameter for the stop, start etc... instead of the sid/hash like that we can api.StartService(api.GetService()) this will be easier to delegate to a system service

If API accepts service then who manage databases to get them?
grpc server should manage communication layer only.

The flow right now is:
grpc -> api -> service -> container

If you want to accept `service struct` in api package then we need something between grpc and ap

> I like the service manager but why not even delegating this directly to the API to simplify

To avoid api package becoming bigger. And I felt that manager struct belongs more to service (as it only operates on service). Moving this to api means also that api imports container package so it measn it depends on service and container package rather then service itself.

> Why reverting to events/task using map and not array

From discussion on discord we end up with:

The array keeps the order or parameters and it's more deterministic definition of service so the array is preferable way here. The second thing is that definitions and protobuf should be same if possible, so if we have array we will have them in mesg.yaml as well (@antho1404 please confim that what I wrote cover our discussion or add missing part)

ilgooz:
This PR is too big, it needs to be split to smaller ones with explanatory commit messages.

antho1404:
> The array keeps the order or parameters and it's more deterministic definition of service so the array is preferable way here. The second thing is that definitions and protobuf should be same if possible, so if we have array we will have them in mesg.yaml as well

Not exactly.
- Keeping the array structure because more precise 👍 
- definition and protobuf similar is a nice to have
- mesg.yml is a representation so don't need to have arrays, this is oversimplification that we can do if we want too but this impacts a lot existing services and reduces a lot the readability of the yaml so i suggest to keep the parsing of the yaml and convert it in the data structure with array except if we all agree to move with array only on the yaml

core/main.go
```diff
@@ -56,15 +62,10 @@ func initDependencies() (*dependencies, error) {
...
 		logrus.Infof("Deploying service %q from %q", service.Key, service.URL)
-		s, valid, err := api.DeployServiceFromURL(service.URL, service.Env)
-		if valid != nil {
-			return valid
-		}
+		s, err := api.DeployServiceFromURL(service.URL, service.Env, nil)
 		if err != nil {
 			return err
 		}
-		service.Sid = s.Sid
```
antho1404: These 2 lines of code are important in order to have the informations in the config especially when we get informations with the `Info` api

database/service_db.go
```diff
@@ -165,9 +164,6 @@ func (d *LevelDBServiceDB) Save(s *service.Service) error {
 	if s.Hash == "" {
 		return errCannotSaveWithoutHash
 	}
-	if s.Sid == "" {
```
NicolasMahe: Why removing this test?

The db still save the sid:

```
	// save sid-hash pair of service.
	if err := tx.Put([]byte(sidKeyPrefix+s.Sid), []byte(s.Hash), nil); err != nil {
		tx.Discard()
		return err
```

execution/execution.go
```diff
@@ -1,10 +1,11 @@
 package execution
 
 import (
+	"encoding/hex"
 	"time"
 
 	"github.com/mesg-foundation/core/service"
-	"github.com/mesg-foundation/core/x/xstructhash"
+	"github.com/mesg-foundation/core/utils/structhash"
```
NicolasMahe: You are using this new package introduced by https://github.com/mesg-foundation/core/pull/730 but not merged yet.

database/service_db.go
krhubert: > Why removing this test?
Because core dosen't autogenerate (the previous generated sid was invliad domain)

> The db still save the sid:
I pushed check for sid == empty

execution/execution.go
```diff
@@ -69,7 +69,7 @@ func New(service *service.Service, eventID string, taskKey string, inputs map[st
 	}
 
 	h := structhash.Sha1(exec, hashVersion)
-	exec.ID = hex.EncodeToString(h[:])
+	exec.ID = hex.EncodeToString(h)
```
NicolasMahe: Could you change ID type to `[]byte` while your at it ;) ?

service/service.go
```diff
@@ -1,272 +1,390 @@
...
-}
-
-func (e ErrNotDefinedEnv) Error() string {
-	return fmt.Sprintf("environment variable(s) %q not defined in mesg.yml (under configuration.env key)",
-		strings.Join(e.env, ", "))
+// volumeKey creates a key for service's volume based on the sid to make sure that the volume
+// will stay the same for different versions of the service.
+func volumeKey(sid, dependency, volume string) string {
+	h := sha1.New()
+	h.Write([]byte(sid))
```
NicolasMahe: The sid that is now optional is still used here.
I suggest to add a fallback to hash if sid is empty, but hash need to be passed as parameter as well.

```suggestion
	if sid != "" {
		h.Write([]byte(sid))
	} else {
		h.Write([]byte(hash))
	}
```

NicolasMahe:
Hey guys, I used a new tool to review this PR file by file, I really recommend you to use it also: https://reviewable.io/reviews/mesg-foundation/core/799#-

api/api.go
```diff
@@ -1,39 +1,155 @@
...
+}
+
+// GetService returns service serviceID.
+func (api *API) GetService(service string) (*service.Service, error) {
+	s, err := api.db.Get(service)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := api.sm.Status(s); err != nil {
```
NicolasMahe: Just as a note for future development, the Status should be saved and update directly in the service database.

service/manager.go
```diff
@@ -0,0 +1,294 @@
...
+
+var containerToServiceStatus = map[container.StatusType]Status{
+	container.UNKNOWN:  StatusUnknown,
+	container.STOPPED:  StatusStopped,
+	container.STARTING: StatusStarting,
+	container.RUNNING:  StatusRunning,
+}
+
+// ContainerManager is service manager that uses docker container
+// as platfor for deploy, start, ... services.
```
NicolasMahe: Typo on `platfor` no?

x/xvalidator/validator.go
```diff
@@ -35,3 +36,8 @@ func IsPortMapping(fl validator.FieldLevel) bool {
 	}
 	return true
 }
+
+// IsEnv validates if given field is valid env variable declaration.
+func IsEnv(fl validator.FieldLevel) bool {
```
NicolasMahe: Comment from @krhubert https://github.com/mesg-foundation/core/pull/821#pullrequestreview-215347607

> It will be good to check env name (for special chars etc)
> https://stackoverflow.com/questions/2821043/allowed-characters-in-linux-environment-variable-names

service/testdata/mesg.invalid.yml
```diff
@@ -0,0 +1,52 @@
...
+      - v1
+      - v1 # not unique
+    volumesFrom:
+      - none # not existing
+    ports: # invalid ports
+      - 0
+      - 65536
+      - 1:a
+      - a:1
+    env: # invalid env
```
NicolasMahe: Comment from @krhubert https://github.com/mesg-foundation/core/pull/821#pullrequestreview-215347607

> Do we need this check twice?

service/manager.go
krhubert: platform

api/api.go
antho1404: not in the service database, we should have another database/resource that has the records `serviceHash => status`

the service database should contains only the data of the services not their states

commands/provider/service_provider.go
```diff
@@ -198,39 +200,16 @@ func (p *ServiceProvider) ServiceStop(id string) error {
...
-		return fmt.Sprintf("%s Dockerfile does not exist", pretty.FailSign), nil
-	}
-	if !validation.IsValid() {
-		return fmt.Sprintf("%s Service is not valid", pretty.FailSign), nil
-	}
-
-	return fmt.Sprintf(`%s Dockerfile exists
-%s mesg.yml is valid
-%s Service is valid`, pretty.SuccessSign, pretty.SuccessSign, pretty.SuccessSign), nil
+	return fmt.Sprintf(`%s mesg.yml is valid`, pretty.SuccessSign), nil
```
NicolasMahe: Validate function should also check the Dockerfile

krhubert: No it shoudn't, because this is diffrent layer. 

Dockerfile is strictly related to Docker container and the service shouldn't be aware how the given manager starts a container. If you have multiple options to deploy service then `service` package shouldn't be responsible for validation if the service could be started. In that case, ContainerManager should be responsible and maybe we need to add `Validate(contextPath) error)` method to `Manager` Interface, but not to service itself.

service/definition/service.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+func toServiceDependency(key string, dep *Dependency) *service.Dependency {
+	return &service.Dependency{
+		Key:         key,
+		Image:       dep.Image,
+		Volumes:     dep.Volumes,
+		VolumesFrom: dep.VolumesFrom,
+		Ports:       dep.Ports,
+		Command:     dep.Command,
```
NicolasMahe: `Args` and `Env` are missing

protobuf/coreapi/api.proto
```diff
@@ -465,6 +465,7 @@ message Service {
   string description = 2;                   // Service's description.
   repeated Task tasks = 5;                  // The list of tasks this service can execute.
   repeated Event events = 6;                // The list of events this service can emit.
+  Dependency configuration = 13;            // The Docker Core.
```
NicolasMahe:
```suggestion
  Dependency configuration = 13;            // The Docker configuration of the service.
```

service/definition/service.go
```diff
@@ -14,20 +18,20 @@ type ServiceDefinition struct {
...
 
 	// Events are the list of events that service can emit.
 	Events map[string]*Event `yaml:"events" json:"events,omitempty" validate:"dive,keys,printascii,endkeys,required"`
 
-	// Dependencies are the Docker containers that service can depend on.
-	Dependencies map[string]*Dependency `yaml:"dependencies" json:"dependencies,omitempty" validate:"dive,keys,printascii,ne=service,endkeys,required"`
-
 	// Configuration is the Docker container that service runs inside.
-	Configuration *Dependency `yaml:"configuration" json:"configuration,omitempty"`
+	Configuration Dependency `yaml:"configuration" json:"configuration,omitempty"`
```
NicolasMahe: Is there a reason to remove the pointer here but keep them on Tasks, Events, etc..?

Could we remove all pointer?

service/definition/service.go
```diff
@@ -38,7 +42,7 @@ type Event struct {
 	// Description is the description of event.
 	Description string `yaml:"description" json:"description,omitempty" validate:"printascii"`
 
-	// Data holds the input parameters of event.
+	// Data holds the input inputs of event.
```
NicolasMahe:
```suggestion
	// Data holds the parameters of event.
```

service/definition/service.go
```diff
@@ -52,7 +56,7 @@ type Dependency struct {
 	Volumes []string `yaml:"volumes" json:"volumes,omitempty" validate:"unique,dive,printascii"`
 
 	// VolumesFrom are the docker volumes-from from.
-	VolumesFrom []string `yaml:"volumesfrom" json:"volumesFrom,omitempty" validate:"unique,dive,printascii"`
+	VolumesFrom []string `yaml:"volumesFrom" json:"volumesFrom,omitempty" validate:"unique,dive,printascii"`
```
NicolasMahe:
```suggestion
	VolumesFrom []string `yaml:"volumesfrom" json:"volumesFrom,omitempty" validate:"unique,dive,printascii"`
```

This is not a typo (or was?) and should not be changed in this PR because it's a breaking change.

service/definition/service.go
```diff
@@ -75,7 +79,7 @@ type Task struct {
 	// Description is the description of task.
 	Description string `yaml:"description" json:"description,omitempty" validate:"printascii"`
 
-	// Inputs are the definition of the execution inputs of task.
+	// Parameters are the definition of the execution inputs of task.
```
NicolasMahe:
```suggestion
	// Inputs are the definition of the execution inputs of task.
```

service/definition/service.go
```diff
@@ -90,28 +94,112 @@ type Output struct {
 	// Description is the description of task output.
 	Description string `yaml:"description" json:"description,omitempty" validate:"printascii"`
 
-	// Data holds the output parameters of a task output.
+	// Data holds the output inputs of a task output.
```
NicolasMahe:
```suggestion
	// Data holds the output parameters of a task output.
```

service/definition/service.go
```diff
@@ -90,28 +94,112 @@ type Output struct {
...
 	Data map[string]*Parameter `yaml:"data" json:"data,omitempty" validate:"required,dive,keys,printascii,endkeys,required"`
 }
 
-// Parameter describes task input parameters, output parameters of a task
-// output and input parameters of an event.
+// Parameter describes task input inputs, output inputs of a task
```
NicolasMahe: why replacing ALL `parameters` by `inputs`?

It doesn't make sense to me.

service/manager.go
```diff
@@ -0,0 +1,301 @@
...
+
+// Manager is an interface that defines every service manager.
+type Manager interface {
+	// Deploy deploys the service according to specific manager.
+	Deploy(service *Service, contextDir string, env map[string]string) error
+	// Start runs the service and all its dependencies.
+	Start(service *Service) error
+	// Stop stops the proviously started service and all its Dependencies.
+	Stop(service *Service) error
+	// Delete deletes all persisent data used by service.
```
NicolasMahe:
```suggestion
	// Delete deletes all persistent data used by service.
```

service/definition/service.go
NicolasMahe: Done in https://github.com/mesg-foundation/core/pull/824

service/definition/service.go
NicolasMahe: Done in https://github.com/mesg-foundation/core/pull/824

service/manager.go
NicolasMahe: Done in https://github.com/mesg-foundation/core/pull/824

protobuf/coreapi/api.proto
NicolasMahe: Done in https://github.com/mesg-foundation/core/pull/825

service/definition/service.go
NicolasMahe: Done in https://github.com/mesg-foundation/core/pull/825

service/definition/service.go
NicolasMahe: Done in https://github.com/mesg-foundation/core/pull/825

service/definition/service.go
NicolasMahe: Done in https://github.com/mesg-foundation/core/pull/825

NicolasMahe:
@antho1404 @ilgooz please review.

service/manager.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+	if err := m.c.DeleteNetwork(service.namespace()); err != nil {
+		return err
+	}
+	service.Status = StatusStopped
+	return nil
+}
+
+// Delete see Manager.Delete.
+func (m *ContainerManager) Delete(service *Service) error {
```
antho1404: Maybe `DeleteVolumes`. This doesn't delete the service so that's confusing

krhubert: So this is the implementation of the `Manager` interface

```go
  // Delete deletes all persistent data used by service.
  Delete(service *Service) error
```

So for Container, it happens to delete volumes but for other deploy methods, it could be something else.

service/manager.go
```diff
@@ -0,0 +1,301 @@
...
+			Ports:  dep.ports(),
+			Networks: []container.Network{
+				{ID: networkID, Alias: dep.Key},
+				{ID: sharedNetworkID},
+			},
+		})
+		return err
+	}
+
+	if err := start(service.Configuration, service.namespace()); err != nil {
```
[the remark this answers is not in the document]
krhubert: Can be moved to start at the end.

database/service_db.go
```diff
@@ -115,10 +114,12 @@ func (d *LevelDBServiceDB) delete(tx *leveldb.Transaction, hashOrSid string) err
 	if err != nil {
 		return err
 	}
-	if err := tx.Delete([]byte(hashKeyPrefix+s.Hash), nil); err != nil {
-		return err
+	if s.Sid != "" {
```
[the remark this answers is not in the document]
krhubert: So they are not mandatory right now (they are marked as optional).

commands/service_execute.go
```diff
@@ -186,6 +173,15 @@ func (c *serviceExecuteCmd) readFile() (string, error) {
 	return string(content), err
 }
 
+func taskByKey(s *coreapi.Service, key string) *coreapi.Task {
```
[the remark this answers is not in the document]
krhubert: we can but we can check for the null as well, it's just a helper function only used inside this file.

api/deploy.go
```diff
@@ -72,168 +29,96 @@ type DeployStatus struct {
...
+			return nil, err
+		}
+	} else {
+		// if not git repo then it must be tarball
+		resp, err := http.Get(url)
+		if err != nil {
+			return nil, err
+		}
+		defer resp.Body.Close()
+		if err := archive.Untar(resp.Body, contextDir, nil); err != nil {
```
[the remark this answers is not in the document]
krhubert: we wanted service to accept `contextDir` as arguments. So there are few cases here:

- api receives io.Reader that could be read only once
- api needs to `formalizeContext` (remove .git etc...) so it needs to extract the tar
- service needs a directory to calculate the hash
- service needs tar.Reader to pass it to docker deploy api

So combining all this either service will receive io.reader, extract tar, calc the hash and then archive it back, or the other way.

Dockerfile
```diff
@@ -13,9 +13,10 @@ RUN go build -o mesg-core \
       -ldflags="-X 'github.com/mesg-foundation/core/version.Version=$version'" \
       core/main.go
 
+# NOTE: docker uses xz-utils for uncompress tar.xz files
 FROM ubuntu:18.04
 RUN apt-get update && \
-      apt-get install -y --no-install-recommends ca-certificates=20180409 && \
+      apt-get install -y --no-install-recommends ca-certificates=20180409 xz-utils && \
```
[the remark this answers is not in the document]
krhubert: Because docker untar uses this for xz format (and we are using docker.Archvie package)

api/deploy.go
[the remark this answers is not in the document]
antho1404: that's not my point here, why not use `DeployService(resp.Body, ...)` and have a pseudo function like that
```
if isGit {
  dir = createTmp()
  gitClone(arg, dir)
  return deploy(dir)
}
resp = http.Get(arg)
return DeployService(resp.Body, ...)
```
Here we have only one function responsible for the untar, not 2 and nothing else changes and we still have the context

commands/service_execute.go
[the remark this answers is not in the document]
antho1404: just seems better to adopt this pattern all the time. Also we are returning the error anyway just after so let's get the error from the function itself

database/service_db.go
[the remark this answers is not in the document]
antho1404: What about volumes ? 2 services with no sid will share the same volume.
Right now they are not mandatory in the mesg.yml but generated with a value by default (`_HASH` if i remember correctly) so the definition file is optional but in the core mandatory.

This change breaks volumes for services

service/manager.go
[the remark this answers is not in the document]
antho1404: we could change the behavior if there is good reason but it makes more sense for me to start all dependencies first.

Dockerfile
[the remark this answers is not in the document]
antho1404: we are also using `docker.archive` on the dev without the `xz-utils`. It was added in https://github.com/mesg-foundation/core/pull/731 then removed in the same PR

protobuf/coreapi/api.proto
```diff
@@ -506,12 +507,12 @@ message Parameter {
...
 
-// A dependency is a configuration of an other Docker container that runs separately from the service.
+// A dependency is a configuration of an other container that runs separately from the service.
 message Dependency {
   string key = 8;                   // Dependency's key.
-  string image = 1;                 // Image's name of the Docker.
+  string image = 1;                 // Image's name of the container.
   repeated string volumes = 2;      // List of volumes.
-  repeated string volumesfrom = 3;  // List of volumes mounted from other dependencies.
+  repeated string volumesFrom = 3;  // List of volumes mounted from other dependencies.
```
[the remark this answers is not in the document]
NicolasMahe: It's fine here. The mesg.yml still have the `volumesfrom`.

service/manager.go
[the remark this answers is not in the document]
NicolasMahe: I agree with Anthony, let's start the dependencies first.

database/service_db.go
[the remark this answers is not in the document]
NicolasMahe: In service manager, if sid is empty it uses the hash.
The behavior is exactly like before.

protobuf/coreapi/api.proto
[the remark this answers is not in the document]
NicolasMahe: And no, it's not a breaking change, gRPC uses the number (in this case 3) to decode the data. So this is compatible with previous version.

service/manager.go
NicolasMahe: Instead of having the stop and delete logic in API, it could be done at the service manager.
Stop is just stopping the service.
Delete: stop + delete persistent data.

commands/service_execute.go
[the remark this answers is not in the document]
NicolasMahe: I feel it's not a really good pattern in Go to return a pointer without error.
I speak about this with Hubert about should we keep pointer in the data structure of service or not. We didn't find a good solution.
In both TypeScript and in Swift, they force to check if the variable is nil / undefined if it can be, before accessing it. This is not something that the Go compiler enforce. I feel it opens the place for many nil pointer error...

What about removing all pointers in the struct of the service? Like this, for example, this function `taskByKey`, will need to have the following signature: `taskByKey(s coreapi.Service, key string) (coreapi.Task, error)`.
Yeah it's will consume more memory, but we are forced to follow a good pattern.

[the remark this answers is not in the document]
antho1404: the pointer discussion is something else, the problem will be the same. Either we test something based on the value returned or based on an error.
After I kind of agree if it's pure data we don't necessary need pointers as long as it's not huge data and not really important that are present everywhere in all the modules. For this one I don't mind keeping the pointers. I don't see any really good reason to change it

database/service_db.go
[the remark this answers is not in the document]
antho1404: great I didn't see that. Also behavior is not exactly, we are taking the hash, before we where appending an character in front of the hash to make sure that we cannot create a service with the hash of another one as sid and then overwrite the previous service/volume. But anyway that was not a strong security anyway.

I still feel more confident to make sure that we can rely on a service to have a sid (that is generate when we deploy it like the hash) instead of having to check all the time we want to use the sid. I'm not necessary against it either it's just something I would like to have agreement to see the pro and con on that and not change something that impact the data.

protobuf/coreapi/api.proto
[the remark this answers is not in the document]
antho1404: We are changing the data so it is a breaking change if any client like cli for example are reading this property which is quite unlikely but still this is a breaking change

api/deploy.go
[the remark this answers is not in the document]
krhubert: To send status about `Service downloaded`

Dockerfile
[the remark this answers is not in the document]
krhubert: Yes we had a bug because this package is needed to support xz compression while using docker.archive

service/manager.go
krhubert: No, it's manager part to have deleted and stop function separated. so api (or any other packge) can use them, and there they should be splited.

event/event.go
```diff
@@ -15,25 +14,12 @@ type Event struct {
...
-func Create(s *service.Service, eventKey string, eventData map[string]interface{}) (*Event, error) {
-	event, err := s.GetEvent(eventKey)
-	if err != nil {
-		return nil, err
-	}
-	if err := event.RequireData(eventData); err != nil {
-		return nil, err
-	}
+// New creates an event eventKey with eventData for given service.
+func New(s *service.Service, eventKey string, eventData map[string]interface{}) *Event {
```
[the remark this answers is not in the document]
krhubert: It was moved to api package (so api handles all validation) and in that case event and execution packages depend only on service structure (and not the way how to validate them). So both package can take care only of it's logic (like update complete status for execution)

commands/service_execute.go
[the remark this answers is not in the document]
krhubert: we can change to error, it's really small change (but error won't ever happen here) but it's ok

database/service_db.go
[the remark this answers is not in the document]
krhubert: In that case, we have to generate different sid because one we generated one is an invalid domain. Also, sid should be used only for local development and should not be mandatory.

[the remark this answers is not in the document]
antho1404: sid is the base for the markeplace too so it's not only about local development. We put sid as optional with generation in order not to break previous services but every services needs to have a sid to identify this service and "group" the different version to the same service.

How do you manage services without sid on the marketplace for example ?

event/event.go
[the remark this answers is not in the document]
antho1404: I saw that this was moved on the API but what happen if we use the event/execution packages directly without the API ? We might introduce invalid data ? Also this doesn't reduce any dependencies as event/execution packages already depends on the service package.

NicolasMahe:
Closing this PR as it's not relevant anymore

PR #801, opened 2019-03-08
Add Hash to reply of Deploy API

NicolasMahe:
Needed for https://github.com/mesg-foundation/core/pull/798


protobuf/coreapi/api.proto
```diff
@@ -243,10 +243,17 @@ message DeployServiceReply {
     Type type = 2;      // Type of status message.
   }
 
+  message Service {
+    string SID = 1;   // `SID` is the service ID.
+    string hash = 2;  // `hash` is the service hash.
+  }
+  
+  reserved 1, 3;
```
NicolasMahe: We should use this `reserved` everytime we remove a parameter from a proto definition.
It prevents future use of the number.

krhubert: I think we could use reserved when we reach alpha or beta stage, for now, let's make use of where we are right now :) But in the future we should use `reserved` of course

protobuf/coreapi/api.proto
```diff
@@ -243,10 +243,17 @@ message DeployServiceReply {
     Type type = 2;      // Type of status message.
   }
 
+  message Service {
```
krhubert: I don't like that we define the second time Service message in the same package.

In fact, this replay got quite bolted.

We should replace `validationError` with just return an error from deploy method (don't remember why we need this special kind of error).

In new #799 PR we only send 2 statuses about deploying so we can think of replacing this status api to something more general than any of the methods can use if wanted.

Applying the above issues will result in simple replay:
```
message DeployServiceReply {
  string sid = 1;   // `sid` is the service ID.
  string hash = 2;  // `hash` is the service hash.
}
```

NicolasMahe: As agreed with Hubert, it will be improved in PR refactor

commands/service_deploy.go
```diff
@@ -68,8 +68,8 @@ func (c *serviceDeployCmd) runE(cmd *cobra.Command, args []string) error {
 			errors.New("to get more information, run: mesg-core service validate"),
 		}
 	}
-	fmt.Printf("%s Service deployed with id: %v\n", pretty.SuccessSign, pretty.Success(id))
-	fmt.Printf("To start it, run the command:\n\tmesg-core service start %s\n", id)
+	fmt.Printf("%s Service deployed with sid %v and hash %s\n", pretty.SuccessSign, pretty.Success(sid), hash)
```
ilgooz:
```suggestion
	fmt.Printf("%s Service deployed with sid %s and hash %s\n", pretty.SuccessSign, pretty.Success(sid), pretty.Success(hash))
```

better coloring

commands/service_dev.go
```diff
@@ -78,11 +78,12 @@ func (c *serviceDevCmd) runE(cmd *cobra.Command, args []string) error {
 			errors.New("to get more information, run: mesg-core service validate"),
 		}
 	}
-	fmt.Printf("%s Service deployed with ID: %v\n", pretty.SuccessSign, pretty.Success(id))
+	fmt.Printf("%s Service deployed with sid %v and hash %s\n", pretty.SuccessSign, pretty.Success(sid), hash)
```
ilgooz:
```suggestion
	fmt.Printf("%s Service deployed with sid %s and hash %s\n", pretty.SuccessSign, pretty.Success(sid), pretty.Success(hash))
```

better coloring

interface/grpc/core/deploy_integration_test.go
```diff
@@ -17,9 +17,10 @@ func TestIntegrationDeployService(t *testing.T) {
 	stream := newTestDeployStream(url)
 
 	require.Nil(t, server.DeployService(stream))
-	defer server.api.DeleteService(stream.serviceID, false)
+	defer server.api.DeleteService(stream.sid, false)
```
antho1404:
```suggestion
	defer server.api.DeleteService(stream.hash, false)
```
let's keep hash everywhere

interface/grpc/core/list_services_test.go
```diff
@@ -15,7 +15,7 @@ func TestListServices(t *testing.T) {
 
 	stream := newTestDeployStream(url)
 	require.NoError(t, server.DeployService(stream))
-	defer server.api.DeleteService(stream.serviceID, false)
+	defer server.api.DeleteService(stream.sid, false)
```
antho1404:
```suggestion
	defer server.api.DeleteService(stream.hash, false)
```

NicolasMahe:
@mesg-foundation/core modif done

protobuf/coreapi/api.proto
antho1404: Agree with @krhubert here, let's not add complexity for now especially that this will be the only api with that. We should do it everywhere or nowhere to have consistency. If we want to do this can be done in another PR for all the apis. I would remove this part for this PR.

NicolasMahe:
```suggestion
```

NicolasMahe: Removed.

NicolasMahe:
@mesg-foundation/core `reserved` keyword removed from proto and dev merged. please review.

PR #813, opened 2019-03-14
Make service provider use provider client

NicolasMahe:


NicolasMahe:
Merging after fixing the mocks

PR #814, opened 2019-03-15
Set stop timeout before docker kill the container to 10sec, default value of docker.

NicolasMahe:
This pr change the timeout between stopping a container and killing the container from 1 to the default value of docker: 10 sec.


container/service.go
```diff
@@ -94,7 +94,7 @@ func (c *DockerContainer) deletePendingContainer(namespace []string) error {
 	// This hack for Docker to stop and then remove the container.
 	// See issue https://github.com/moby/moby/issues/32620
 	if container.ContainerJSONBase != nil {
-		timeout := 1 * time.Second
+		timeout := 10 * time.Second // 10 second is the default value docker use
```
krhubert:
```suggestion
		c.client.ContainerStop(ctx, container.ID, nil)
```

If you wan't the default value then pass nil to ContainerStop (this is why `time.Duration` is pointer here.

Link to docs - https://github.com/moby/moby/blob/8e610b2b55bfd1bfa9436ab110d311f5e8a74dcb/client/container_stop.go#L15

Otherwise if you want to change docker default value what's the motivation behind?

NicolasMahe: 👍

container/container.go
```diff
@@ -54,7 +54,7 @@ type Option func(*DockerContainer)
 // New creates a new Container with given options.
 func New(options ...Option) (*DockerContainer, error) {
 	c := &DockerContainer{
-		callTimeout: 10 * time.Second,
+		callTimeout: 60 * time.Second,
```
NicolasMahe: I changed this  value because some test always reach this timeout.
Anyone know why with have this timeout setup in the first place?

krhubert: Probably to reduce default timeout which may be from 180 up to 360 sec as I remember

NicolasMahe: Actually it makes a lot of test failed or timeout... what is happening?

antho1404:
It there a way to reproduce the bug that this PR is solving ? I don't see the need of changing anything here.
I would like to really understand and reproduce the problem before jumping on a quick fix (maybe not a quick fix but I can't actually reproduce any problem and don't really understand the issue)

NicolasMahe:
> It there a way to reproduce the bug that this PR is solving ? I don't see the need of changing anything here.
> I would like to really understand and reproduce the problem before jumping on a quick fix (maybe not a quick fix but I can't actually reproduce any problem and don't really understand the issue)

Run the Core with this branch (don't merge https://github.com/mesg-foundation/core/pull/816), start a bunch of service, stop the Core (without stopping the service), list docker service, see that there is a bunch of them still running.

The reason is from 2 things:
- This 1 sec timeout between stopping and killing the container.
- Core takes more than 1 sec to stop thus get killed before stopping all services.
PR https://github.com/mesg-foundation/core/pull/816 makes the Core way faster to stop, but still more than 1 sec.

container/network.go
```diff
@@ -54,9 +54,7 @@ func (c *DockerContainer) DeleteNetwork(namespace []string, event EventType) err
 	if err != nil {
 		return err
 	}
-	ctx, cancel := context.WithTimeout(context.Background(), c.callTimeout)
-	defer cancel()
-	messageChan, errChan := c.client.Events(ctx, types.EventsOptions{
```
NicolasMahe: This was causing the test to fail: a stream with a timeout in the context...

PR #815, opened 2019-03-15
stop service on core stop in cli

NicolasMahe:
Revert https://github.com/mesg-foundation/core/pull/792 because the way https://github.com/mesg-foundation/core/pull/790 stop the services is not good enough.
Also removing a function that is not used anymore.

Docker, by default, gives only 10 sec, and before https://github.com/mesg-foundation/core/pull/814, only 1 sec, between stop a container and killing it. The core is taking more than 10 sec to stop services when there is more than just 1 running service.


krhubert:
What do you mean by 
> stop the services is not good enough.

Are there any bugs, bottlenecks, etc...?

NicolasMahe:
> What do you mean by
> 
> > stop the services is not good enough.
> 
> Are there any bugs, bottlenecks, etc...?

Yes, because Docker kills container only after 10sec by default, and with current Core config, only 1 sec. So the Core doesn't have enough time to properly stop the services. The cli should do it before stopping the core.
I also leave the removal of service in the Core in case the user delete remove the docker service Core manually, in this case, the core has 10sec to remove all services. That's not perfect, but should be enough if just 1 service is running.

krhubert:
So to me the fix should be done inside the core (not by moving this into cli). Cli shoudn't manage stoping services in case of stoping core.

NicolasMahe:
It's actually doing both.

The problem is the core has a limited time (that we don't necessary control), by default 10sec, to delete all services and stop itself.
I think stopping service in both cli and core is the safest solution.

krhubert:
I think only core should be responsible to properly stop the services because the core is the one starting them. And we shouldn't move this logic into cli (someone can just send grpc request to stop the core and it should work) . Cli is just easy to use command - nothing else but to solution

https://github.com/moby/moby/blob/8e610b2b55bfd1bfa9436ab110d311f5e8a74dcb/client/container_stop.go#L15

Can't we just increase the core container stop timeout?

antho1404:
Totally agree too, the cli shouldn't be responsible for this, the core should.
Also the core don't need to wait for the different services to stop. Maybe we can just send the instruction to delete all docker services and shutdown the api, this will be fast.

NicolasMahe:
> someone can just send grpc request to stop the core and it should work

There is no gRPC API to stop the core. The CLI directly ask Docker to delete the docker service of the Core. Maybe we should have a stop gRPC API... It could actually help with the problem.

> Can't we just increase the core container stop timeout? Or just find a way to move it to the core, because cli shouldn't do this at all.

> I agree with @krhubert, let's keep this logic inside Core and make stop timeout configurable and extend timeout for to a few minutes for Core's container.

As said previously, the CLI is just telling Docker to delete the docker service of the Core. It means that we have control on the timeout ONLY in the CLI.
If a user directly execute `docker service remove XXX` then the timeout will be 10 sec (default value).
So basically, the Core has 10 sec (by default) to gracefully stop all services + shutdown before getting killed.

I already improve the time needed to stop all services with PR https://github.com/mesg-foundation/core/pull/816 but I feel in some case it will be still too long and some services will still be running after the killing of the Core.

NicolasMahe:
Closing because of fixes of https://github.com/mesg-foundation/core/pull/820

PR #820, opened 2019-03-16
Fix container package

NicolasMahe:
This PR fixes and simplifies a few stuff on the package container:

- Introduce the option StopGracePeriod.
This option tells docker the time to wait between the service is removed and killing the associated container. Now it's possible to set a custom on when starting a docker service.

- Upgrade function `deletePendingContainer` to wait for a grace period before removing the container.
It will either use the grace period is set in the docker service data or use the default one (10sec) if the first is not set.
The previous logic of `deletePendingContainer` was killing and removing the container after only a few second, without anyway to change the time (timeout of containerStop was somehow not taking into account, maybe because of docker service).

- Simplifying / improving function `DeleteNetwork`.
The `DeleteNetwork` was using Docker.Events function to listen for network deletion event with a context with timeout. The problem is, sometime, the network is deleted either too early or too late (after the context timeout), so the `DeleteNetwork` was falling in such case.
I update the code to use a similar logic that the function `deletePendingContainer`, a recursive call that check if the network is actually deleted first.

- Change default `container.callTimeout` to 60sec.

### DIFF
--- a/container/service.go
+++ b/container/service.go
@@ -67,8 +67,6 @@ func (c *DockerContainer) StopService(namespace []string) error {
 	if status == STOPPED {
 		return nil
 	}
-	ctx, cancel := context.WithTimeout(context.Background(), c.callTimeout)
-	defer cancel()
 	service, err := c.FindService(namespace)
 	if err != nil && !docker.IsErrNotFound(err) {
 		return err
@@ -77,7 +75,7 @@ func (c *DockerContainer) StopService(namespace []string) error {
 	if service.Spec.TaskTemplate.ContainerSpec != nil && service.Spec.TaskTemplate.ContainerSpec.StopGracePeriod != nil {
 		stopGracePeriod = *service.Spec.TaskTemplate.ContainerSpec.StopGracePeriod
 	}
-	ctx, cancel = context.WithTimeout(context.Background(), c.callTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), c.callTimeout)
 	defer cancel()
 	if err := c.client.ServiceRemove(ctx, c.Namespace(namespace)); err != nil && !docker.IsErrNotFound(err) {
 		return err
--- a/container/service.go
+++ b/container/service.go
@@ -67,38 +67,46 @@ func (c *DockerContainer) StopService(namespace []string) error {
 	if status == STOPPED {
 		return nil
 	}
-
 	ctx, cancel := context.WithTimeout(context.Background(), c.callTimeout)
+	defer cancel()
+	service, err := c.FindService(namespace)
+	if err != nil && !docker.IsErrNotFound(err) {
+		return err
+	}
+	stopGracePeriod := c.defaultStopGracePeriod
+	if service.Spec.TaskTemplate.ContainerSpec != nil && service.Spec.TaskTemplate.ContainerSpec.StopGracePeriod != nil {
+		stopGracePeriod = *service.Spec.TaskTemplate.ContainerSpec.StopGracePeriod
+	}
+	ctx, cancel = context.WithTimeout(context.Background(), c.callTimeout)
 	defer cancel()
 	if err := c.client.ServiceRemove(ctx, c.Namespace(namespace)); err != nil && !docker.IsErrNotFound(err) {
 		return err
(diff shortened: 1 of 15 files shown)


antho1404:
Why removing the part to listen when the network is actually deleted ?

We need to ensure that the network is deleted when we remove the service otherwise we will have some problems again. Also adding 1s a bit everywhere in the code is not a good solution, sometime it might take more that 1s to delete the network/container/service

NicolasMahe:
> Why removing the part to listen when the network is actually deleted ?
> 
> We need to ensure that the network is deleted when we remove the service otherwise we will have some problems again. Also adding 1s a bit everywhere in the code is not a good solution, sometime it might take more that 1s to delete the network/container/service

Did you see the explanation about this in the PR description? Do you want more explanation?

> The `DeleteNetwork` was using Docker.Events function to listen for network deletion event with a context with timeout. The problem is, sometime, the network is deleted either too early or too late (after the context timeout), so the `DeleteNetwork` was falling in such case.
> I update the code to use a similar logic that the function `deletePendingContainer`, a recursive call that check if the network is actually deleted first.

antho1404:
Sorry i missed the recursive call. It's better but I feel having a polling or any kind of solutions like that is not so good when we can have events. Why not just using a `context.Background` and remove the timeout and still keep the event based system ?

NicolasMahe:
> Sorry i missed the recursive call. It's better but I feel having a polling or any kind of solutions like that is not so good when we can have events. Why not just using a `context.Background` and remove the timeout and still keep the event based system ?

That's what I tried first and it doesn't solve the issue.
The problem is I guess some event are fired before the listener is actually ready (same problem we had and solve with the gRPC acknowledgement header system) or we don't listen for the right event, or the event is never trigger under certain condition..

container/service.go
```diff
@@ -80,24 +72,20 @@ func (c *DockerContainer) StopService(namespace []string) error {
...
 	// Sometime, the ServiceRemove function doesn't remove the associated containers.
 	// This hack for Docker to stop and then remove the container.
 	// See issue https://github.com/moby/moby/issues/32620
-	if container.ContainerJSONBase != nil {
-		timeout := 1 * time.Second
-		c.client.ContainerStop(ctx, container.ID, &timeout)
-		c.client.ContainerRemove(ctx, container.ID, types.ContainerRemoveOptions{})
-	}
+	c.client.ContainerStop(context.Background(), container.ID, nil)
+	c.client.ContainerRemove(context.Background(), container.ID, types.ContainerRemoveOptions{})
```
NicolasMahe: Can try to delay the call to ContainerStop & ContainerRemove manually.

NicolasMahe:
- [ ] Put back timeout

NicolasMahe:
@mesg-foundation/core Modification done, please review

container/service.go
```diff
@@ -67,38 +67,44 @@ func (c *DockerContainer) StopService(namespace []string) error {
...
+	if err := c.deletePendingContainer(namespace, time.Now().Add(stopGracePeriod)); err != nil {
 		return err
 	}
 	return c.waitForStatus(namespace, STOPPED)
 }
 
-func (c *DockerContainer) deletePendingContainer(namespace []string) error {
-	ctx, cancel := context.WithTimeout(context.Background(), c.callTimeout)
-	defer cancel()
+func (c *DockerContainer) deletePendingContainer(namespace []string, maxGraceTime time.Time) error {
```
ilgooz: We can simplify time logic by using [<-time.After(stopGracePeriod)](https://golang.org/pkg/time/#After).

NicolasMahe: I don't see how it will simplify.

The goal here is to "return" as soon as the container is not found.
The maxGraceTime is kind of a timeout.

I think using `<-time.After(stopGracePeriod)` will always to call `ContainerRemove` or to prevent this, another chan done need to be created to tell the go routine to stop. So actually it seems more complex.

ilgooz: yes, i guess current implementation is fine because after we remove the service and check if the container exists right after, Docker still might be in the deleting process of container. So it's good that we don't block for the whole timeout and check the status of container again in short periods.

PR #821, opened 2019-03-17
Service refactor nico suggestion

NicolasMahe:
My suggestion for PR #799 

This change is [Reviewable](https://reviewable.io/reviews/mesg-foundation/core/821)


x/xvalidator/validator.go
```diff
@@ -35,3 +36,8 @@ func IsPortMapping(fl validator.FieldLevel) bool {
 	}
 	return true
 }
+
+// IsEnv validates if given field is valid env variable declaration.
+func IsEnv(fl validator.FieldLevel) bool {
```
krhubert: It will be good to check env name (for special chars etc) 

https://stackoverflow.com/questions/2821043/allowed-characters-in-linux-environment-variable-names

service/testdata/mesg.invalid.yml
```diff
@@ -41,6 +44,8 @@ dependencies:
       - 65536
       - 1:a
       - a:1
+    env: # invalid env
```
krhubert: Do we need this check twice?

PR #822, opened 2019-03-17
add cache for go mod. improve cache for go.

NicolasMahe:


.circleci/config.yml
```diff
@@ -8,47 +8,65 @@ runOnDocker: &runOnDocker
   docker:
     - image: circleci/golang:1.11.4
 
+restoreGoModCache: &restoreGoModCache
```
krhubert: it would be nice to not mix camelCase with underscores

everywhere else in config.yml `_` notation is used so it will be goo to change names

NicolasMahe: That's not totally true. All yml ref are camelCase (Eg: `&runOnMachine`).
But I agree, it should be only one style. I will update.

NicolasMahe: Done

antho1404:
We talked about that in a previous PR
https://forum.mesg.com/t/use-gocache-in-circleci/173

Don't we have a problem with code coverage with this modification ? Or is it fixed ?

krhubert:
It's not yet fixed but we can safely add cache:

- cache for downloading module (we removed vendor dir)
- cache for a build (when code coverage will be fixed it will be used automatically) - here golanglint-ci may use it.

PR #823, opened 2019-03-18
Improvements of commands service init and service gendoc

NicolasMahe:


commands/service_docs.go
```diff
@@ -26,15 +26,17 @@ func newServiceDocsCmd(e ServiceExecutor) *serviceDocsCmd {
 		Use:   "gen-doc",
 		Short: "Generate the documentation for the service in a README.md file",
 		Example: `mesg-core service gen-doc
-mesg-core service gen-doc ./PATH_TO_SERVICE`,
+mesg-core service gen-doc ./PATH_TO_SERVICE
+mesg-core service gen-doc --force`,
```
ilgooz:
```suggestion
mesg-core service gen-doc --force ./PATH_TO_SERVICE`,
```

commands/service_init.go
```diff
@@ -36,22 +35,21 @@ func newServiceInitCmd(e ServiceExecutor) *serviceInitCmd {
 	
 To get more information, see the page [service file from the documentation](https://docs.mesg.com/guide/service/service-file.html)`,
 		Example: `mesg-core service init
-mesg-core service init --name NAME
-mesg-core service init --current`,
-		Args:    cobra.NoArgs,
+mesg-core service init ./PATH_TO_SERVICE
+mesg-core service init --template TEMPLATE_URL`,
```
ilgooz:
```suggestion
mesg-core service init --template TEMPLATE_URL ./PATH_TO_SERVICE`,
```

commands/service_docs.go
NicolasMahe: The path is optional, that's why it is only show once.

commands/service_init.go
NicolasMahe: The path is optional, that's why it is only show once.

commands/service_docs.go
```diff
@@ -26,15 +26,17 @@ func newServiceDocsCmd(e ServiceExecutor) *serviceDocsCmd {
...
 		PreRunE: c.preRunE,
 		RunE:    c.runE,
 	})
+	c.cmd.Flags().BoolVarP(&c.force, "force", "f", c.force, "No confirmation. Will replace existing README.md file")
```
antho1404: I don't see the use of this flag in the command

NicolasMahe: https://github.com/mesg-foundation/core/pull/823/files#diff-aaa5128dba1f3dafaedf0d439b020430R41

NicolasMahe: The logic was already implemented but the flag was not exposed

ilgooz: I think having force option is good for non interactive terminals. But we can change it to `--yes` because we use it in general, in other command too. It exists for answering all terminal prompts as yes by default in a command action. But this improvement can be done in an another PR.

PR #828, opened 2019-03-19
Service marketplace: encode / decode hash to base58

NicolasMahe:


systemservices/marketplace/src/contracts/utils.ts
```diff
@@ -24,13 +25,11 @@ const fromUnit = (x: string|BigNumber) => new BigNumber(x).dividedBy(1e18)
...
 const parseTimestamp = (x: string) => new Date(new BigNumber(x).times(1000).toNumber())
 
-const hashToHex = (x: string) => {
-  if (x.startsWith('0x')) {
-    return x
-  }
-  return '0x' + x
+const hashToHex = (x: string): string => {
+  if (x.startsWith('0x')) return x
+  return '0x' + base58.decode(x).toString('hex')
```
krhubert: `0x` is a prefix for base16. Function shouldn't print or accept hash starting with `0x` as it's invalid base58

NicolasMahe: I will return an error rather than returning the string 👍

NicolasMahe: Done

systemservices/marketplace/src/contracts/utils.ts
```diff
@@ -24,13 +25,11 @@ const fromUnit = (x: string|BigNumber) => new BigNumber(x).dividedBy(1e18)
...
-  if (x.startsWith('0x')) {
-    return x
-  }
-  return '0x' + x
+const hashToHex = (x: string): string => {
+  if (x.startsWith('0x')) throw new Error('hash format is invalid. It starts with 0x')
+  return '0x' + base58.decode(x).toString('hex')
 }
-const hexToHash = (x: string) => x.replace(/^0x/g, '')//.replace(/0*$/g, '')
+const hexToHash = (x: string): string => base58.encode(Buffer.from(x.replace(/^0x/g, ''), 'hex'))
```
ilgooz:
```suggestion
const hexToHash = (x: string): string => base58.encode(Buffer.from(x.replace(/^0x/, ''), 'hex'))
```

systemservices/marketplace/src/contracts/utils.ts
```diff
@@ -1,12 +1,13 @@
-import Web3 from "web3"
-import BigNumber from "bignumber.js";
-import Contract from "web3/eth/contract";
-import { TaskInputs } from "mesg-js/lib/service";
+import Web3 from 'web3'
+import BigNumber from 'bignumber.js';
+import Contract from 'web3/eth/contract';
+import { TaskInputs } from 'mesg-js/lib/service';
+const base58 = require('base-x')('123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz')
```
antho1404: Are we sure about that ?
It seems some character are missing.
`0` `I` `O` `l`
Seems obvious that is to avoid error with `0O` and `Il` but any reason to totally remove them ?

NicolasMahe: https://en.wikipedia.org/wiki/Base58

PR #829, opened 2019-03-19
systemservices/marketplace: manifest validation

ilgooz:
manifest data is created by users/devs. thus, its content is not trusted and it needs to be validated. this way, it's possible to ignore any service versions that have invalid manifest data.


antho1404:
Just seeing this PR I was thinking why not delegating more to the system service and have the publish command that actually take the content of the manifest in the `mesg.yml` (like that we have the verification from the core) and the marketplace take care of sending this to IPFS. That simplifies the logic on the CLI also we could have a link between the system service and the version that the system service can actually handle

systemservices/marketplace/src/contracts/manifest.ts
```diff
@@ -1,5 +1,16 @@
...
 import Request from "request-promise-native"
-import { Manifest } from "../types/manifest";
+import { resolve } from "path"
+import { Manifest } from "../types/manifest"
+
+// TODO(ilgooz) create a Definition TS type and generate this schema over it.
+import definitionSchema from '../types/definition.json'
+
+const program = TJS.getProgramFromFiles([resolve(__dirname, '../types/manifest.ts')]);
+const schema = TJS.generateSchema(program, "Manifest", { required: true });
```
NicolasMahe: I think the json schema should be generated once using the command:
```
typescript-json-schema <path-to-typescript-files-or-tsconfig> <type>
```
https://github.com/YousefED/typescript-json-schema


Like we are doing for protobuf files

package-lock.json
```diff
@@ -0,0 +1,50 @@
+{
```
NicolasMahe: please remove this useless file

systemservices/marketplace/src/contracts/manifest.ts
ilgooz: why? dealing with command would be an extra work, an another thing to maintain.

having it programmatically is simpler. schema is created and cached on the start time of the program so, there're no performance drawbacks anyway.

NicolasMahe: What about reusing the json schema in other app?
Having it as source of truth is good anyway.

The code will be simpler and easier to understand.

ilgooz: For reusing it somewhere else, it might make sense.

package-lock.json
NicolasMahe: You removed the `package-lock.json` of the marketplace service..

systemservices/marketplace/src/contracts/manifest.ts
NicolasMahe: Can you please update it.

ilgooz: done

systemservices/marketplace/src/contracts/manifest.ts
```diff
@@ -1,5 +1,10 @@
+import { Validator } from "jsonschema"
 import Request from "request-promise-native"
-import { Manifest } from "../types/manifest";
+import { Manifest } from "../types/manifest"
+import manifestSchema from '../types/schema/manifest.json'
+import definitionSchema from '../types/schema/definition.json'
```
NicolasMahe:
```suggestion
import serviceSchema from '../types/schema/service.json'
```

Could you rename definition for service? It's easier to guess what it is.

systemservices/marketplace/src/contracts/version.ts
```diff
@@ -12,7 +12,8 @@ const getServiceVersions = async (contract: Marketplace, sid: string): Promise<V
   const versionsLength = new BigNumber(await contract.methods.serviceVersionsLength(sidHex).call())
   const versionsPromise: Promise<Version|undefined>[] = []
   for (let j = new BigNumber(0); versionsLength.isGreaterThan(j); j = j.plus(1)) {
-    versionsPromise.push(getServiceVersionWithIndex(contract, sid, j))
+    const version = getServiceVersionWithIndex(contract, sid, j)
+    if (version) versionsPromise.push(version)
```
NicolasMahe: This is useless because it returns a promise, and 2 lines under, empty elements are removed from array

systemservices/marketplace/src/contracts/version.ts
```diff
@@ -30,6 +31,10 @@ const getServiceVersion = async (contract: Marketplace, versionHash: string): Pr
   }
   const version = await contract.methods.serviceVersion(versionHashHex).call()
   const manifestData = await getManifest(hexToAscii(version.manifestProtocol), hexToAscii(version.manifest))
+  if (!manifestData) {
+    console.warn('version', versionHash, 'is not a valid version')
```
NicolasMahe: I think this console.warn is useless. There is already 2 console.warn in getManifest function. It should be enough

systemservices/marketplace/gen-schema
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+./node_modules/typescript-json-schema/bin/typescript-json-schema \
+  -o src/types/schema/manifest.json \
+  --required \
+  src/types/manifest.ts Manifest
```
NicolasMahe: Can you move it inside the package.json in scripts array? Like this you don't need to specify the long path in the node_modules
```
typescript-json-schema -o src/types/schema/manifest.json --required src/types/manifest.ts Manifest
```

systemservices/marketplace/src/contracts/manifest.ts
ilgooz: done

systemservices/marketplace/src/contracts/version.ts
ilgooz: done

systemservices/marketplace/src/contracts/version.ts
ilgooz: removed

systemservices/marketplace/gen-schema
ilgooz: done

antho1404:
This PR https://github.com/mesg-foundation/core/pull/837 can overlap a bit with this one.
This PR is still useful when we read the manifest, not when we write it (validation done by the core).

We still need validation of manifest+definition but i'm not sure also the json-schema for the service is actually accurate with the validation that we have on the core and if it's not really valid is it even useful to have it or isn't it better to have the core that can handle and fallback correctly for invalid data ?

systemservices/marketplace/src/contracts/version.ts
```diff
@@ -30,12 +30,14 @@ const getServiceVersion = async (contract: Marketplace, versionHash: string): Pr
   }
   const version = await contract.methods.serviceVersion(versionHashHex).call()
   const manifestData = await getManifest(hexToAscii(version.manifestProtocol), hexToAscii(version.manifest))
-  return {
-    versionHash: versionHash,
-    manifest: hexToAscii(version.manifest),
-    manifestProtocol: hexToAscii(version.manifestProtocol),
-    manifestData: manifestData,
-    createTime: parseTimestamp(version.createTime),
+  if (manifestData) {
```
NicolasMahe: The version should be returned even if manifest data is not valid. Otherwise, there is no knowledge about the version. See comment https://github.com/mesg-foundation/core/pull/837#pullrequestreview-219372297

systemservices/marketplace/src/types/schema/manifest.json
```diff
@@ -0,0 +1,55 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "properties": {
+        "service": {
+            "properties": {
+                "definition": {
+                    "properties": {
+                    },
+                    "type": "object"
```
NicolasMahe: Is it possible to directly "link" to other JSON schema?

systemservices/marketplace/src/types/schema/manifest.json
```diff
@@ -0,0 +1,55 @@
...
+                "readme": {
+                    "type": "string"
+                }
+            },
+            "required": [
+                "definition",
+                "deployment",
+                "hash",
+                "hashVersion",
+                "readme"
```
NicolasMahe: Readme could be empty.
Test if an empty string is ok or not.

systemservices/marketplace/src/contracts/version.ts
antho1404: I'm not sure this is something nice to have. For me the service should shield from any wrong data and just return stuff that we can exploit. If we want the raw data that we cannot use we can have them directly on ethereum. In which case do we need to return invalid data ?

systemservices/marketplace/src/contracts/manifest.ts
```diff
@@ -21,12 +26,18 @@ const getManifest = async (protocol: string, source: string): Promise<Manifest|u
       console.warn('protocol', protocol, 'is not compatible with this service')
       return
     }
-    const manifest = await get[protocol](source)
-    if (typeof manifest === 'object') {
-      return manifest as Manifest
+    const manifest: Manifest = await get[protocol](source)
+    const validation = validator.validate(manifest, manifestSchema)
+    if (!validation.valid) {
+      console.warn('manifest', protocol, '::', source, 'is not a valid manifest')
```
antho1404: can we throw error in this function and catch it if we need to when we call this function. Just log in the console and return nothing is not really safe...

systemservices/marketplace/src/types/schema/definition.json
```diff
@@ -0,0 +1,246 @@
+{
+  "$schema": "http://json-schema.org/draft-06/schema#",
```
antho1404: are we sure that this definition is still valid with the one we have in the core ?
I see sid for example:
Core:
```
Sid string `yaml:"sid" json:"sid,omitempty" validate:"omitempty,printascii,max=63,domain"`
```
Here:
```
"sid": {
      "type": "string",
      "maxLength": 39,
      "pattern": "^[0-9a-zA-Z\\._-]+$"
 },
```

This is just one example I looked at but we might have more and I think now the core only should be responsible for this validation. Not this service (or this service should have a way to delegate to the core with an api or another service that the core uses with service composition). I really don't think this is a good idea, I would prefer to return invalid data then have a good error when we deploy than having invalid errors and miss some versions.

systemservices/marketplace/src/types/schema/manifest.json
antho1404: changing the type to something more accurate like this would probably solve this and add better auto completion
```ts
export interface Manifest {
  version: '1'
  service: {
    hash: string
    hashVersion: '1'
    deployment: {
      type: string
      source: 'ipfs'|'http'|'https'
    }
    definition: any,
    readme?: string
  }
}
```

systemservices/marketplace/src/types/schema/definition.json
NicolasMahe: I update the validation of the service definition.
I think this validation is not perfect but already cover 90% of the definition and 100% of the structure. I think it's anyway a good thing to make sure the structure is correct.
Tell me if your ok to keep it or completely discard it.

antho1404: I just feel this is really risky because we will probably update the structure in the core multiple time and this validation will not possibly keep up with that.
I think we should consider this definition as valid by default and let the core handle this.
One solution could be to provide an api for validation either with service composition or a core api.

antho1404: I'm not 100% against it, I'm ok to keep it but with a big warning and I think better to run with no validation for now and do something nice than run on a partial validation that will be really hard to maintain

NicolasMahe: > I just feel this is really risky because we will probably update the structure in the core multiple time and this validation will not possibly keep up with that.
I think we should consider this definition as valid by default and let the core handle this.
One solution could be to provide an api for validation either with service composition or a core api.

i totally agree. But what to do until there is a validation api? And even, the validation api will be available on the core api but this is a service...

> I'm not 100% against it, I'm ok to keep it but with a big warning and I think better to run with no validation for now and do something nice than run on a partial validation that will be really hard to maintain

i'm very afraid that the website can be broken because of wrong service structure...

PR #817, opened 2019-03-15
Add command marketplace

NicolasMahe:
Dependent on https://github.com/mesg-foundation/core/pull/810

Deployed at address [`0x94F4cB92Fe9f547574Aec617B1594B13aBd47Ad3`](https://ropsten.etherscan.io/address/0x94F4cB92Fe9f547574Aec617B1594B13aBd47Ad3) with token address [`0x5861B3DC52339d4f976B7fa5d80dB6cd6f477F1B`](https://ropsten.etherscan.io/token/0x5861b3dc52339d4f976b7fa5d80db6cd6f477f1b).

Procedure to test:

####  Import account that has MESG token into wallet
```
./dev-cli wallet import
```
---

#### Try to deploy a paid service
```
./dev-cli service deploy mesg://marketplace/service/6de8cf8fd74582eb46318058154f79c6ba4e04e83889399fc414adc89c880025
```
See error message

#### Purchase service
```
./dev-cli marketplace purchase com.mesg.bitcoin
```

#### Deploy again
```
./dev-cli service deploy mesg://marketplace/service/6de8cf8fd74582eb46318058154f79c6ba4e04e83889399fc414adc89c880025
```
Should work 🎊 🍾  (only for 3min because offer is only for 3min)

----

#### Publish
```
./dev-cli marketplace publish ./PATH_TO_SERVICE
```

#### Create offer
```
./dev-cli marketplace create-offer SID_OF_SERVICE
```

PR #835, opened 2019-03-21
Ss/marketplace decode receipt

NicolasMahe:
This PR mainly add 3 new tasks to decode the transaction receipt and return useful information about what the transaction did.
It also detailed the return data of task `sendSignTransaction` and merge the different events into one data struct.

I will create another PR that implement this modification in the marketplace commands.

This PR is the service side of stuff needed for this comments:
https://github.com/mesg-foundation/core/pull/817/files/a56eb0e3aa55f2fc435a0e76537b7c52eb4375c9#r266858115
https://github.com/mesg-foundation/core/pull/817/files/a56eb0e3aa55f2fc435a0e76537b7c52eb4375c9#r266861843


antho1404:
I was thinking something based on the event for the command and this, the refactoring for the events is good but I don't see the need of creating the tasks for this. The service should expose the data and task really necessary and should abstract as much as possible to technology it uses.

For me this service should only expose data and tasks related to the data of the marketplace not to ethereum (except maybe the transaction to sign and propagate). So adding tasks here about processing data from eth is not necessary especially that we have these data with the events and we just need to listen the events.

NicolasMahe:
> I was thinking something based on the event for the command and this, the refactoring for the events is good but I don't see the need of creating the tasks for this. The service should expose the data and task really necessary and should abstract as much as possible to technology it uses.
> 
> For me this service should only expose data and tasks related to the data of the marketplace not to ethereum (except maybe the transaction to sign and propagate). So adding tasks here about processing data from eth is not necessary especially that we have these data with the events and we just need to listen the events.

There is no way to fully link a transaction to the event, except by decoding the event from the log of the transaction receipt (what this PR is doing).

As there is no know index, it will be a big guess to associate and triggered event from the execution of a task. What happen when many people are using the smart contract at the same time? Then some execution of task will return the result of other. Otherwise, it needs to listen until a event match some criteria, but i feel it's overcomplicated.

NicolasMahe:
The only way to get a nice result with only one function, will be to merge, for example, `createServiceOrder`, with `signTransaction` and `decodeLogCreateServiceOrder`. But to do so, it will require a service composition to use signTransaction of service `ethwallet`.

antho1404:
FYI, here is a proof of concept that solves the problem with the CLI and doesn't need any modification here https://github.com/mesg-foundation/core/compare/feature/command-marketplace...feature/command-marketplace-with-listen

NicolasMahe:
closing in favor of https://github.com/mesg-foundation/core/pull/844

PR #837, opened 2019-03-24
Marketplace service new api for manifest

antho1404:
The goal here is to add validation of the manifest data directly with the core by passing all arguments necessary to the service and giving responsibility to the service to manage how it wants to upload this manifest (with IPFS or whatever).

This remove a lot of logic from the client and also remove the management of version from the client, now the service manages the version (we change the service == we change the version).

The client doesn't need to know anything about how data are stored in the marketplace IPFS or no. For now we still need to upload the sources but ultimately we should even give directly the tarball to the system service and let it handle everything

@NicolasMahe want to have your feedback before I continue to make sure that this is aligned with what you want with #829 and #835 

TODO:
- [x] fix error with `./dev-cli service execute marketplace --task getService --data sid=myservice`


commands/provider/marketplace_provider.go
```diff
@@ -85,49 +80,23 @@ func (p *MarketplaceProvider) IsAuthorized(sid string, versionHash string, addre
 	return output.Authorized, output.Sid, output.Source, output.Type, p.call("isAuthorized", input, &output)
 }
 
-// UploadServiceFiles upload the tarball and the definition file, and returns the address of the definition file
-func (p *MarketplaceProvider) UploadServiceFiles(path string, manifest MarketplaceManifestData) (protocol string, source string, err error) {
+// UploadSources upload the tarball, and returns the address of the uploaded sources
+func (p *MarketplaceProvider) UploadSources(path string) (SourceDeployment, error) {
```
NicolasMahe:
```suggestion
func (p *MarketplaceProvider) UploadSource(path string) (SourceDeployment, error) {
```

commands/provider/marketplace_types.go
```diff
@@ -52,19 +46,25 @@ type MarketplaceService struct {
 	} `json:"purchases"`
 }
 
+// SourceDeployment is the information related to a deployment
+type SourceDeployment struct {
```
NicolasMahe:
```suggestion
type MarketplaceDeployedSource struct {
```

commands/provider/marketplace_types.go
```diff
@@ -52,19 +46,25 @@ type MarketplaceService struct {
...
+	Type   string `json:"type"`
+	Source string `json:"source"`
+}
+
+// MarketplaceServiceData is the data present to the manifest and sent to create a new service's version
```
NicolasMahe:
```suggestion
// MarketplaceManifestServiceData is the service's data present in the manifest.
```

systemservices/marketplace/mesg.yml
```diff
@@ -49,18 +49,26 @@ tasks:
...
+                        type: Object
+                        object:
+                          hash: { type: String }
+                          hashVersion: { type: String }
+                          deployment:
+                            type: Object
+                            object:
+                              type: { type: String }
+                              source: { type: String }
+                          definition: { type: Object }
```
NicolasMahe: why using `{ }` style?

systemservices/marketplace/mesg.yml
```diff
@@ -49,18 +49,26 @@ tasks:
...
-                    description: ""
-                    type: String
-                  manifestData: &manifestData
                     name: ""
                     description: ""
                     type: Object
                     optional: true
+                    object: 
+                      version: { type: String }
+                      service: &serviceInformation
```
NicolasMahe:
```suggestion
                      service: &manifestServiceData
```

commands/provider/marketplace_types.go
```diff
@@ -52,19 +46,25 @@ type MarketplaceService struct {
...
+type MarketplaceServiceData struct {
```
NicolasMahe:
```suggestion
type MarketplaceManifestServiceData struct {
```

systemservices/marketplace/mesg.yml
```diff
@@ -130,9 +138,7 @@ tasks:
     name: ""
     description: ""
     inputs:
-      sid: *sid
-      manifest: *manifest
-      manifestProtocol: *manifestProtocol
+      service: *serviceInformation
```
NicolasMahe:
```suggestion
      service: *manifestServiceData
```

systemservices/marketplace/src/contracts/version.ts
```diff
@@ -10,33 +10,70 @@ const getServiceVersions = async (contract: Marketplace, sid: string): Promise<V
...
+          source: version.manifest.service.deployment.source,
+        },
+        hash: version.manifest.service.hash,
+        hashVersion: version.manifest.service.hashVersion,
+        readme: version.manifest.service.readme || '',
+      }
+    }
+  }
+}
+
```
NicolasMahe: I think `validVersion` and `normalizeVersion` should be removed in favor of validation done in https://github.com/mesg-foundation/core/pull/829.

systemservices/marketplace/src/events/serviceVersionCreated.ts
```diff
@@ -1,12 +1,13 @@
 import Service, { EmitEventReply } from "mesg-js/lib/service/service"
 import { EventLog } from "web3/types";
 import { hexToAscii } from "../contracts/utils";
+import { getManifest } from "../contracts/manifest";
 
-export = (mesg: Service, event: EventLog): Promise<EmitEventReply | Error> => {
+export = async (mesg: Service, event: EventLog): Promise<EmitEventReply | Error> => {
+  const manifest = await getManifest(hexToAscii(event.returnValues.manifestProtocol), hexToAscii(event.returnValues.manifest))
```
NicolasMahe:
```suggestion
```

I think event should not return the content of manifest. I see event as small and fast message to trigger other function. In this case, the triggered function can get the manifest if it needs to.
Otherwise, why not returning also ALL the service's data?

systemservices/marketplace/src/contracts/version.ts
```diff
@@ -10,33 +10,70 @@ const getServiceVersions = async (contract: Marketplace, sid: string): Promise<V
     return []
   }
   const versionsLength = new BigNumber(await contract.methods.serviceVersionsLength(sidHex).call())
-  const versionsPromise: Promise<Version|undefined>[] = []
+  const versions: Version[] = []
   for (let j = new BigNumber(0); versionsLength.isGreaterThan(j); j = j.plus(1)) {
-    versionsPromise.push(getServiceVersionWithIndex(contract, sid, j))
+    try {
+      const version = normalizeVersion(await getServiceVersionWithIndex(contract, sid, j))
```
NicolasMahe: please fetch the version in a parallel way

systemservices/marketplace/src/types/version.ts
```diff
@@ -2,8 +2,6 @@ import { Manifest } from "./manifest";
 
 export interface Version {
   versionHash: string;
-  manifest: string;
-  manifestProtocol: string;
-  manifestData: Manifest|undefined;
+  manifest: Manifest;
```
NicolasMahe: I'm not sure removing basic data like `manifestProtocol` and `manifest` is a good thing.
I see the advantage of creating a separate layer between the api of the smart contract and the api of this service, but here it could still be useful to return the manifest source "url".

systemservices/marketplace/mesg.yml
antho1404: just more compact no other reasons

systemservices/marketplace/src/contracts/version.ts
antho1404: Parallel here makes it really complicated to read with some mapping and filtering.
That's the problem of async either we use full async/await or full promise but mixing is really hard to read. But I can put the parallel version and we can improve later

systemservices/marketplace/src/types/version.ts
antho1404: I don't see any reasons to return these information if it is information managed by the service itself. The question to ask is will we ever need this information to access the data if we always have the data.

systemservices/marketplace/src/events/serviceVersionCreated.ts
antho1404: Are you saying that this should be moved to the method that calls this function or that it should simply disappear and we just return the raw data without interpretation of these data.

I would actually be ok to return all the service data and have the same structure than the `getService` or `listServices` that would make even more sense.

systemservices/marketplace/src/types/version.ts
NicolasMahe: > Only returning the manifest's data but not manifestProtocol and manifestData is too much for my point of view. Those 2 info, even if they are completely managed by the service, are still valid to return.
As the marketplace is on Ethereum, anyone can add a new version with wrong manifest data. In this PR's current implementation, the version is discarded and it will appear that the service has no version, even if it actually has one. The wrong version cannot be deployed on a core, but should still be return as a version with wrong manifest. Othercase: the manifest is valid but for some reason is not available on IPFS anymore?
I think returning the version even if the manifest is not valid / available is better than just discard it.

Even more, what about service that have the manifest in private host that require authentication?

systemservices/marketplace/mesg.yml
NicolasMahe: Can you keep the same style as the rest of the file

systemservices/marketplace/src/contracts/version.ts
antho1404: I can remove but then it will break some stuff because now the core is actually validating some data (because we don't have an any anymore but the actual structure of data) and this breaks a lot of services.

I was thinking more to move the validation in this function `validate` but i'm ok to revert as long as the other one is merged pretty much at the same time in order not to break everything

antho1404:
All the unnecessary dev on this PR has been reverted, now only the `publishService` has been updated, the rest will be done in other PR if necessary

commands/marketplace_publish.go
```diff
@@ -110,9 +105,9 @@ func (c *marketplacePublishCmd) runE(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	fmt.Printf("%s Service published with success\n", pretty.SuccessSign)
-	fmt.Printf("%s See it on the marketplace: https://marketplace.mesg.com/services/%s\n", pretty.SuccessSign, c.manifest.Service.Definition.Sid)
+	fmt.Printf("%s See it on the marketplace: https://marketplace.mesg.com/services/%s\n", pretty.SuccessSign, c.sid)
 
-	fmt.Printf("%s To create a service offer, execute the command:\n\tmesg-core marketplace create-offer %s\n", pretty.SuccessSign, c.manifest.Service.Definition.Sid)
+	fmt.Printf("%s To create a service offer, execute the command:\n\tmesg-core marketplace create-offer %s\n", pretty.SuccessSign, c.sid)
```
NicolasMahe: `c.sid` is always an empty string. You forgot to set its value after the deploy.

NicolasMahe: Same for `c.hash`

From:
```
	sid, hash, err := deployService(c.e, c.path, map[string]string{})
```
to
```
	c.sid, c.hash, err := deployService(c.e, c.path, map[string]string{})
```

https://github.com/mesg-foundation/core/pull/837/files#diff-18a1eba1f644f5f71dc449b67f9e1f24R50

antho1404: good catch, fixed

systemservices/marketplace/src/tasks/publishServiceVersion.ts
```diff
@@ -1,16 +1,21 @@
...
   try {
+    const manifestHash = await uploadManifest({
+      version: '1',
+      service: inputs.service
+    })
     const transactionData = marketplace.methods.publishServiceVersion(
-      asciiToHex(inputs.sid),
-      asciiToHex(inputs.manifest),
-      asciiToHex(inputs.manifestProtocol)
+      asciiToHex(inputs.service.definition.sid),
```
NicolasMahe: This is super dangerous...
Service is of type Object without any more detailed in the mesg.yml.
I will recommend to at least specify the sid in the mesg.yml and validate it with the json schema of the other PR.

NicolasMahe: i add the sid in the mesg.yml in commit https://github.com/mesg-foundation/core/pull/837/commits/e2d724eeab8ab298fbcd1b088c6f740b3ddd1bd9

PR #842, opened 2019-03-29
Tag image built

antho1404:
Right now sources are deterministic because based on the source code but the data we save in the database are not because of the `Image` that is the sha of the docker layer and this one is not deterministic.

This PR is a proposal to change the data in the database by saving a tag instead of the id of the docker layer.

- We build the docker image but we tag the output with `sid:hash`
- We save the tag in the database instead of the sha from the docker layer
- We have a service in database that is constant whatever computer/docker version.... 🎉 

These data can now be exchanged on the network and on different computer and so be shared on the marketplace.

Bonus: Now we have some nice informations with `docker images`

Screen Shot 2019-03-29 at 2 24 51 pm
vs
Screen Shot 2019-03-29 at 2 24 58 pm

### DIFF
--- a/container/build.go
+++ b/container/build.go
@@ -22,7 +22,7 @@ type BuildResponse struct {
 }
 
 // Build builds a docker image.
-func (c *DockerContainer) Build(path string) (tag string, err error) {
+func (c *DockerContainer) Build(path, name, version string) (tag string, err error) {
 	excludeFiles, err := dockerignoreFiles(path)
 	if err != nil {
 		return "", err
@@ -33,15 +33,17 @@ func (c *DockerContainer) Build(path string) (tag string, err error) {
 		return "", err
 	}
 	defer buildContext.Close()
+	tag = name + ":" + version
 	response, err := c.client.ImageBuild(context.Background(), buildContext, types.ImageBuildOptions{
 		Remove:         true,
 		ForceRemove:    true,
 		SuppressOutput: true,
+		Tags:           []string{tag},
 	})
 	if err != nil {
 		return "", err
 	}
-	return parseBuildResponse(response)
+	return tag, parseBuildResponse(response)
 }
 
 // dockerignoreFiles reads excluded files from .dockerignore.
@@ -57,20 +59,20 @@ func dockerignoreFiles(path string) ([]string, error) {
 	return dockerignore.ReadAll(f)
 }
 
-func parseBuildResponse(response types.ImageBuildResponse) (tag string, err error) {
+func parseBuildResponse(response types.ImageBuildResponse) (err error) {
 	lastOutput, err := extractLastOutputFromBuildResponse(response)
 	if err != nil {
-		return "", err
+		return err
 	}
 	var buildResponse BuildResponse
 
 	if err := json.Unmarshal([]byte(lastOutput), &buildResponse); err != nil {
-		return "", fmt.Errorf("could not parse container build response. %s", err)
+		return fmt.Errorf("could not parse container build response. %s", err)
 	}
 	if buildResponse.Error != "" {
-		return "", fmt.Errorf("image build failed. %s", buildResponse.Error)
+		return fmt.Errorf("image build failed. %s", buildResponse.Error)
 	}
-	return strings.TrimSuffix(buildResponse.Stream, "\n"), nil
+	return nil
 }
 
 func extractLastOutputFromBuildResponse(response types.ImageBuildResponse) (lastOutput string, err error) {
--- a/container/build.go
+++ b/container/build.go
@@ -22,7 +22,7 @@ type BuildResponse struct {
 }
 
 // Build builds a docker image.
-func (c *DockerContainer) Build(path, name, version string) (tag string, err error) {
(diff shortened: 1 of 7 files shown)


container/build.go
```diff
@@ -33,15 +33,17 @@ func (c *DockerContainer) Build(path string) (tag string, err error) {
 		return "", err
 	}
 	defer buildContext.Close()
+	tag = name + ":" + version
```
NicolasMahe: what is the limitation on docker tag length and characters?

NicolasMahe: also, i suggest to add the prefix `mesg/` or `mesg-service/` (and create the org on docker hub) so we prevent people to upload service image on docker hub (because they will not have the org access).

krhubert: from https://docs.docker.com/engine/reference/commandline/tag/

A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and dashes. A tag name may not start with a period or a dash and may contain a maximum of 128 characters.

container/build.go
```diff
@@ -57,20 +59,20 @@ func dockerignoreFiles(path string) ([]string, error) {
...
 	}
 	var buildResponse BuildResponse
 
 	if err := json.Unmarshal([]byte(lastOutput), &buildResponse); err != nil {
-		return "", fmt.Errorf("could not parse container build response. %s", err)
+		return fmt.Errorf("could not parse container build response. %s", err)
 	}
 	if buildResponse.Error != "" {
-		return "", fmt.Errorf("image build failed. %s", buildResponse.Error)
+		return fmt.Errorf("image build failed. %s", buildResponse.Error)
```
krhubert:
```suggestion
		return fmt.Errorf("image build failed: %s", buildResponse.Error)
```

container/build.go
```diff
@@ -57,20 +59,20 @@ func dockerignoreFiles(path string) ([]string, error) {
...
 	lastOutput, err := extractLastOutputFromBuildResponse(response)
 	if err != nil {
-		return "", err
+		return err
 	}
 	var buildResponse BuildResponse
 
 	if err := json.Unmarshal([]byte(lastOutput), &buildResponse); err != nil {
-		return "", fmt.Errorf("could not parse container build response. %s", err)
+		return fmt.Errorf("could not parse container build response. %s", err)
```
krhubert:
```suggestion
		return fmt.Errorf("could not parse container build response: %s", err)
```

container/build.go
antho1404: so do we prepend `mesg` in that case ?
We have sid `63` and hash `64` plus the separator `:` we have a total of 128 but more if we add the `mesg`

container/build.go
```diff
@@ -22,7 +22,7 @@ type BuildResponse struct {
 }
 
 // Build builds a docker image.
-func (c *DockerContainer) Build(path string) (tag string, err error) {
+func (c *DockerContainer) Build(path, name, version string) (tag string, err error) {
```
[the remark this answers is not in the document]
antho1404: docker has an flag named `tag` that takes `name:tag` is it me or this is really not clear, what's the tag ? I feel version is less confusing

container/build.go
NicolasMahe: only 44 character for the hash if it's in base58 ;)

container/build.go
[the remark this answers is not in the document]
NicolasMahe: i prefer name and version when it's the 'mesg' level.

container/build.go
antho1404: so i'll keep the `mesg/` and we will change the representation of the hash in `base58` :)

antho1404:
I'm not sure I really follow here. This doesn't change the system in docker, it just creates a tag/alias to access this non-deterministic "ID" from docker. It still doesn't fix anything related to the actual build of docker (this will need to be solved but not the purpose of this PR).

We are not sharing anyway any compiled sources and have to rebuild everything all the time so this will always be linked to the latest built version. But it give us a way to have something that every computer will generate the same way and so everyone can verify that the source code was not changed. The version built is another thing that for now is not verifiable.

ilgooz:
> I'm not sure I really follow here. This doesn't change the system in docker, it just creates a tag/alias to access this non-deterministic "ID" from docker. It still doesn't fix anything related to the actual build of docker (this will need to be solved but not the purpose of this PR).
> 
> We are not sharing anyway any compiled sources and have to rebuild everything all the time so this will always be linked to the latest built version. But it give us a way to have something that every computer will generate the same way and so everyone can verify that the source code was not changed. The version built is another thing that for now is not verifiable.

What I mean is, yes Docker image is not a part of actual source code but it's a close relative. A piece of dependency installed by the instructions inside Dockerfile may affect how the source will behave. Which means it'll make the source code to change indirectly.

It is also possible to manipulate the actual source code with the instructions inside Dockerfile. For example, Dockerfile might be fetching some extra source code files from the internet and put them next to the service's source files. This is a possible hack & a security issue which there is no way to detect it unless user check it manually.

As a result, it's not good to force deterministic behaviour. It's not possible to know if the program's behaviour will change or not between different builds of Dockerfile.

PR #843, opened 2019-03-29
Refactor event and tasks

NicolasMahe:


systemservices/marketplace/src/events/index.ts
```diff
@@ -0,0 +1,72 @@
+import { hexToAscii, fromUnit, findInAbi, parseTimestamp } from "../contracts/utils";
+import { ABIDefinition } from "web3/eth/abi";
+import _marketplaceABI from "../contracts/Marketplace.abi.json"
+const marketplaceABI = _marketplaceABI as ABIDefinition[]
+
+const eventHandlers: {[ethName: string]: {
```
ilgooz: I'd rather to simplify this whole structure to something like below:

```js
export eventHandlers = {
  serviceCreated: () => {}, // handler func
  ...
}
```

We have the same event names for both smart contract and mesg with only the char case difference which we can dynamically convert to each other with a simple func.

This way readability for the code will be better and complexity will be avoided.

antho1404: handlers are usually function, this should be renamed to `eventDefinition`, `eventMapping`...

I also don't like much this way of managing the events, this creates a huge dependency between the events from the smart contract and the events from the service (that are different even if we are just doing a proxy). 

We could have a `eventMapping` that does the mapping between smart contract to the data needed for the events on the marketplace but we will still have separated events.

eg:
```js
eventMapping = {
  smartcontracteventX: event => ({ x: ... })
}

// event/serviceCreated.ts
export data => mesg.emitEvent('serviceCreated', eventMapping['smartcontracteventX'](data)
```

NicolasMahe:
I updated this PR:
- I create a new file `contract/parseEvents.ts` that only normalize event from ethereum (convert timestamp to date, unit 'wei' to 'ether', hex to ascii, hash to base58, etc..)
- I moved the logic about event from `index.ts` to `events/index.ts`. I keep the same logic: this file map smart contract event to individual event handler. Each event handler then emit to mesg (if they want to) and parse the event using the functions in first point.
- I also moved the logic about task from `index.ts` to `tasks/index.ts`. No logic changed.

PR #844, opened 2019-03-29
Add publish tasks to service marketplace

NicolasMahe:
Dependent on https://github.com/mesg-foundation/core/pull/843

I created 3 task to replace the generic `sendSignTransaction`. Each task return the decoded event from the receipt of the ethereum transaction. It means that the service output the data that has been added to ethereum.

Eg:
```
➜  core git:(ss/marketplace-publish-tasks) ✗ ./dev-cli service execute marketplace --task publishPurchase --data signedTransactions=0xf8e91d843b9aca00830f42409494f4cb92fe9f547574aec617b1594b13abd47ad380b8843a1835e1000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000011636f6d2e6d6573672e657468657265756d0000000000000000000000000000002aa0068bc218cdc2cc3355ae4d0cf1d76299fa472a53efcceebe519f9529fa01abaea0713b0ba42f117fdcb0cab2062aa9b6d71ae496901dedb241794a834d2fccf68f
✔ Task "publishPurchase" executed
✔ Task publishPurchase returned output success with data:
{"duration":"86400","expire":"2019-03-31T10:53:03.000Z","offerIndex":"0","price":"1000","purchaser":"0xcAB79fA69c68CB4C65fa5C6E05BC4dBa5FB57D11","sid":"com.mesg.ethereum"}
```

Because there is no service composition yet, for example to create an offer, the user first needs to call `prepareCreateServiceOffer`, then sign the transaction with the service ethwallet, and then call `publishCreateServiceOffer`.


systemservices/marketplace/src/tasks/publishPurchase.ts
```diff
@@ -0,0 +1,24 @@
...
+import { decodeLog } from "../contracts/utils";
+import { eventHandlers } from "../events";
+
+export default (
+  web3: Web3,
+) => async (inputs: TaskInputs, outputs: TaskOutputs): Promise<void> => {
+  try {
+    let receipt = null
+    for (let i=0; i<inputs.signedTransactions.length; i++) {
+      receipt = await web3.eth.sendSignedTransaction(inputs.signedTransactions[i])
```
antho1404: why multiple tx here but we collect only the results from the last one

systemservices/marketplace/src/tasks/publishPurchase.ts
```diff
@@ -0,0 +1,24 @@
...
+    }
+    if (receipt === null || receipt.logs === undefined) throw new Error('receipt does not contain logs')
+    const eventHandler = eventHandlers['ServicePurchased']
+    const decodedLog = decodeLog(web3, eventHandler.abi, receipt.logs[receipt.logs.length - 1])
```
antho1404: I feel its a bit hacky to take the last log. Can't we just iterate over the logs and filter for the one who match the event we want ?

systemservices/marketplace/src/tasks/publishPurchase.ts
NicolasMahe: There is either 1 or 2 transaction.
The optional first one is the erc20 approval of the marketplace to spend token on behalf of user.
The second, is the actual purchase of the service.
Only the last one is the important one.

systemservices/marketplace/src/tasks/publishPurchase.ts
NicolasMahe: Updated by filtering the array of logs.

systemservices/marketplace/src/tasks/publishPurchase.ts
```diff
@@ -0,0 +1,25 @@
...
+import { extractEventFromLogs } from "../contracts/utils";
+import { Marketplace } from "../contracts/Marketplace";
+import { servicePurchased } from "../contracts/parseEvents";
+
+export default (
+  web3: Web3,
+  marketplace: Marketplace,
+) => async (inputs: TaskInputs, outputs: TaskOutputs): Promise<void> => {
+  try {
+    if (inputs.signedTransactions.length === 2) {
```
antho1404: what happen if `length === 3` ? I agree in the context of us using it it should never happen but nothing stops someone to do it with the api.

I would do it in 2 different ways:

## Strong check
```js
if (inputs.signedTransactions.length > 2) {
  throw new Error("Can only accept maximum signed transactions")
}
// Then now you can play with the index directly
```

## Soft way
```js
const txs = [...inputs.signedTransactions] // Copy of the array just because it's better if you change it
const importantTx = txs.pop()
for (var tx in txs) {
  await web3.eth.sendSignedTransaction(tx)
}
const receipt = await web3.eth.sendSignedTransaction(importantTx)
```

NicolasMahe: I don't understand what your asking.

The original way was like the soft way you are proposing:
```ts
let receipt = null
for (let i=0; i<inputs.signedTransactions.length; i++) {	
  receipt = await web3.eth.sendSignedTransaction(inputs.signedTransactions[i])
}
```

It could even be done with a reduce

antho1404: Not exactly the same.

My version: You are splitting the data and doing 2 different processing (one loop and one result collection:
You version: You are doing the same processing, getting the result but in the end this processing for the first element is kind of ignored and that's confusing. I agree the result is exactly the same, it's just that it's easier to understand.
Basically:
```
extractTransactions()
processTransactionsWhereWeDontNeedResult()
processTransactionWhereWeNeedResult()
```

It's nothing really blocking anyway but can be improved. It's never good to assign a variable if you don't use it especially if replace it right after.

NicolasMahe: Done

NicolasMahe:
@mesg-foundation/core this PR is ready to review

PR #845, opened 2019-03-30
Marketplace publish service raw data

antho1404:
depends on https://github.com/mesg-foundation/core/pull/842

The marketplace was sending the content of the `mesg.yml` as definition, I feel it's better to have the data that has been processed (the ones directly in the database after de deployment) like that if we want we can have another source of validation and also the `mesg.yml` can always be compiled down to these data, the other was is not necessarily possible.

With this PR I also realized that `hash` and `hashVersion` may not be necessary in the marketplace manifest (https://github.com/mesg-foundation/core/blob/12e1b5e61aa664dc33678a155c24994489868c75/commands/marketplace_publish.go#L91-L92)
- `hash` is present in the raw data
- `hashVersion` is something that should be present in the raw data

This is for another PR but I would like to remove this from the marketplace and move the `hashVersion`  that I would call just `version` (for service version) directly in the core.


antho1404:
pull request updated without the update on the image tag so ready to review cc @mesg-foundation/core

commands/provider/marketplace_types.go
```diff
@@ -54,11 +54,11 @@ type MarketplaceDeployedSource struct {
 
 // MarketplaceManifestServiceData is the data present to the manifest and sent to create a new service's version
 type MarketplaceManifestServiceData struct {
-	Definition  importer.ServiceDefinition `json:"definition"`
-	Readme      string                     `json:"readme,omitempty"`
-	Hash        string                     `json:"hash"`
-	HashVersion string                     `json:"hashVersion"`
-	Deployment  MarketplaceDeployedSource  `json:"deployment"`
+	Definition  *coreapi.Service          `json:"definition"`
```
ilgooz: use `definition.Service`?

antho1404: good catch, that's because we merged this but still kept the service also in the definition and this PR was done before that... fixed

antho1404:
> can you express why we introduce xxx-service?

Because I did a wrong merge to resolve some conflicts and actually took some local files that I didn't wanted to commit. Sorry for that, it's reverted

commands/marketplace_publish.go
```diff
@@ -48,15 +47,19 @@ func (c *marketplacePublishCmd) preRunE(cmd *cobra.Command, args []string) error
...
+	_, hash, err := deployService(c.e, c.path, map[string]string{})
 	if err != nil {
 		return err
 	}
-	fmt.Printf("%s Service deployed with sid %s and hash %s\n", pretty.SuccessSign, pretty.Success(c.sid), pretty.Success(c.hash))
+	c.service, err = c.e.ServiceByID(hash)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("%s Service deployed with sid %s and hash %s\n", pretty.SuccessSign, pretty.Success(c.service.Definition.Sid), pretty.Success(c.service.Definition.Hash))
```
ilgooz: it might be better to explicitly use `hash` got from `deployService()` here. because in future, we'll make service sids to have one-to-many relationships with service hashes.  

thus, `ServiceByID()` may return multiple versions and maybe in desc order on deploy time but it's not guaranteed that the last hash in the array will be belong to the last publish because two publish commands may run in parallel for the different versions of same service.

ilgooz: if `ServiceByID(**sid**)` is going to return the last service version and `ServiceByID(**hash**)` will return the same version with `hash`, then all ok.

ilgooz: anyway, we're also using `c.service.Definition` so we have to make this call. let's make sure that `ServiceByID(hash)` will return the version for `hash` in future and `ServiceByID(sid)` may return the latest hash.

PR #846, opened 2019-04-01
Support docker machine

antho1404:
Small feature to support [docker machine](https://docs.docker.com/machine/) from the CLI

We can now with that connect to any remote core either on a VirtualBox instance on your computer or an instance for example on DigitalOcean.

```bash
eval $(docker-machine env NAME)
./dev-cli ...
```


interface/cli/main.go
```diff
@@ -52,3 +59,19 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func coreAddress(cfg *config.Config) (string, error) {
```
krhubert: This func be moved to config package (and docker_host should be parsed and load with `Load` Function)

antho1404: No this is something related to the CLI only, config is a package shared between the Core and CLI. Ultimately CLI should not access the config at all and only rely on the `Info` API but to fetch this api it needs an endpoint for the core address and a detection for this endpoint.

krhubert: But for now, it reads all data from the config and using it (so config is shared between cli and server). So this one should be handled there as well (as all other envs/variables used by CLI)

antho1404: but config on the Core don't rely on the host environment, this config does. We are reading the env variables from the host that the core will never have. I could agree to have that on the core if the core itself can give its own address and in that case we don't even need any docker env variable but this is not the case.

Other solution that might be the way to go is to actually inject the config when we run the CLI like that the CLI script will inject the address based on the docker machine and the CLI still get the data from the config but the config package doesn't need to have any dependency with the env from the host. Tell me what you think about that ?

krhubert: I'm not quite following the part:

> We are reading the env variables from the host that the core will never have. (....)
 any dependency with the env from the host.

I don't see reason these env variables shouldn't be read in config package (here is just simply read - same as in config), but it you feel this one should stay in `main.go` - i'm ok with this.

antho1404: Otherwise you can have a look at this modification https://github.com/mesg-foundation/core/compare/feature/docker-machine-support...feature/docker-machine-support-2

It's not adding support in the CLI but in the dev tools and we can see later about the support for the CLI especially if we move the CLI in node... which will force us not to use the config package

krhubert: Yes I see, is this only for dev purpose? Because you mention about DO in PR comment above so I guess it won't work with such script. So if this is only for dev it's better to be in dev-cli then.

antho1404: For my specific need it's for dev (I run all my docker stack in a VM). I think anyway it's a nice feature to have in the CLI in general to have the support of docker machine and DO and other but this can be documented with the `MESG_CLIENT_ADDRESS` env variable anyway so i'm fine to have this just for dev and merge this other branch inside this one to revert

dev-cli
```diff
@@ -4,4 +4,10 @@ VERSION="local"
 LDFLAGS="-X 'github.com/mesg-foundation/core/version.Version=$VERSION'"
 
 go build -o ./bin/cli -ldflags="$LDFLAGS" interface/cli/main.go
-./bin/cli $@
+
+mesg_port=50052
+endpoint="localhost:$mesg_port"
+if [[ $DOCKER_HOST != "" ]]; then
+  endpoint="$(echo $DOCKER_HOST | awk '{while ( match($0,/([0-9]+\.){3}[0-9]+/) ) { print substr($0,RSTART,RLENGTH); $0=substr($0,RSTART+RLENGTH) } }'):$mesg_port"
```
NicolasMahe: I feel this regex as super dirty..

Why not using `docker-machine ip default`?
source: https://docs.docker.com/machine/get-started/

antho1404: It just feel more natural to extract the value of `DOCKER_HOST` if we test for it but I agree using docker-machine is way simpler.
Fixed

PR #847, opened 2019-04-02
api: simplify the code & switch to container mocks for testing

ilgooz:
*  refactored to simplify the code.
* improved testing helpers.
* improved some tests by testing public funcs instead of the inner(private) implementations. we were going to do this improvement at some point but wasn't the subject of this commit. it's started here now to keep the tests of the removed private funcs. this improvement will be applied to the rest of the existing tests later.
* use container mocks instead of dockertest for testing.
* removed unnecessary tests: TestCheckService, TestCreateTempFolder, TestRemoveTempFolder.
* improved some comments.
* simplify the file structure.

similar to https://github.com/mesg-foundation/core/pull/799 about the refactoring of api pkg.

please review each commit individually.


api/deploy.go
```diff
@@ -72,6 +47,31 @@ type DeployStatus struct {
...
+	return newServiceDeployer(a, env, options...).FromArchive(r)
+}
+
+// DeployServiceFromURL deploys a service living at a Git host.
+// Supported URL types:
+// - https://github.com/mesg-foundation/service-ethereum
+// - https://github.com/mesg-foundation/service-ethereum#branchName
+func (a *API) DeployServiceFromURL(url string, env map[string]string, options ...DeployServiceOption) (*service.Service,
+	*importer.ValidationError, error) {
+	return newServiceDeployer(a, env, options...).FromURL(url)
```
antho1404: Why not refactoring this one as well with just `DeployServiceFromURL` and `DeployService` that uses directly the content of `FromURL` and `FromArchive`?

api/api_test.go
```diff
@@ -15,26 +14,36 @@ const (
 	execdbname    = "exec.db.test"
 )
 
-func newAPIAndDockerTest(t *testing.T) (*API, *dockertest.Testing, func()) {
-	dt := dockertest.New()
+type apiTesting struct {
+	*testing.T
+	serviceDB   *database.LevelDBServiceDB
+	executionDB *database.LevelDBExecutionDB
+	cm          *mocks.Container
```
antho1404: can we rename it `container` it's hard to know `cm` in the other tests

api/api.go
```diff
@@ -37,3 +45,165 @@ func ContainerOption(container container.Container) Option {
...
+	}
+	if err = a.execDB.Save(exec); err != nil {
+		return "", err
+	}
+	go pubsub.Publish(s.TaskSubscriptionChannel(), exec)
+	return exec.ID, nil
+}
+
+// ExecuteAndListen executes given task and listen for result.
+func (a *API) ExecuteAndListen(serviceID, task string, inputs map[string]interface{}) (*execution.Execution, error) {
```
antho1404: I'm not sure this is used anywhere and there is a big hack on the `time.Sleep` so I would recommend to remove it

api/deploy.go
ilgooz: yeah, we can do that and keep the `newServiceDeployer` since it's used in both places

ilgooz: done

api/api.go
ilgooz: done

api/api_test.go
ilgooz: renamed to `containerMock`

PR #849, opened 2019-04-02
Use base58 to store hash

krhubert:
In order to avoid panic, you may remove old database servcies.


api/deploy_test.go
```diff
@@ -30,10 +30,9 @@ func TestDeployService(t *testing.T) {
 		archive, err := xarchive.GzippedTar(path, nil)
 		require.NoError(t, err)
 
-		service, validationError, err := a.DeployService(archive, nil, DeployServiceStatusOption(statuses))
+		_, validationError, err := a.DeployService(archive, nil, DeployServiceStatusOption(statuses))
```
antho1404: why removing the check of the service in this tests ?

service/service.go
```diff
@@ -120,7 +120,7 @@ func New(tarball io.Reader, env map[string]string, options ...Option) (*Service,
 	if err != nil {
 		return nil, err
 	}
-	s.Hash = hex.EncodeToString(hash)
+	s.Hash = base58.Encode(hash)
```
antho1404: I'm not sure here if the idea was to store the bytes in the raw data and just have the representation for docker and for the grpc api.
I'm fine with that, just want to have some clarification about that

krhubert: Storing has as bytes is a different case and should be done in different PR.

api/deploy_test.go
krhubert: Because it wasn't necessary here - it's not responsible of api to check what length has hash

service/namespace.go
```diff
@@ -14,7 +15,7 @@ const resultChannel string = "Result"
 
 // namespace returns the namespace of the service.
 func (service *Service) namespace() []string {
-	h, err := hex.DecodeString(service.Hash)
+	h, err := base58.Decode(service.Hash)
 	if err != nil {
 		panic(err)
```
antho1404: can't we simplify this method and solve this panic at the same time?

If I remember we were decoding the and re-encoding in sha1 just because of the length that was too long for docker but here base58 is short enough so this could be removed.

Also we have `service.Hash` which is already encoded in base58 so this function could simply be `return service.Hash`, maybe even `service.Hash[:xxx]` with xxx the maximum length allowed by docker and in this case we simplify this function and we don't need to delete the previous services (just they might have conflicts if you are really unlucky)

krhubert: we can but as I remember it required keeping hash as bytes first (and it's not the case in this PR)

antho1404: not sure how it is related to the bytes here. I'm just saying that here we are decoding and re-encoding, we could send the encoded value directly (with a limit on the size).
```go
  return service.Hash[:xxx]
}
```

With the hash as byte it will require encoding here but for this PR we already have the encoded version in database

krhubert: you can't take part in the hash as it's not good practice (and it doesn't guarantee uniqueness)  - this is why sha1 sum is calculated here and not taking a substring of hash.

krhubert: Done

antho1404: I still think that we could go a step further with like I say just returning the Hash especially that this is what we are doing on the latest release
https://github.com/mesg-foundation/core/blob/348aa982644ff31e9571813b298919dd99ed1c85/service/namespace.go#L13-L15
So this will not break anything and we don't have to cap anything because that was never released

krhubert: We can't because of service.Hash will be too long for docker label (this is why we using sha1 here)

antho1404: this is the namespace of the service anyway not docker. Also right now the hash for docker is something like

`core_hashxxxxxxx_service`

And I think sha1 is 40 characters and base58 is 44 so this is short enough and anyway this sha for docker should be done in docker but that's a separated PR.

PR #850, opened 2019-04-03
Split proto service

antho1404:
A first step to the discussion yesterday about having a proto definition for the service that we use everywhere and that is totally independent of the local environment.

Here I splitted the service into it's own proto definition and update the apis:
- ListServices
- GetService

They return now a `ServiceDetail` that contains 2 data:
- Status
- Service

I'm not really fan of the naming for the api so feel free to propose something better


antho1404:
Thanks for the feedback, I agree naming was bad. I was able to change few things:
- `definitions/service.proto` and `coreapi/api.proto` does not share the same proto package anymore
- `service.proto` has its own proto package `definitions`
- `api.proto` include `definitions` package and uses now `definitions.Service` instead of just `Service`
- Now `api.proto` can have its own `Service` message
- New API is `Service { Status: enum, Definition: definitions.Service }` so we can access like `response.service.definition` 

Only problem with that is `definitions/service.proto` has to define the go package `option go_package = "github.com/mesg-foundation/core/protobuf/definitions";` in order to generate correctly the `pb.go` file but it also automatically export the `service.pb.go` in this directory so I added a hack to mv the file after generated in the `build-proto.sh`

NicolasMahe:
> Only problem with that is `definitions/service.proto` has to define the go package `option go_package = "github.com/mesg-foundation/core/protobuf/definitions";` in order to generate correctly the `pb.go` file but it also automatically export the `service.pb.go` in this directory so I added a hack to mv the file after generated in the `build-proto.sh`

Fix in https://github.com/mesg-foundation/core/pull/850/commits/b09e24f679e228613a10696c6675cf8eb9f7bbd9

protobuf/definitions/service.proto
```diff
@@ -0,0 +1,62 @@
+syntax = "proto3";
+
+package definitions;
```
NicolasMahe:
```suggestion
package definition;
```

Why the plural on the package name? @ilgooz suggested a singular and I also prefer the singular.

utils/servicecasting/cast_test.go
```diff
@@ -109,15 +109,15 @@ func TestServiceCast(t *testing.T) {
 }
 
 // createTestServcieWithInputs creates test service with given inputs name and type under "test" task key.
-func createTestServcieWithInputs(inputs map[string]string) *coreapi.Service {
-	s := &coreapi.Service{
-		Tasks: []*coreapi.Task{
+func createTestServcieWithInputs(inputs map[string]string) *definitions.Service {
```
NicolasMahe:
```suggestion
func createTestServiceWithInputs(inputs map[string]string) *definitions.Service {
```

PR #851, opened 2019-04-05
Add args and env to proto definition of Dependency

NicolasMahe:
Dependant on https://github.com/mesg-foundation/core/pull/850


antho1404:
I'm not sure about this one. I see that it's interesting to share all the data but I feel it can be really risky to expose the env variables used for the deployment.
Here it mean we can read the env of service just by reading the api.

NicolasMahe:
> I'm not sure about this one. I see that it's interesting to share all the data but I feel it can be really risky to expose the env variables used for the deployment.
> Here it mean we can read the env of service just by reading the api.

That's right, I didn't think about this ;)

But the `args` should be present.

Can you partially merge this one to yours?

antho1404:
done in this commit https://github.com/mesg-foundation/core/pull/850/commits/1d543690b36aa9e21cd66d4c452451cefeda2e39

PR #853, opened 2019-04-05
Update marketplace cli commands with last version of service marketplace

NicolasMahe:
Dependant on https://github.com/mesg-foundation/core/pull/844

Output of commands look like:
```
➜ ./dev-cli marketplace purchase influxdb
? Enter the account to use 0x978369695a3758abc8c0706a9b7cee4b1cb9652c
? Enter passphrase *
? Enter the offer index to purchase 0
? Are you sure to purchase service "influxdb" for 20 MESG Tokens and a duration of 60 seconds? Yes
✔ Offer index "0" of service "influxdb" purchased with success and expires on 2019-04-05 18:45:24 +0700 +07
```
```
➜ ./dev-cli marketplace create-offer com.mesg.bitcoin
? Enter the account to use 0x978369695a3758abc8c0706a9b7cee4b1cb9652c
? Enter passphrase *
? Enter the price (in MESG Token) of the offer 1000
? Enter the duration (in second) of the offer 30
? Are you sure to create offer for service "com.mesg.bitcoin" with a price of 1000 MESG Tokens and a duration of 30 seconds? Yes
✔ Offer created with success with index "0"
✔ See it on the marketplace: https://marketplace.mesg.com/services/com.mesg.bitcoin#offers
```
```
➜ ./dev-cli marketplace publish ../services/service-bitcoin
? Enter the account to use 0x978369695a3758abc8c0706a9b7cee4b1cb9652c
? Enter passphrase *
✔ Image built with success
✔ Service deployed with sid com.mesg.bitcoin and hash 55sdquWL8U6sujUyRjYS5xQFiSdGFemPjbtcYkdQXRgw
? Are you sure to publish a new version of service "com.mesg.bitcoin" from path "../services/service-bitcoin" using account "0x978369695a3758abc8c0706a9b7cee4b1cb9652c"? Yes
✔ Service published with success with sid "com.mesg.bitcoin" and marketplace hash "HNybfCZZw6uKSArfPjZbeF6R7FJznGLFC8dr3SvcFS13"
✔ See it on the marketplace: https://marketplace.mesg.com/services/com.mesg.bitcoin/HNybfCZZw6uKSArfPjZbeF6R7FJznGLFC8dr3SvcFS13
✔ To create a service offer, execute the command:
	mesg-core marketplace create-offer com.mesg.bitcoin
```


commands/provider/marketplace_provider.go
```diff
@@ -18,46 +19,67 @@ func NewMarketplaceProvider(c coreapi.CoreClient) *MarketplaceProvider {
...
+// PublishPublishServiceVersion executes the task publish service version task
+func (p *MarketplaceProvider) PublishPublishServiceVersion(signedTransaction string) (sid, versionHash, manifest, manifestProtocol string, err error) {
+	input := marketplacePublishTaskInputs{
+		SignedTransaction: signedTransaction,
 	}
-	var output Transaction
-	return output, p.call("createServiceOffer", input, &output)
+	var o marketplacePublishPublishServiceVersionTaskOutputs
+	err = p.call("publishPublishServiceVersion", input, &o)
+	return o.Sid, o.VersionHash, o.Manifest, o.ManifestProtocol, err
```
antho1404: Why not keeping the same pattern in the code. You could expose `marketplacePublishPublishServiceVersionTaskOutputs` and have something like that
```go
var o marketplacePublishPublishServiceVersionTaskOutputs
return o, p.call("publishPublishServiceVersion", input, &o)
```
I feel a function that returns too much data should return a struct.

Same for the `PublishCreateServiceOffer` and `PublishPurchase`

NicolasMahe: Yes, the function has too many returned value but:
- the CLI doesn't directly depend on the package provider, all the function are declared in commands.go
- the commands actually only read 2 or 3 returned values. Should I only returned value useful for the commands?

antho1404: > the CLI doesn't directly depend on the package provider, all the function are declared in commands.go

It does on few functions already: `UploadSource`, `PreparePublishServiceVersion`, `PrepareCreateServiceOffer`, `PreparePurchase`, `GetService`... basically all except these ones

> it returns only simple types

No it doesn't:
```
PreparePublishServiceVersion(service provider.MarketplaceManifestServiceData, from string) (provider.Transaction, error)
GetService(sid string) (provider.MarketplaceService, error)
```

> the commands actually only read 2 or 3 returned values. Should I only returned value useful for the commands?
 
If we need only these data we could just return them but if we have the struct let's return the struct like that we have all the data without having a long list of outputs

krhubert:
```
diff --git a/systemservices/marketplace/mesg.yml b/systemservices/marketplace/mesg.yml
index 1452a275..c1da8997 100644
--- a/systemservices/marketplace/mesg.yml
+++ b/systemservices/marketplace/mesg.yml
@@ -3,7 +3,7 @@ sid: marketplace
 description: ""
 configuration:
   env:
-    - PROVIDER_ENDPOINT=https://ropsten.infura.io/v3/6115ae2531f64c04a9a392cf500e5fbe
+    - PROVIDER_ENDPOINT=https://localhost:1234
```

If you change ENDPOINT then marketplace stucks on 

```
⣯ Publishing service on the marketplace...
```

Forever (or maybe there is some timeout but it would be strange to have timeout to nonexisting port)

NicolasMahe:
> ```
> diff --git a/systemservices/marketplace/mesg.yml b/systemservices/marketplace/mesg.yml
> index 1452a275..c1da8997 100644
> --- a/systemservices/marketplace/mesg.yml
> +++ b/systemservices/marketplace/mesg.yml
> @@ -3,7 +3,7 @@ sid: marketplace
>  description: ""
>  configuration:
>    env:
> -    - PROVIDER_ENDPOINT=https://ropsten.infura.io/v3/6115ae2531f64c04a9a392cf500e5fbe
> +    - PROVIDER_ENDPOINT=https://localhost:1234
> ```
> If you change ENDPOINT then marketplace stucks on
> 
> ```
> ⣯ Publishing service on the marketplace...
> ```
> Forever (or maybe there is some timeout but it would be strange to have timeout to nonexisting port)

Can you check the log of the service? `./dev-cli service logs marketplace`.
I think the service throw an error and actually never properly start. This is not an issue only related to this service but to all service. The core doesn't check if the service is properly running before sending a task.

krhubert:
Yes it throws error, but at the same time

`EQcbh3CqmjdxFatGao959tc5VQkX3j7DJToMYvQoR2VK    marketplace                                      Marketplace        running `

Is still running, so, in that case, can task return error as well, or service should be killed (and it will be seen as stopped )

PR #855, opened 2019-04-05
simplify service pkg & make types logicless

ilgooz:
These are some preparations to simply code before refactoring and having a `Manager` inside `service` pkg.

We'll be having the same `Manager` interface introduced in #799 in the next PRs except that I plan to remove `Deploy()` func from it. Because Manager is about managing containers and its implementation will vary depending on which container orchestration system we want to support. So, Manager shouldn't be responsible for "deploying" or in another means, "building" the Docker image. It's a native functionally.

The very next PR will be about moving the implementation of `Deploy()` to `api` pkg. In future, it can be moved to somewhere else, to a new package if needed.

Please review commits individually.


interface/grpc/server.go
```diff
@@ -20,7 +21,8 @@ import (
 
 // Server contains the server config.
 type Server struct {
-	api *api.API
+	container container.Container
```
antho1404: I don't think this is good. The API already has the container so there is no reason for the Server to have it too.
2 solutions: 
- The server has no container information and we need a `Status` function in the API 👍 
- The server only has the container and not the api and the server give the container to the api when executing if needed. We could have different container system configurable from the api

I would definitely choose the first one as the second we don't have multiple containers system and i'm not even sure we will want to configure it from the api

interface/grpc/core/type.go
```diff
@@ -2,14 +2,16 @@ package core
...
 type Server struct {
-	api *api.API
+	container container.Container
+	api       *api.API
 }
 
 // NewServer creates a new Server.
-func NewServer(api *api.API) *Server {
-	return &Server{api: api}
+func NewServer(c container.Container, api *api.API) *Server {
```
antho1404: Same comment as interface/grpc/server.go

service/volume.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 // DeleteVolumes deletes the data volumes of service and its dependencies.
-func (s *Service) DeleteVolumes() error {
+func (s *Service) DeleteVolumes(c container.Container, ss *Service) error {
```
antho1404: we already have the service here `func (s *Service)`

interface/grpc/server.go
ilgooz: yes i agree. this is something that we'll refactor in future. either by going with the option 1 of yours or have `grpc.GetService()` to return status by default or optionally. let's do not add more to this PR.

interface/grpc/core/type.go
ilgooz: https://github.com/mesg-foundation/core/pull/855#discussion_r272795971

service/volume.go
ilgooz: done

interface/grpc/server.go
antho1404: I think this has to be done in this PR.
This PR is adding an unnecessary dependency that actually adds even more review in the code.
With the PR https://github.com/mesg-foundation/core/pull/858 many files from this PR will be untouched or the same piece of code will be changed anyway.
I really don't want to add a new dependency in this grpc api that is not necessary just for one single function that should anyway be in the api

antho1404:
@ilgooz can you fix the conflicts in order to validate.

service/service.go
```diff
@@ -201,27 +133,26 @@ func (s *Service) deploy() error {
...
 			Type:    typ,
 		}
 	}
 }
 
 // closeStatus closes statuses chan.
-func (s *Service) closeStatus() {
-	if s.statuses != nil {
-		close(s.statuses)
+func (s *Service) closeStatus(statuses chan DeployStatus) {
```
NicolasMahe: [NOT BLOCKING] Same here

api/deploy.go
```diff
@@ -133,28 +131,28 @@ func (d *serviceDeployer) preprocessArchive(r io.Reader, path string) error {
...
 		path = filepath.Join(path, dirs[0].Name())
 	}
 
-	return archive.Tar(path, archive.Uncompressed)
+	return path, nil
 }
 
 // deploy deploys a service in path.
-func (d *serviceDeployer) deploy(r io.Reader) (*service.Service, *importer.ValidationError, error) {
+func (d *serviceDeployer) deploy(path string) (*service.Service, *importer.ValidationError, error) {
```
NicolasMahe: [NOT BLOCKING] This modif should have been done in another PR

core/main.go
```diff
@@ -41,8 +43,14 @@ func initDependencies() (*dependencies, error) {
...
 
+	// init container.
+	c, err := container.New()
+	if err != nil {
+		return nil, err
+	}
+
 	// init api.
-	api, err := api.New(serviceDB, executionDB)
+	api, err := api.New(serviceDB, executionDB, api.ContainerOption(c))
```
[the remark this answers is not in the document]
ilgooz: yes, that change is already in the next PR. https://github.com/mesg-foundation/core/pull/856

service/service.go
```diff
@@ -132,64 +105,23 @@ func New(tarball io.Reader, env map[string]string, options ...Option) (*Service,
...
+	s.sendStatus(statuses, "Building Docker image...", DRunning)
 
-	imageHash, err := s.container.Build(s.tempPath)
+	imageHash, err := c.Build(contextDir)
 	if err != nil {
 		return err
 	}
 
-	s.sendStatus("Image built with success", DDonePositive)
+	s.sendStatus(statuses, "Image built with success", DDonePositive)
```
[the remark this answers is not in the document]
ilgooz: deployment will be moved to `api` pkg in a following PR. this func will be removed anyway

service/service.go
ilgooz: https://github.com/mesg-foundation/core/pull/855#discussion_r273485889

api/deploy.go
ilgooz: in that case we couldn't have remove `tempPath` private field from `Service` type which is the actual purpose of this PR. we needed to make `Service` type logicless by removing all private fields

PR #860, opened 2019-04-09
Store configuration in service as a separated attribute

antho1404:
Move the configuration from the list of dependencies in the service structure to its own attribute.

This still fallback correctly to previous services (with configuration in the dependencies)

Still few questions:
- Does the configuration need to expose the image in the protobuf definition?
- Should we expose the env variables (maybe not the value but at least the keys)?


protobuf/definition/service.proto
```diff
@@ -51,6 +52,14 @@ message Parameter {
   repeated Parameter object = 10; // Optional object structure definition when type is set to `Object`
 }
 
+message Configuration {
+  repeated string volumes = 1;      // List of volumes.
+  repeated string volumesfrom = 2;  // List of volumes mounted from other dependencies.
```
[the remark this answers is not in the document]
antho1404: this is not related to the configuration, will be done in another pr

service/stop.go
```diff
@@ -24,17 +24,22 @@ func (s *Service) StopDependencies(c container.Container) error {
...
+	stop := func(d *Dependency) {
+		defer wg.Done()
+		errStop := d.Stop(c, s)
+		mutex.Lock()
+		defer mutex.Unlock()
+		if errStop != nil && err == nil {
+			err = errStop
+		}
+	}
+	if s.Configuration != nil {
```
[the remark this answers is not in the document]
antho1404: yes that's what it is
```go
go stop(s.Configuration)
for _, dep := range s.Dependencies {
  go stop(dep)
}
```

protobuf/definition/service.proto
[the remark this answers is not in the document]
antho1404: done in this PR https://github.com/mesg-foundation/core/pull/866

service/start.go
```diff
@@ -19,7 +19,7 @@ func (s *Service) Start(c container.Container) (serviceIDs []string, err error)
 	}
 	// If there is one but not all services running stop to restart all
 	if status == PARTIAL {
-		if err := s.StopDependencies(c); err != nil {
+		if err := s.Stop(c); err != nil {
```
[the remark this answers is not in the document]
antho1404: reverted and moved to https://github.com/mesg-foundation/core/pull/867

service/service.go
```diff
@@ -151,6 +157,9 @@ func (s *Service) closeStatus(statuses chan DeployStatus) {
 
 // getDependency returns dependency dependencyKey or a not found error.
 func (s *Service) getDependency(dependencyKey string) (*Dependency, error) {
+	if dependencyKey == MainServiceKey {
```
ilgooz: this pr is about moving service configuration out of dependencies. it doesn't make much sense this func to return the service configuration

service/start.go
```diff
@@ -29,14 +29,22 @@ func (s *Service) Start(c container.Container) (serviceIDs []string, err error)
 	}
 	// BUG: https://github.com/mesg-foundation/core/issues/382
 	// After solving this by docker, switch back to deploy in parallel
-	serviceIDs = make([]string, len(s.Dependencies))
-	for i, dep := range s.Dependencies {
+	serviceIDs = make([]string, 0)
+	for _, dep := range s.Dependencies {
 		serviceID, err := dep.Start(c, s, networkID)
```
ilgooz: same pattern https://github.com/mesg-foundation/core/pull/860/files#diff-be8c3a41ba62becfd4d278926d06a683R27 can be applied here by having a `start := func()` above. but not blocking for this pr.

service/service.go
antho1404: this simplifies a lot the `extractVolumesFrom` function. I agree it's not ideal but will add quite a lot of complexity otherwise

service/start.go
antho1404: done

service/service.go
antho1404: also this ensures that previous services (the one with still the service in dependencies) can run

antho1404: this logic has been moved to the volumefrom

PR #863, opened 2019-04-09
Improve marketplace mesg.yml definition

NicolasMahe:
Better to check commit by commit.

Modifs:
- remove empty `name: ""` and `description: ""` (will add them back with content later).
- Use definition of `manifestServiceData` in `preparePublishServiceVersion`
- Add a ~~basic~~ full definition of service definition


antho1404:
Any reason why the "configuration" is not part of the mesg.yml or is it just something you forgot ?
configuration has been extracted from dependencies

NicolasMahe:
> Any reason why the "configuration" is not part of the mesg.yml or is it just something you forgot ?
> configuration has been extracted from dependencies

Nice catch. This PR has been done before the extraction of configuration ;)

systemservices/marketplace/mesg.yml
```diff
@@ -11,129 +11,107 @@ configuration:
...
+                          hashVersion:
+                            type: String
+                          deployment:
+                            type: Object
+                            object:
+                              type:
+                                type: String
+                              source:
+                                type: String
+                          definition:
```
ilgooz: I don't think we should add a "partial" validation for _definition_. 
to me, it's an all-or-nothing thing.

NicolasMahe: Done

systemservices/marketplace/mesg.yml
```diff
@@ -72,17 +72,99 @@ tasks:
...
+                                      type:
+                                        type: String
+                                      optional:
+                                        type: Boolean
+                                        optional: true
+                                      repeated:
+                                        type: Boolean
+                                        optional: true
+                                      object:
+                                        type: Object
```
ilgooz: we could have typed Parameter here as well if we implemented this feature https://forum.mesg.com/t/service-definition-support-recursive-types/260/6

PR #866, opened 2019-04-10
Rename volumefrom

antho1404:
depending on https://github.com/mesg-foundation/core/pull/860
fixes https://github.com/mesg-foundation/core/issues/865

Renaming `volumefrom` into `volumeFrom` in the protobuf definition


protobuf/definition/service.proto
```diff
@@ -9,6 +9,7 @@ message Service {
   string sid = 12;                          // Service's sid.
   string name = 1;                          // Service's name.
   string description = 2;                   // Service's description.
+  Configuration configuration = 8;          // Configurations related to the service
```
krhubert: This is not related to this PR - it should be done in different PR.

antho1404: that's the case, look at the description of the PR, depending on #860

PR #868, opened 2019-04-10
Change ipfs gateway to ipfs.app.mesg.com

krhubert:


antho1404:
I can't make it work.

Screen Shot 2019-04-11 at 1 38 53 pm

NicolasMahe:
@krhubert can you make it work? ipfs seems to not run on the server

NicolasMahe:
@krhubert any news?

krhubert:
ipfs is up && running - should be fine now.

```
➜  core git:(feature/ipfs-gateway) curl -F file=@tools.go http://ipfs.app.mesg.com:5001/api/v0/add 
{"Name":"tools.go","Hash":"Qmb5WZZgYJCNb6Jw588ysKW1X9x8bVR7ktEWxGEHNhnmsv","Size":"416"}
```

krhubert:
we discussed enabling https and it's unnecessary for now (ipfs node is public and everyone can download the full content of it so https, in that case, doesn't give anything (expected waisted time on setting SSL)

krhubert:
> Do we know the reason why the node ipfs was down.

Yes, not enough memory, I've already increased it

PR #869, opened 2019-04-10
Validate env variable

krhubert:


service/importer/service_file.go
```diff
@@ -105,6 +105,15 @@ func newValidator() (*validator.Validate, ut.Translator) {
 	uni := ut.New(en, en)
 	trans, _ := uni.GetTranslator("en")
 	validate := validator.New()
+
+	validate.RegisterValidation("env", xvalidator.IsEnv)
+	validate.RegisterTranslation("env", trans, func(ut ut.Translator) error {
+		return ut.Add("portmap", "{0} must be a valid env variable name", false)
```
antho1404: why `portmap` here ?
```suggestion
		return ut.Add("env", "{0} must be a valid env variable name", false)
```

x/xvalidator/validator.go
```diff
@@ -13,8 +14,11 @@ const (
 	maxPort = 65535
 
 	portSeparator = ":"
+	envSeparator  = "="
 )
 
+var envNameRegexp = regexp.MustCompile("^[a-zA-Z][a-zA-Z0-9_]*$")
```
antho1404: we could enforce capital and _ only

ilgooz: numbers are ok

krhubert: > we could enforce capital and _ only

this is valid regex for env variable:
- numbers are allowed (but not at 1st pos)
- all letter cases are allowed - so some people use lower and some upper cases and all of them are valid

I don't see the value of changing standard env naming. If there is some we can consider it, otherwise let's keep to what devs used to.

PR #873, opened 2019-04-11
moved service deployment logic from service pkg to api pkg

ilgooz:
there are no logical changes made on code


service/inject_definition.go
```diff
@@ -7,8 +7,8 @@ import (
 	"github.com/mesg-foundation/core/x/xstrings"
 )
 
-// injectDefinition applies service definition to Service type.
-func (s *Service) injectDefinition(def *importer.ServiceDefinition) {
+// InjectDefinition applies service definition to Service type.
+func (s *Service) InjectDefinition(def *importer.ServiceDefinition) {
```
antho1404: This function should not be exported. It's an internal function that changes a lot the data. It should be controlled from inside not outside.
One solution could be to move that directly in the deploy api, the deploy is responsible to populate this service

service/service.go
```diff
@@ -159,13 +60,13 @@ func (s *Service) getDependency(dependencyKey string) (*Dependency, error) {
 	return nil, fmt.Errorf("dependency %s do not exist", dependencyKey)
 }
 
-// validateConfigurationEnv checks presence of env variables in mesg.yml under env section.
-func (s *Service) validateConfigurationEnv(env map[string]string) error {
+// ValidateConfigurationEnv checks presence of env variables in mesg.yml under env section.
+func (s *Service) ValidateConfigurationEnv(env map[string]string) error {
```
antho1404: Same comment, this one could be quite ok because it doesn't have any impact on the data but could be moved to the deploy api

service/inject_definition.go
ilgooz: done. I was thinking about doing these improvements later after seeing what service pkg will become to when refactoring is completed. but it's fine

Issue #838, opened 2019-03-24
Marketplace error management


antho1404:
The marketplace service should catch errors from ethereum, and should never return this kind of error

<img width="692" alt="Screen Shot 2019-03-24 at 11 19 55 pm" src="https://user-images.githubusercontent.com/1783126/54882336-80cf4a00-4e8b-11e9-8754-e65109ddc62c.png">

This error is when the same service is published twice

NicolasMahe:
There is no way to get the error from a transaction receipt..
More pre-check could be added though.

NicolasMahe:
todos from #874:
- [ ] [preparePublishServiceVersion.ts] add check on SID is domain name (see go implementation)
- [ ] [preparePublishServiceVersion.ts] check if version already exist on marketplace

PR #874, opened 2019-04-12
Marketplace improve checks

NicolasMahe:
Related to https://github.com/mesg-foundation/core/issues/838

Move a lots of checks from CLI to the service itself.

Todo that can be done in other PR:
- [x] [preparePublishServiceVersion.ts] add check on SID is domain name (see go implementation)
- [x] [preparePublishServiceVersion.ts] check if version already exist on marketplace
Edit: done in https://github.com/mesg-foundation/core/pull/899


commands/marketplace_purchase.go
```diff
@@ -52,24 +52,17 @@ func (c *marketplacePurchaseCmd) preRunE(cmd *cobra.Command, args []string) erro
 		}
 	}
 
-	pretty.Progress("Getting service data...", func() {
+	pretty.Progress("Getting offer data...", func() {
```
NicolasMahe: It's still getting the offer data to display the price and duration in the confirmation prompt.

commands/marketplace_purchase.go
```diff
@@ -52,24 +52,17 @@ func (c *marketplacePurchaseCmd) preRunE(cmd *cobra.Command, args []string) erro
...
 	}
-
 	offerIndex, err := strconv.Atoi(c.offerIndex)
 	if err != nil {
 		return err
 	}
-	if offerIndex >= len(service.Offers) {
-		return fmt.Errorf("offer index %d doesn't exist", offerIndex)
-	}
 	offer := service.Offers[offerIndex]
```
antho1404: what if I put an index that is out of range? offer will be nil and the print will cash i guess

systemservices/marketplace/src/tasks/prepareCreateServiceOffer.ts
```diff
@@ -1,16 +1,32 @@
...
+    if (duration.isNegative() || duration.isZero()) throw new Error('duration cannot be negative or equal to zero')
+
+    // check service
+    const service = await getService(marketplace, inputs.sid)
+
+    // check ownership
+    if (service.owner.toLowerCase() !== inputs.from.toLowerCase()) throw new Error(`service's owner is different`)
+
+    // check service version
+    if (service.versions.length === 0) throw new Error('cannot create an offer on a service with 0 version')
```
antho1404:
```suggestion
    if (service.versions.length === 0) throw new Error('cannot create an offer on a service with no version')
```

systemservices/marketplace/src/tasks/preparePublishServiceVersion.ts
```diff
@@ -1,19 +1,38 @@
...
+import { getService, isServiceExist } from "../contracts/service";
 
 export default (
   marketplace: Marketplace,
   createTransaction: CreateTransaction
 ) => async (inputs: TaskInputs, outputs: TaskOutputs): Promise<void> => {
   try {
+    // check inputs
+    const sid = inputs.service.definition.sid
+    if (!(1 <= sid.length && sid.length <= 63)) throw new Error('sid must have a length between 1 and 63')
```
antho1404: the `63` should be in a constant somewhere to know what is this 63
the condition is a bit complicated. Either invert it `sid.length < 1 || sid.length > 63` or even 2 if with their own message, one for min length, one for max length

commands/marketplace_purchase.go
NicolasMahe: I put back the test

systemservices/marketplace/src/tasks/preparePublishServiceVersion.ts
NicolasMahe: I simplify the test but leave the 63 hardcoded because the go implementation is also hardcoded.
Anyway I think this limitation will be removed in the future, so I don't want to spend too much time on this test.

antho1404: > because the go implementation is also hardcoded

haha, this is the worst argument ever but anyway it's not blocking at all

systemservices/marketplace/src/tasks/preparePublishServiceVersion.ts
```diff
@@ -1,19 +1,39 @@
...
+import * as assert from "assert";
 
 export default (
   marketplace: Marketplace,
   createTransaction: CreateTransaction
 ) => async (inputs: TaskInputs, outputs: TaskOutputs): Promise<void> => {
   try {
+    // check inputs
+    const sid = inputs.service.definition.sid
+    assert.ok(sid.length >= 1 && sid.length <= 63, 'sid must have a length between 1 and 63') // See Core service validation (https://github.com/mesg-foundation/core)
```
ilgooz:
```suggestion
    assert.ok(sid.length >= 1 && sid.length <= 63, 'sid's length must be 1 at min and 63 at max') // See Core service validation (https://github.com/mesg-foundation/core)
```

between word is kind of confusing here because grammatically it says that sid's length cannot be 1 nor 63 it should only in between. and this is not correct

systemservices/marketplace/src/tasks/preparePublishServiceVersion.ts
```diff
@@ -1,19 +1,39 @@
...
+import * as assert from "assert";
 
 export default (
   marketplace: Marketplace,
   createTransaction: CreateTransaction
 ) => async (inputs: TaskInputs, outputs: TaskOutputs): Promise<void> => {
   try {
+    // check inputs
+    const sid = inputs.service.definition.sid
+    assert.ok(sid.length >= 1 && sid.length <= 63, 'sid's length must be 1 at min and 63 at max') // See Core service validation (https://github.com/mesg-foundation/core)
```
ilgooz:
```suggestion
    assert.ok(sid.length >= 1 && sid.length <= 63, 'sid\'s length must be 1 at min and 63 at max') // See Core service validation (https://github.com/mesg-foundation/core)
```

systemservices/marketplace/src/tasks/preparePurchase.ts
```diff
@@ -13,12 +13,26 @@ export default (
...
   try {
+    // get service
+    const service = await getService(marketplace, inputs.sid)
+    
+    // check ownership
+    if (service.owner.toLowerCase() !== inputs.from.toLowerCase()) throw new Error(`service's owner is different that the specified 'from'`)
+
     // get offer data
-    const offer = await getServiceOffer(marketplace, inputs.sid, new BigNumber(inputs.offerIndex))
+    const offerIndex = new BigNumber(inputs.offerIndex).toNumber()
```
ilgooz: should check for negative numbers as well

NicolasMahe: It's checked at the next line

ilgooz: it only checks for bigger indexes but how about `...-2, -1`?

ilgooz: secondly, _toNumber()_ can return a float value number too. we need to make sure that we're using integer inside arrays. gotto check this in other similar places too. there is `integerValue ()` func inside `BigNumber` maybe it can suit here.

systemservices/marketplace/src/tasks/prepareCreateServiceOffer.ts
```diff
@@ -1,16 +1,28 @@
...
 ) => async (inputs: TaskInputs, outputs: TaskOutputs): Promise<void> => {
   try {
+    // check inputs
+    const duration = new BigNumber(inputs.duration)
+    if (duration.isNegative() || duration.isZero()) throw new Error('Duration cannot be negative or equal to zero')
+
+    // check service
+    const service = await getService(marketplace, inputs.sid)
+    if (service.owner.toLowerCase() !== inputs.from.toLowerCase()) throw new Error(`Service's owner is different that the specified 'from'`)
+    if (service.versions.length === 0) throw new Error('Cannot create an offer on a service with 0 version')
```
ilgooz: a service must be published with at least 1 version at the beginning, why do we have this check here?

systemservices/marketplace/src/tasks/preparePurchase.ts
NicolasMahe: it's already done: https://github.com/mesg-foundation/core/pull/874/files#diff-ad93658a1e6d166abbf4848e45444a02R25

systemservices/marketplace/src/tasks/prepareCreateServiceOffer.ts
NicolasMahe: Because the smart contract can be called without using the service and doesn't prevent this case.

systemservices/marketplace/src/tasks/preparePurchase.ts
ilgooz: oh okay, I was reviewing the first commit related to this, didn't see this update yet ;) but we need to make sure to use integers for the second one.

systemservices/marketplace/src/tasks/prepareCreateServiceOffer.ts
ilgooz: how about updating the smart contact to check this?

systemservices/marketplace/src/tasks/preparePurchase.ts
NicolasMahe: Done in https://github.com/mesg-foundation/core/compare/ss/marketplace-improve-checks-more?expand=1

systemservices/marketplace/src/tasks/prepareCreateServiceOffer.ts
NicolasMahe: The current smart contract structure totally allow this.
Let's think about this for the next version of the smart contract.

PR #884, opened 2019-04-16
service: simplify Stop() & improve errors

ilgooz:


ilgooz:
> why did you remove the mutex here ? I'm not really sure why we had one in the first place maybe to fix some strange behaviors with docker when we stop containers or something like that

That mutex was there to prevent concurrent writes to `err`. But with the updated version, it's handled inside `errs.Append()`.

service/stop.go
```diff
@@ -13,39 +14,27 @@ func (s *Service) Stop(c container.Container) error {
...
-		errStop := d.Stop(c, s)
-		mutex.Lock()
-		defer mutex.Unlock()
-		if errStop != nil && err == nil {
-			err = errStop
+	var (
+		wg   sync.WaitGroup
+		errs xerrors.SyncErrors
+	)
+	for _, d := range append(s.Dependencies, s.Configuration) {
```
NicolasMahe: Configuration should be stopped first ;)

ilgooz: good catch!

PR #887, opened 2019-04-17
Add to ci task testNoCoverage to speed up review process

NicolasMahe:
I add a new task that run test without the coverage so it uses the go test cache to not run test that doesn't need to.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,6 +54,24 @@ jobs:
       - <<: *save_go_cache
       - run: bash <(curl -s https://codecov.io/bash)
 
+  "testNoCoverage":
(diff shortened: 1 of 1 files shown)


NicolasMahe:
It seems to cache only 3 packages...
@krhubert any suggestion to improve caching?

krhubert:
no idea have to check

NicolasMahe:
@krhubert please check because it will improve a lot the merge workflow! no need to wait 12min for the test to finish!

NicolasMahe:
Closing this PR as no solution has been found and as the Engine get lighter and lighter, we will find other way to speed up the tests.

PR #888, opened 2019-04-17
service: simplify Logs() & improve docs

ilgooz:


service/logs.go
```diff
@@ -14,36 +16,43 @@ type Log struct {
...
+			rerr, ew := io.Pipe()
+			go func(dstout, dsterr io.Writer, r io.ReadCloser) {
+				if _, err := stdcopy.StdCopy(dstout, dsterr, r); err != nil {
+					r.Close()
+					logrus.Errorln(err)
+				}
+			}(sw, ew, r)
 			if err != nil {
-				return err
+				return nil, err
```
NicolasMahe: There is no way that `if err != nil` will return true. This test can be removed.

ilgooz: done

NicolasMahe:
I have a question that is related to logs but not blocking for this PR.

https://github.com/mesg-foundation/core/blob/1ecd99db1c8c90a60231d77f8e857309c45ddf35/service/logs.go#L23-L58

Seeing this function, I don't see anything that will make the core to 'stop' listening the logs when let's say the user stops the command `service logs`. I think the go routine will continue to live forever..
Am I right or there is a mechanism that will stop the stream of logs?

ilgooz:
@NicolasMahe  goroutine will eventually be closed when [these](https://github.com/mesg-foundation/core/pull/888/files#diff-5e7c3981a052702ee7ff89a6212160daR42) writers closed inside [grpc core](https://github.com/mesg-foundation/core/blob/3f34196/interface/grpc/core/logs.go#L34-L35) pkg by the api cancellation.

but we should definitely improve it later to close the [reader](https://github.com/mesg-foundation/core/pull/888/files#diff-5e7c3981a052702ee7ff89a6212160daR42) as well.

_-i'm actually thinking about creating a custom `io.Closer` to simplify the process. not relevent with this but I plan to do even more and improve the chunker pkg to implement custom io.Writers/Readers and simplify log streams. these are some future refactoring plans._

because for now, `stdcopy` will only be able to cancel after reading from the reader and trying to write to closed writers(writers closed when their readers are closed -io.Pipe()-). but we can make it cancel sooner by closing the reader as well so it doesn't need to wait data to appear in the reader.

PR #891, opened 2019-04-17
Replace custom pubsub

krhubert:


api/api.go
```diff
@@ -22,6 +24,7 @@ type API struct {
 // New creates a new API with given options.
 func New(c container.Container, db database.ServiceDB, execDB database.ExecutionDB) *API {
 	return &API{
+		ps:        pubsub.New(0),
```
antho1404: 0 is the capacity of the pubsub. Is it safe to add 0, don't we have any problem when there is multiple subscribers on this pubsub?

krhubert: 0 is the capacity of the channels that are created by pubsub, so i n that case pubsub will create unbuffered channel (as previous version)

PR #893, opened 2019-04-18
Move mocks.go to internal package

krhubert:


krhubert:
I've tested it and it worked (mock that are not generated is executor)

Issue #861, opened 2019-04-09
Move marketplace's hashVersion in the core


NicolasMahe:
From comment on https://github.com/mesg-foundation/core/pull/845

antho1404:
closed with https://github.com/mesg-foundation/core/pull/906

PR #864, opened 2019-04-09
Marketplace remove hash from manifest

NicolasMahe:
Start to implements https://github.com/mesg-foundation/core/issues/861 for CLI and Service Marketplace:
only remove unnecessary hash for now

PR #894, opened 2019-04-18
Marketplace move hashVersion

NicolasMahe:
Dependant on https://github.com/mesg-foundation/core/pull/864.
Closes https://github.com/mesg-foundation/core/issues/861.

This PR move the `hashVersion` from the manifest data to the service definition. Like this, the core itself can manage the version used to calculate the service's hash.


NicolasMahe:
@mesg-foundation/core ready to review

NicolasMahe:
@antho1404 do you agree with this modification?

antho1404:
let's rename `hashVersion` into `version` this is a version of a service not only the hash

service/service.go
```diff
@@ -27,6 +27,9 @@ type Service struct {
 	// It represents the service uniquely.
 	Hash string `hash:"-"`
 
+	// HashVersion is the version of the algorithm used to calculate the of the hash
+	HashVersion string `hash:"-"`
```
antho1404: I don't think `hash:"-"` is used anymore but just to keep in consistent, let's pretend it is and in that case the version is required in the hashing

NicolasMahe:
Closing in favor of https://github.com/mesg-foundation/core/pull/906

PR #895, opened 2019-04-18
Refactor/merge listeners

krhubert:
Depend on #891 

Changes:
- create filters for execution and event listeners
- merge results and tasks listeners into execution listener

There is one breaking change:
Execution status counts from 1. Why? Because a filter needs a way to filter *any* status so this is why I have changed it. If you have other solutions let me know.


api/event_listener.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// EventFilter store fileds for matching events.
+type EventFilter struct {
+	Key string
+}
+
+// Match matches event.
+func (f *EventFilter) Match(e *event.Event) bool {
+	return f == nil || f.Key == "" || f.Key == e.Key
```
antho1404: what about filters with `*`?

api/event_listener.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// EventListener provides functionalities to listen MESG events.
+type EventListener struct {
+	C chan *event.Event
```
antho1404: if this is exposed it should have a better name, something like `eventC`

api/api.go
```diff
@@ -130,10 +131,108 @@ func (a *API) ExecuteTask(serviceID, taskKey string, inputData map[string]interf
...
 }
 
+// ListenEvent listens events matches with eventFilter on serviceID.
+func (a *API) ListenEvent(service string, f *EventFilter) (*EventListener, error) {
+	s, err := a.db.Get(service)
+	if err != nil {
+		return nil, err
+	}
+
+	if f != nil && f.Key != "" {
```
antho1404: what about `*`?

api/api.go
```diff
@@ -130,10 +131,108 @@ func (a *API) ExecuteTask(serviceID, taskKey string, inputData map[string]interf
...
+
+// ListenExecution listens executions on service.
+func (a *API) ListenExecution(service string, f *ExecutionFilter) (*ExecutionListener, error) {
+	s, err := a.db.Get(service)
+	if err != nil {
+		return nil, err
+	}
+
+	if f != nil {
+		if f.TaskKey == "" && f.OutputKey != "" {
```
antho1404: any reason to force this? For now we are able to listen to any `OutputKey`

api/execution_listener.go
```diff
@@ -0,0 +1,77 @@
...
+	OutputKey string
+	Tags      []string
+}
+
+// Match matches execution.
+func (f *ExecutionFilter) Match(e *execution.Execution) bool {
+	if f == nil {
+		return true
+	}
+	if f.TaskKey != "" && f.TaskKey != e.TaskKey {
```
antho1404: `*` support?

api/execution_listener.go
```diff
@@ -0,0 +1,77 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+// ExecutionListener provides functionalities to listen MESG tasks.
+type ExecutionListener struct {
+	// Channel receives matching executions for tasks.
+	C chan *execution.Execution
```
antho1404:
```suggestion
	ExecutionC chan *execution.Execution
```

api/event_listener.go
krhubert: I forget about them, adding. Nice catch

api/api.go
krhubert: is it make sense to listen for any `OutputKey` without `TaskKey`? What such listening would mean?

I can change it but it seemed to me rather like a bug not a feature ;)

api/event_listener.go
krhubert: C is fine. the struct name (`EventListener`) gives the information about what type of channel it is.
If you want to more clarity in the code you can use longer var names .
```
eventListener  := NewEventListener()
<-eventListenr.C
```

or shorter if it's clear from context.

Example with `C` naming https://golang.org/pkg/time/#Timer

api/api.go
antho1404: I used it in some cases when I know that the service always returns the output `error` then I can only listen to that `outputKey` and catch all errors

antho1404: This behavior is normal (I used it in an old version of the core). We can have outputFilter only if taskFilter is present, that makes more sense but we just need to update the documentation https://github.com/mesg-foundation/core/issues/922

Can you revert this modification and sorry for that

krhubert: done

interface/grpc/service/service.go
```diff
@@ -30,7 +30,7 @@ func (s *Server) EmitEvent(context context.Context, request *serviceapi.EmitEven
 
 // ListenTask creates a stream that will send data for every task to execute.
 func (s *Server) ListenTask(request *serviceapi.ListenTaskRequest, stream serviceapi.Service_ListenTaskServer) error {
-	ln, err := s.api.ListenTask(request.Token)
+	ln, err := s.api.ListenExecution(request.Token, nil)
```
NicolasMahe: Shouldn't it listen only for "Created/InProgress" execution?

api/api.go
NicolasMahe: When using command `service dev`, I get the error:
`⨯ Listening events error: rpc error: code = Unknown desc = Event "*" not found in service "Bitcoin"`

interface/grpc/service/service.go
NicolasMahe: I get the following error when running a service with `service dev` and calling multiple time `service execute`:
```
service | (node:1) UnhandledPromiseRejectionWarning: Error: 2 UNKNOWN: Execution status error: "completed" instead of "in progress"
```

to reproduce:

In one terminal:
```
mesg-core service dev https://github.com/mesg-foundation/service-js-function --event-filter=""
```

in an other:
```
mesg-core service execute js-function --task execute --data code="(a) => a + ' michel'" --data inputs='hello'
// no error
mesg-core service execute js-function --task execute --data code="(a) => a + ' michel'" --data inputs='hello'
// error in log of dev command
```

api/api.go
krhubert: done

api/api.go
```diff
@@ -130,10 +131,108 @@ func (a *API) ExecuteTask(serviceID, taskKey string, inputData map[string]interf
...
+			return nil, fmt.Errorf("execution filter: output key given without task key")
+		}
+		if f.HasTaskKey() {
+			task, err := s.GetTask(f.TaskKey)
+			if err != nil {
+				return nil, err
+			}
+			if _, err := task.GetOutput(f.OutputKey); f.OutputKey != "" && err != nil {
```
NicolasMahe: The output key should also accept `*`.
See previous implementation:
https://github.com/mesg-foundation/core/blob/d7500d9251bb5e562e9ac98d76607f558a5f100c/api/listen_result.go#L128-L138

I think you should add a function `HasTaskOutputKey` similar to `HasTaskKey`.

To reproduce:
```
mesg-core service dev https://github.com/mesg-foundation/service-ethereum-erc20 -t name -o=*
⨯ Listening results error: rpc error: code = Unknown desc = Output "*" of task "name" not found in service "Ethereum ERC20 token"
```

api/api.go
```diff
@@ -169,8 +169,10 @@ func (a *API) ListenExecution(service string, f *ExecutionFilter) (*ExecutionLis
 			if err != nil {
 				return nil, err
 			}
-			if _, err := task.GetOutput(f.OutputKey); f.OutputKey != "" && err != nil {
-				return nil, err
+			if f.HasOutputKey() {
+				if _, err := task.GetOutput(f.OutputKey); f.OutputKey != "" && err != nil {
```
NicolasMahe: `f.OutputKey != ""` Is this still necessary?

NicolasMahe:
@krhubert this PR cannot even build....

api/api.go
krhubert: What do you mean?

The output must exist in the filter in order to get output and check if exists. If the output is empty or `*` then it doesn't check. So basically this line is necessary or am I missing your question

NicolasMahe: The `f.OutputKey != ""` check is in function `HasOutputKey` so it doesn't need to be in the condition line 176.
But anyway, this is not blocking. So let's leave it.

krhubert: fixed, wating for tests to complete

api/event_listener.go
```diff
@@ -0,0 +1,60 @@
...
+		ps:     ps,
+		topic:  topic,
+		c:      ps.Sub(topic),
+		filter: f,
+	}
+}
+
+// Close stops listening for events.
+func (l *EventListener) Close() {
+	go func() {
```
ilgooz: it's better `Close()` to work in sync. I'll approve this PR but can you please open another pr to improve this on closers?

krhubert: Can you tell more what do you mean by "work in sync"? to what should be sync close with?

NicolasMahe: Issue created to continue discussion: https://github.com/mesg-foundation/core/issues/966

PR #896, opened 2019-04-18
Marketplace: load manifest on demand

NicolasMahe:
This pr change the way the manifest is downloaded.
Before, getting the version also download the manifest.
Now, the manifest needs to be explicitly downloaded.

Task `getService` still returns the manifest, but task `listService` doesn't

TODO:
- [x] improve mesg.yml when https://github.com/mesg-foundation/core/pull/863 is merged (can be done in another PR).


systemservices/marketplace/src/types/version.ts
```diff
@@ -1,9 +1,6 @@
-import { Manifest } from "./manifest";
-
 export interface Version {
   versionHash: string;
   manifest: string;
   manifestProtocol: string;
-  manifestData: Manifest|null;
```
ilgooz: how about changing this to `manifestData?: Manifest;` and after this change we can add the _manifestData_ directly into https://github.com/mesg-foundation/core/pull/896/files#diff-1497f2b1650c529c8f58159491cfe107R8 instead of recreating the object

NicolasMahe: I feel it's safer to have multiple type. Like this, `manifestData` doesn't exist in Version and the developer cannot confuse between a not loaded `manifestData` and an not existing one (when the file is no more accessible from IPFS).

NicolasMahe:
> need to update `mesg.yml` as well

Done. Please review

systemservices/marketplace/mesg.yml
```diff
@@ -36,41 +36,110 @@ tasks:
...
+                optional: true
+                object:
+                  version:
+                    type: String
+                  service: &manifestServiceData	
                     type: Object
-                    optional: true
                     object:
-                      version:
+                      hashVersion:
```
antho1404: this one was supposed to be in definition and needs to be removed anyway based on https://github.com/mesg-foundation/core/pull/906

PR #901, opened 2019-04-19
Marketplace improve gas system

NicolasMahe:
Estimate gas instead of the default gas env variable!


systemservices/marketplace/src/contracts/utils.ts
```diff
@@ -42,22 +42,26 @@ interface CreateTransaction {
...
-  chainID: chainID,
-  to: contract.options.address,
-  nonce: (await web3.eth.getTransactionCount(inputs.from)) + (shiftNonce || 0),
-  gas: inputs.gas || defaultGas,
-  gasPrice: inputs.gasPrice || defaultGasPrice,
-  value: '0',
-  data: data
-})
+) => {
+  const tx = {
```
antho1404: Not something blocking but this is quite messy, I think `estimateGas` might need the `nonce`, `gasPrice` and `gas` to run the simulation, if it doesn't need it it probably just ignore it so we could write something like that that feels easier to understand
```js
const tx = { chainId, to: ..., value: 0, data, nonce: ..., gasPrice: ..., gas: inputs.gas }
if (!tx.gas) {
  tx.gas = await web3.eth.estimateGas({ ...tx, from: inputs.from })
}
return tx
```

NicolasMahe: I updated the code to your suggestion.

PR #903, opened 2019-04-19
service: move some types to service.go

ilgooz:


ilgooz:
> Why do you want to move the type in service.go?

It is to make it easy the review the next PR here https://github.com/mesg-foundation/core/pull/904. Otherwise you'll see more code changes on #904 where some blocks of code is removed from some files and added to other ones. It's hard to see if there are any lines of code is actually changed or not during this kind code move.

PR #904, opened 2019-04-19
introduce Container Manager 

ilgooz:
depends on https://github.com/mesg-foundation/core/pull/903


api/api.go
```diff
@@ -9,22 +9,25 @@ import (
...
+	"github.com/mesg-foundation/core/service/manager"
 	uuid "github.com/satori/go.uuid"
 )
 
 // API exposes all functionalities of MESG core.
 type API struct {
 	ps *pubsub.PubSub
 
+	m         manager.Manager
+	container container.Container
```
antho1404: I think container should not be needed anymore in the api because the logic is delegated to the manager

service/namespace.go
```diff
@@ -12,37 +12,37 @@ const eventChannel string = "Event"
 const taskChannel string = "Task"
 const resultChannel string = "Result"
 
-// namespace returns the namespace of the service.
-func (service *Service) namespace() []string {
+// Namespace returns the namespace of the service.
+func (service *Service) Namespace() []string {
```
antho1404: I think namespace is mostly used for all docker related functions so this should be moved directly to the manager package instead of being exposed.

service/manager/manager.go
```diff
@@ -0,0 +1,24 @@
+package manager
+
+import (
+	"github.com/mesg-foundation/core/service"
+)
+
+// Manager is responsible for managing Docker Containers of MESG services.
+// it can be implemented for any container orchestration tool.
+type Manager interface {
```
antho1404: I feel `Manager` is too generic, what do you think about `Runner`

service/service.go
```diff
@@ -56,8 +57,42 @@ type Service struct {
 	DeployedAt time.Time `hash:"-"`
 }
 
-// getDependency returns dependency dependencyKey or a not found error.
-func (s *Service) getDependency(dependencyKey string) (*Dependency, error) {
+// StatusType of the service.
+type StatusType uint
+
+// Possible statuses for service.
+const (
```
antho1404: I don't see any reason to keep this in the service and not put it in the manager but i could be ok to keep it like that

ilgooz: why not, it's a generic status indicator for services. it's not specific to Docker Service. It'll be used by Kubernetes or in any other implementation of Manager.

api/api.go
ilgooz: it's needed for deploy

service/namespace.go
ilgooz: it can be used by Kubernetes or any other implementation of Manager to identify a mesg service.

service/manager/manager.go
ilgooz: I don't mind but maybe we can come up with a much better name than these too. let's think about this, we can always change later

service/namespace.go
antho1404: but they will have different restrictions, here for example we are hashing in sha1 to make sure that the name will not be too long for docker, maybe kubernetes doesn't have this limitation or don't have a namespace system at all

ilgooz: i mean, we'll still need to assign some unique names to services in any orchestration tool that's implemented. we can just have one common unique name creator in service pkg and use it everywhere like this.

we can also change this `Namespace()` func to commonly return a string instead and Manager implementations can use it the way how it fits for themselves. tho, i'm not pushing to have this func here. we can refactor this in future if needed or in an another PR. i'm all ok about it

api/api.go
antho1404: and don't you think some part of the deploy make sense to have in the manager with a `prepare/init/build/create` function and have the deploy like that 
```
deploy() {
  service = import.From(...)
  metaInfo = manager.Prepare(service)
  db.Save(metaInfo)
  db.Save(service)
}
```

ilgooz: you mean actually create the docker image inside an implementation of `manager.Manager`? what do you mean about `prepare/init/build/create`? i understand the `build` part, what do you mean by the other ones?

for the build, it cannot be inside of the implementations of `manager.Manager` type. because it's not specific the orchestration tool that used. it's a generic thing for all. but sure, it can be moved inside to the `manager` pkg as a standalone func but not a part of `manager.Manager` interface.

antho1404:
Maybe the manager package is not necessary and the logic can directly be moved into the api package , API already have the container

ilgooz:
> Maybe the manager package is not necessary and the logic can directly be moved into the api package , API already have the container

if api pkg would become more generic to expose more features that not directly to work that Manager do, in this case this separation is still really good to me. for example, api pkg can expose doc generation, service validation and some other stuff.

api pkg also does not contain logic related service, execution database and pubsub. so, moving Manager logic inside doesn't make much sense. api pkg is a high level layer that interacts with different high level packages in Core

NicolasMahe:
@ilgooz can you fix lint and merge conflict

api/api.go
NicolasMahe: What do you mean it's not specific to the orchestration tool? There are different way to create image for different orchestration/containerization tool.

`dockermanager` should implement its own build function.
But if we have multiple `manager` that actually use the same build function, let's say a `dockermanager` and a `k8manager`, then they should use a common function from another package, something like `containerd.build`.
Let's not force this 'factorization' at the higher level. It should be at the low level.

I think here, the goal is to reduce the dependency of the `API` package and delegate more to other package.

NicolasMahe: Let's continue this conversation outside this PR: https://github.com/mesg-foundation/core/issues/956

service/namespace.go
NicolasMahe: There is already the `service.Hash` that can be use to get a unique name for a service.
I agree with Anthony, we should move it in Manager.

NicolasMahe: Let's continue the conversation here: https://github.com/mesg-foundation/core/issues/957

service/service.go
NicolasMahe: let's continue this conversation outside the PR: https://github.com/mesg-foundation/core/issues/958

NicolasMahe:
I created a bunch of new issue to continue discussion and not block this PR!

PR #905, opened 2019-04-22
Add version to databases path

### LINKED ISSUE #882: update default database path on breaking changes
this is needed by the next release, because we're breaking it now since we don't nil check the `Configuration` in `Service` anymore. in the past versions it wasn't filled and now we always set it

krhubert:
close #882


ilgooz:
> * Download last version of mesg-core (v0.8.1)
> * Deploy a new service with mesg-core
> * List services with `./dev-cli service list`
> 
> Expected:
> empty list of services
> 
> Actual:
> 
> ```
> ➜  core git:(feature/db-versioning) ✗ ./dev-cli service list
> rpc error: code = Internal desc = grpc: failed to unmarshal the received message unexpected EOF
> ```

mine works

krhubert:
@antho1404 

This PR is only about adding path to core and mine is work as well

PR #926, opened 2019-04-30
Fix isAuthorized task by changing a test that check for empty hex

NicolasMahe:
The function `hexToString` needs a valid utf8 to not throw an error.
It was used to test if the `sidHash` was empty or not. The sidHash is not an utf8 but a keccak.
I replaced it by a conversion to number to make the same test.

### DIFF
--- a/systemservices/marketplace/src/tasks/isAuthorized.ts
+++ b/systemservices/marketplace/src/tasks/isAuthorized.ts
@@ -16,7 +16,7 @@ export default (
       versionHash = inputs.versionHash
       // get service from version hash
       const sidHash = await contract.methods.versionHashToService(hashToHex(versionHash)).call()
-      if (hexToString(sidHash) === "") {
+      if (Number(sidHash) === 0) {
         throw new Error('service with hash ' + versionHash + ' does not exist')
       }
       const service = await contract.methods.services(sidHash).call()


systemservices/marketplace/src/tasks/isAuthorized.ts
```diff
@@ -16,7 +16,7 @@ export default (
       versionHash = inputs.versionHash
       // get service from version hash
       const sidHash = await contract.methods.versionHashToService(hashToHex(versionHash)).call()
-      if (hexToString(sidHash) === "") {
+      if (Number(sidHash) === 0) {
```
krhubert: what type sidHash is? what about `Number` overflow? are you 100% sure that there is not a signel sidHash that give 0 during the cast?

NicolasMahe: I tried with multiple values and it worked correctly:
```
> Number('0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff') === 0 ? "nok" : "ok"
'ok'
> Number('0x0000000000000000000000000000000000000000000000000000000000000001') === 0 ? "nok" : "ok"
'ok'
> Number('0x164475976c7f1b9abeee539b6fc35097c31148be1957c0746f30069b4ad022a4') === 0 ? "nok" : "ok"
'ok'
> Number('0x0000000000000000000000000000000000000000000000000000000000000000') === 0 ? "nok" : "ok"
'nok'
```

The function `versionHashToService` returns `0x0000000000000000000000000000000000000000000000000000000000000000` when the versionHash is not found.

PR #928, opened 2019-05-01
Use volumesForm in definition file

krhubert:


antho1404:
We don't change the attribute in the mesg.yml. Services already depend on this attribute there is no strong reason to change it. This is a breaking change that is not necessary and not the time right now for that.

krhubert:
Everywhere we are using `volumesFrom` (grpc, json) so having `volumesfrom` in mesg.yaml is inconsistent so consider again renaming it, as now is the best time because we break a lot in mesg.yaml (like configuration) and we don't have to do breaking change again.

NicolasMahe:
> Everywhere we are using `volumesFrom` (grpc, json) so having `volumesfrom` in mesg.yaml is inconsistent so consider again renaming it, as now is the best time because we break a lot in mesg.yaml (like configuration) and we don't have to do breaking change again.

We didn't break `mesg.yml`, only the gRPC version and internal version of it. Right?
This is a valid modification that we be merged in another release.

antho1404:
Yes, mesg.yml is not broken for now. Also, I think there is a way to parse the yaml with one attribute or the other something like `volumefrom|volumeFrom`.

The breaking changes of the API are not as critical as the mesg.yml, API is mostly internal (or used with other tools we control such as the js library), We don't have any control on the mesg.yml so this is a critical breaking change and just for cosmetic...

NicolasMahe:
Let's include this change in the future break of the task's output

NicolasMahe:
TODO:
- [ ] rename volumesfrom in docs https://docs.mesg.com/guide/service/dockerize-the-service.html#add-a-configuration-and-dependencies

Issue #932, opened 2019-05-01
mesg.yml accepts image under configuration section


krhubert:
```
configuration:
  image: busybox
```

is valid mesg definition and it shoudn't

PR #934, opened 2019-05-01
Do not allow image under configuration in mesg.yml

krhubert:
for now is ok to set configuration.image in mesg.yaml and it shound't

fixes https://github.com/mesg-foundation/core/issues/932


NicolasMahe:
Please don't forget to ref the issue

PR #942, opened 2019-05-03
Make marketplace configurable by the core

antho1404:
We need a way to use the marketplace in dev (ropsten) and prod (on the mainnet)

@NicolasMahe please update `scripts/build-core.sh` when you have the contract on the mainnet

### DIFF
--- a/dev-core
+++ b/dev-core
@@ -7,6 +7,9 @@ MESG_LOG_LEVEL=${MESG_LOG_LEVEL:-debug}
 MESG_LOG_FORCECOLORS=${MESG_LOG_FORCECOLORS:-false}
 VERSION=local
 LDFLAGS="-X 'github.com/mesg-foundation/core/version.Version=$VERSION'"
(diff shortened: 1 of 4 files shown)


dev-core
```diff
@@ -7,6 +7,9 @@ MESG_LOG_LEVEL=${MESG_LOG_LEVEL:-debug}
 MESG_LOG_FORCECOLORS=${MESG_LOG_FORCECOLORS:-false}
 VERSION=local
 LDFLAGS="-X 'github.com/mesg-foundation/core/version.Version=$VERSION'"
+LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.EnvMarketplaceEndpoint=https://ropsten.infura.io/v3/6115ae2531f64c04a9a392cf500e5fbe'"
```
krhubert: We can set it directly in the mesg.yaml of the service, in that case everone can deploy and use it without dev-core script.

NicolasMahe: The problem is in the file `config/services.go`. It always replace the env.
It could create the env map dynamically only if the variable contains something...
It could be improved later on. Let's merge it this way for now.

NicolasMahe:
I added the mainnet data 👍

PR #948, opened 2019-05-09
Add sid to readme header

krhubert:
fix #946

PR #950, opened 2019-05-09
Fix/service start stop args

### LINKED ISSUE #915: service start & stop with no arguments
When running the command `service start` with no argument we have no error message, we should require at least one argument. Same for stop and possibly dev and logs

krhubert:
fix #915


NicolasMahe:
@krhubert please say that this PR is related to https://github.com/mesg-foundation/core/pull/948

krhubert:
@NicolasMahe Sorry it's not related I mess up commits

PR #955, opened 2019-05-13
service: fix deleting volumes for services never started before

### LINKED ISSUE #931: can't find volume after redeployed services
Steps to reproduce

```
mesg-core service deploy ./some-service-with-volume
mesg-core service delete  ./some-service-with-volume
mesg-core service deploy ./some-service-with-volume
mesg-core service delete  ./some-service-with-volume

rpc error: code = Unknown desc = Error: No such volume: 82cbdec06e95d7a028361c4acc1ab876f8f6a4a5
```

ilgooz:
fixes #931

### DIFF
--- a/service/volume.go
+++ b/service/volume.go
@@ -3,6 +3,7 @@ package service
 import (
 	"sync"
 
+	"github.com/docker/docker/client"
 	"github.com/mesg-foundation/core/container"
 	"github.com/mesg-foundation/core/x/xerrors"
 )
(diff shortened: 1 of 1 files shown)


service/volume.go
```diff
@@ -22,7 +23,9 @@ func (s *Service) DeleteVolumes(c container.Container) error {
 			wg.Add(1)
 			go func(source string) {
 				defer wg.Done()
-				if err := c.DeleteVolume(source); err != nil {
+				// if service is never started before, data volume won't be created and Docker Engine
+				// will return with the not found error. therefore, we can safely ignore it.
+				if err := c.DeleteVolume(source); err != nil && !client.IsErrNotFound(err) {
```
antho1404: could be done directly in the `DeleteVolume` function that ignores the error to avoid the dependency with docker but it's fine like that too

PR #963, opened 2019-05-17
First step of simplifaction of task output -  update importer package

krhubert:
I have removed tests that were using external services.


krhubert:
none of the system services will work in that case. @NicolasMahe requested to do PR following the order from the forum https://forum.mesg.com/t/simplification-of-the-tasks-output/265/6 and there update system services are one of the last tasks.

I'm not quite following how then I should divide this work. From one hand the post on the forum says about very small pull request and from the other hand everything should work. It's impossible to keep both the order of PR, small PR and working PR at the same time.

antho1404:
Please have a look at this PR https://github.com/mesg-foundation/core/pull/965, the core is not broken and it's just an update of the mesg.yml like you did for the tests so it's still the purpose of this PR.

Of course, this will still break the marketplace service but I've created a new version of the mesg-js lib https://github.com/mesg-foundation/mesg-js/tree/feature/output-simplification with the tag `next` so we could just update the marketplace package.json with that version and the pull request still stay small and the core is not broken and works perfectly because the system services are based on success/error outputs already.

The part of the forum to update the services is all the other services not system services. System services are part of the core and are necessary to run the core.

NicolasMahe:
@antho1404 Nice catch about the mesg-js that impacted the system service! I forgot about this one.

I tried and everything seems to work correctly 👍

NicolasMahe:
I changed to base branch of this PR to `feature/single-output` so we can still release the dev branch in case there is bug fix!

PR #964, opened 2019-05-17
2nd step of simplification of task output - change internal structure

krhubert:
depend on https://github.com/mesg-foundation/core/pull/963


antho1404:
some tests are failing because of the `InvalidTaskOutputError` type that has an empty `TaskOutputKey`

krhubert:
they are fixed in 3rd step, so if you can merge this one and then take a look on 3rd it would be great

NicolasMahe:
Test are fixed with commit https://github.com/mesg-foundation/engine/pull/971/commits/b5bbeaed9c4c16987515db436a5f9337dec9b479

@krhubert you should have pick this commit into this PR. It will be cleaner.

PR #965, opened 2019-05-18
Output simplification for system services

antho1404:
Remove the outputs from the definitions of the system service in order to have the core that can start.

I just removed all `error` outputs and returned the data of the previous `success` output


NicolasMahe:
@krhubert merge when you checked it

PR #967, opened 2019-05-19
Add back removed test from #963

NicolasMahe:
This PR add back fixed tests that has been removed from https://github.com/mesg-foundation/core/pull/963


NicolasMahe:
@krhubert merge when you checked it

PR #969, opened 2019-05-20
[mv Engine] Rename docker image tag to mesg/engine

NicolasMahe:


antho1404:
just make sure that we have the correct repository on docker hub

PR #972, opened 2019-05-20
4th step - do not use output filter in cli

krhubert:


NicolasMahe:
@krhubert please include the dependant PRs

PR #973, opened 2019-05-20
5th step - Remove output from service package

krhubert:


NicolasMahe:
@krhubert please include the dependant PRs

execution/execution.go
```diff
@@ -44,7 +44,6 @@ type Execution struct {
 	TaskKey           string                 `hash:"name:taskKey"`
 	Tags              []string               `hash:"name:tags"`
 	Inputs            map[string]interface{} `hash:"name:inputs"`
-	OutputKey         string                 `hash:"-"`
```
NicolasMahe: This modification should be done in another PR.
And also, don't forget to rename `OutputData` to `Outputs`: see "update execution db" in https://forum.mesg.com/t/simplification-of-the-tasks-output/265/6

api/api.go
```diff
@@ -176,8 +176,8 @@ func (a *API) ListenExecution(service string, f *ExecutionFilter) (*ExecutionLis
 }
 
 // SubmitResult submits results for executionID.
-func (a *API) SubmitResult(executionID string, outputKey string, outputs []byte) error {
```
antho1404: Maybe I'm missing some parts here but `SubmitResult` should have both the outputs and errors.
The execution already has an error field so why not getting it from `SubmitResultRequest_Error`. Here when there is an error we store it as valid data which should actually break the validation in the execution package `RequireInputs` and we ignore this data.

But maybe I'm missing something. I haven't tried it yet but I think there is an error with the validation of the data when we save the result. https://github.com/mesg-foundation/engine/blob/625840c3879812e62a732b38ab1945a6263efac1/execution/execution.go#L60

api/api.go
```diff
@@ -176,8 +176,8 @@ func (a *API) ListenExecution(service string, f *ExecutionFilter) (*ExecutionLis
...
-	exec, stateChanged, err := a.processExecution(executionID, outputKey, outputs)
+func (a *API) SubmitResult(executionID string, outputs []byte) error {
```
NicolasMahe: This function should also have an parameter error.
Otherwise how to know if the outputs is an error or not?

PR #974, opened 2019-05-22
Feature/single output

NicolasMahe:
Accumulator branch for other single-output-X branches.


antho1404:
I have a file missing error when I start the `dev-core` command
```
✔ Core stopped
✔ Core started
time="2019-05-28T07:38:32Z" level=fatal msg="file missing"
```

antho1404:
We need to deprecate the service database.
We can set a new `serviceDBVersion`

NicolasMahe:
The marketplace doesn't return error in this current version. This is fixed with https://github.com/mesg-foundation/engine/pull/987

PR #975, opened 2019-05-22
6th step - remove output key from execution

krhubert:


api/api.go
```diff
@@ -176,8 +176,8 @@ func (a *API) ListenExecution(service string, f *ExecutionFilter) (*ExecutionLis
 }
 
 // SubmitResult submits results for executionID.
-func (a *API) SubmitResult(executionID string, outputKey string, outputs []byte) error {
-	exec, stateChanged, err := a.processExecution(executionID, outputKey, outputs)
+func (a *API) SubmitResult(executionID string, outputs []byte) error {
```
NicolasMahe: Don't forget to add `error` next to `outputs` as parameter of `SubmitResult` so the execution can failed from an error from the service (and not only from an error from the Engine itself).

NicolasMahe: This is breaking service's task that send back a result with "error" outputKey:
```
➜  core git:(feature/single-output-pt-6) ./dev-cli service dev https://github.com/mesg-foundation/service-ethereum-erc20/tarball/single-outputs
```

```
➜  core git:(feature/single-output-pt-6) ./dev-cli service execute ethereum-erc20 --task name --data contractAddress='0x420167d87d35c3a249b32ef6225872fbd9ab85d' 
✔ Task "name" executed
⢿ Waiting for result... ^C
```

interface/grpc/service/service.go
```diff
@@ -69,5 +70,8 @@ func (s *Server) ListenTask(request *serviceapi.ListenTaskRequest, stream servic
 
 // SubmitResult submits results of an execution.
 func (s *Server) SubmitResult(context context.Context, request *serviceapi.SubmitResultRequest) (*serviceapi.SubmitResultReply, error) {
-	return &serviceapi.SubmitResultReply{}, s.api.SubmitResult(request.ExecutionID, request.OutputKey, []byte(request.OutputData))
+	if request.OutputKey == "error" {
+		return &serviceapi.SubmitResultReply{}, s.api.SubmitResult(request.ExecutionID, nil, errors.New(request.OutputData))
```
antho1404: should be `request.OutputData.message`, the error here will look like 
"{message: 'xxx'}"

krhubert:
There is the bug in core :
```
// ListenResult listens for results from a services.
func (s *Server) ListenResult(request *coreapi.ListenResultRequest, stream coreapi.Core_ListenResultServer) error {
  filter := &api.ExecutionFilter{
    Status:  execution.Completed,
    TaskKey: request.TaskFilter,
    Tags:    request.TagFilters,
  }
```

So the ListenResult won't listen the exeuction that failed, it's not related to this PR

antho1404:
Here the execution changes state so we should have it on the stream, this is a regression https://github.com/mesg-foundation/engine/blob/60540cfae7e316d917247b78363ed305ff1651c7/interface/grpc/core/core.go#L180
but ok this can be fixed in parallel but needs to be fixed

interface/grpc/service/service.go
krhubert: Not everyone supposes to put an error in the message. Here error needs to be just string. If you want to extend it somehow that's fine, but at the base is a string.

Why do we need to force message here? what's the point if we can have just plain text?

PR #977, opened 2019-05-22
8th step - Change submit results in protobuf

krhubert:


protobuf/serviceapi/api.proto
```diff
@@ -83,14 +83,16 @@ message TaskData {
...
-//   "outputKey":   "__OUTPUT_KEY__",
 //   "outputData":  "{\"foo\":\"super result\",\"bar\":true}"
+//   "error":  "{\"message\":\"some error\"}"
 // }
 // ```
 message SubmitResultRequest {
   string executionID = 1; // The `executionID` received from the [`ListenTask` stream](#listentask).
-  string outputKey = 2;   // The output's key as defined in the [service file](../guide/service/service-file.md).
-  string outputData = 3;  // The result's data encoded in JSON as defined in the [service file](../guide/service/service-file.md).
+  oneof result {
```
NicolasMahe:
```suggestion
  oneof outputs {
      string data = 3;  // The result's data encoded in JSON as defined in the [service file](../guide/service/service-file.md).
```

what about renaming result to outputs, and outputs to data.
So we can access it like: outputs.data and outputs.error.

NicolasMahe: Actually, in JS for example, the name of the oneof is not used. Meaning, we will access it just like:
`SubmitResultRequest.data`

So actually, I recommend to use explicit name all the time even if it a bit longer:
```
oneof outputs {
    string outputData = 3;
    string outputError = 4;
  }
```

PR #978, opened 2019-05-22
9th step -  Remove outputFilter and outputKey

krhubert:


commands/provider/wallet_provider.go
```diff
@@ -96,9 +96,9 @@ func (p *WalletProvider) call(task string, inputs interface{}, output interface{
 }
 
 func (p *WalletProvider) parseResult(r *coreapi.ResultData, output interface{}) error {
-	if r.OutputKey == "error" {
+	if r.Error != "error" {
```
NicolasMahe:
```suggestion
	if r.Error != "" {
```
missing modif here

NicolasMahe: @krhubert why this comment is solved?

krhubert: Because i forced pusehd when I changed it

krhubert: I will fix it in 11th step. so now we won't trigger again tests and builds for 2 branches

PR #983, opened 2019-05-23
Update ethwallet to single output

NicolasMahe:
Dependant on https://github.com/mesg-foundation/engine/pull/982

This PR updates the ethwallet service to use the "single-output" version of the service client.

Its go mod should be updated in the future because I set a specific commit hash instead of a nice tag.


systemservices/ethwallet/go.mod
```diff
@@ -3,15 +3,24 @@ module github.com/mesg-foundation/core/systemservices/ethwallet
...
+	github.com/asaskevich/govalidator v0.0.0-20180319081651-7d2e70ef918f // indirect
 	github.com/btcsuite/btcd v0.0.0-20190109040709-5bda5314ca95 // indirect
 	github.com/cespare/cp v1.1.1 // indirect
+	github.com/cpuguy83/go-md2man v1.0.8 // indirect
 	github.com/deckarep/golang-set v1.7.1 // indirect
 	github.com/ethereum/go-ethereum v1.8.20
 	github.com/go-stack/stack v1.8.0 // indirect
-	github.com/mesg-foundation/core v0.7.0
+	github.com/kr/pretty v0.1.0 // indirect
+	github.com/mesg-foundation/core v0.5.1-0.20190523130836-2ce46ed10def
```
antho1404: maybe that's nothing but why do we use v0.5.1 here when we were using v0.7.0 before, let's try to upgrade instead of downgrade

NicolasMahe: I force to use the repo core at commit #2ce46ed10def but it weirdly set the tag to 0.5.1.
It's working well so I guess it's just a "cosmetic" issue

PR #791, opened 2019-02-20
Remove CLI from Core

antho1404:
I'm working on another CLI separated from the core and in JS here https://github.com/mesg-foundation/cli.

So we take this decision, all the files in this PR are not necessary anymore

PR #985, opened 2019-05-23
remove cli

ilgooz:


antho1404:
we cannot run the core anymore in dev, dev-core is using dev-cli.
Have a look at this PR https://github.com/mesg-foundation/engine/pull/791
The PR is big but with clear commits, if it's clear enough we can work on that PR that fix that and other stuff.

ilgooz:
> we cannot run the core anymore in dev, dev-core is using dev-cli.
> Have a look at this PR #791
> The PR is big but with clear commits, if it's clear enough we can work on that PR that fix that and other stuff.

I was thinking about using the js client in dev-cli when we good to switch to js version. how about that?

antho1404:
We could use the cli but if there is no need for it I think we shouldn't and I would love to have it with just docker that anyone can run for example on a server and then just install the cli in their computer. I think we should start it directly with docker.

If you're ok yes switch to the other one, I think there is more work on the other one, so just review and I think it's ok to merge and any missing part should be other PR

PR #989, opened 2019-05-28
Update service db

antho1404:
Change the source of the service database to avoid issues when the core already contains services from a previous version


antho1404:
closing in favor of https://github.com/mesg-foundation/engine/pull/988

PR #994, opened 2019-05-29
Change Execution Attributes 

krhubert:
Changes mentioned in https://forum.mesg.com/t/execution-db-and-api/280

- rename id to hash
- remove exeuction duration
- replace service struct with hash
- support parent hash


execution/execution.go
```diff
@@ -37,41 +36,33 @@ func (s Status) String() (r string) {
 
 // Execution stores all informations about executions.
 type Execution struct {
-	ID                string                 `hash:"-"`
+	Hash              string                 `hash:"-"`
 	EventID           string                 `hash:"name:eventID"`
 	Status            Status                 `hash:"-"`
-	Service           *service.Service       `hash:"name:service"`
+	ServiceHash       string                 `hash:"name:serviceHash"`
+	ServiceParentHash string                 `hash:"name:serviceHash"`
```
antho1404: Why `serviceParentHash`?
The goal is to have the previous execution hash. The previous service hash will be fetched from the previous execution hash and anyway not really relevant.
Also, make sure that the hash tag is replaces

protobuf/serviceapi/api.proto
```diff
@@ -60,35 +60,35 @@ message ListenTaskRequest {
...
 
 // The request's data for the `SubmitResult` API.
-// The data must contain the `executionID` of the executed task received from the stream of [`ListenTask` API](#listentask).
+// The data must contain the `executionHash` of the executed task received from the stream of [`ListenTask` API](#listentask).
 //
 // **Example:**
 // ```json
 // {
-//   "executionID": "__EXECUTION_ID__",
+//   "executionHass": "__EXECUTION_HASH__",
```
NicolasMahe: typo here

antho1404:
The pull request is doing too many things:
- There are modifications on the volume key
- There are update on the proto API with some stuff that I'm not sure even compile 
- Also, the part to change the type in byte might be better in another PR.

interface/grpc/core/core.go
```diff
@@ -173,7 +173,7 @@ func (s *Server) ListenResult(request *coreapi.ListenResultRequest, stream corea
 				return err
 			}
 			if err := stream.Send(&coreapi.ResultData{
-				ExecutionID:   execution.ID,
+				ExecutionHash: string(execution.Hash),
```
NicolasMahe: Converting bytes to string will not work..
The hash is not directly convertible to string. You'll need to use base58 conversion functions.

protobuf/api/api.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+package api;
+option go_package = "github.com/mesg-foundation/core/protobuf/api";
+
+service Exeuction {
+  rpc (EmitEventRequest) returns (EmitEventReply) {}
```
NicolasMahe: what is this? RPC function without name?

protobuf/api/api.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+package api;
+option go_package = "github.com/mesg-foundation/core/protobuf/api";
+
+service Exeuction {
```
NicolasMahe: typo

service/manager/dockermanager/start.go
```diff
@@ -137,9 +138,9 @@ func extractVolumesFrom(s *service.Service, d *service.Dependency) ([]container.
 // volumeKey creates a key for service's volume based on the sid to make sure that the volume
 // will stay the same for different versions of the service.
 func volumeKey(s *service.Service, dependency string, volume string) string {
-	return xstructhash.Hash([]string{
+	return fmt.Sprintf("%x", xstructhash.Hash([]string{
```
NicolasMahe: why adding fmt.Sprintf? `xstructhash.Hash` is already returning an string

NicolasMahe:
> The pull request is doing too many things:
> 
> * There are modifications on the volume key
> * There are update on the proto API with some stuff that I'm not sure even compile
> * Also, the part to change the type in byte might be better in another PR.

I agree with @antho1404. This PR should be splitted into a few more specific ones.
Eg:
- Rename executionID to executionHash
- Change executionHash type from string to []byte
- Add parentHash
- etc...

execution/execution_test.go
```diff
@@ -5,213 +5,55 @@ import (
 	"testing"
 	"testing/quick"
 
-	"github.com/mesg-foundation/core/service"
-	"github.com/mesg-foundation/core/x/xstructhash"
 	"github.com/stretchr/testify/require"
 )
 
-var (
```
NicolasMahe: @krhubert could you move those test in sdk package in a new PR?
The code coverage of sdk is really low: https://codecov.io/gh/mesg-foundation/engine/compare/9cf2d0b48817a45110ca0189e74419b232f137d0...f20bddbc143aa30752d4d07be3be95b76060e68a/src/sdk/sdk.go

PR #997, opened 2019-05-29
Rename api to sdk

NicolasMahe:
Rename package api to sdk.


sdk/sdk.go
```diff
@@ -37,22 +37,22 @@ func New(m manager.Manager, c container.Container, db database.ServiceDB, execDB
 }
 
 // GetService returns service serviceID.
-func (a *API) GetService(serviceID string) (*service.Service, error) {
+func (a *SDK) GetService(serviceID string) (*service.Service, error) {
```
antho1404: should we rename `a` in `sdk`

`func (sdk *SDK)`

PR #999, opened 2019-05-29
remove useless tar file

NicolasMahe:
I removed what appear a random tar file in the service-test folder that is not used


antho1404:
I think we had a test about the hash calculation based on this tarball but the test is probably gone

PR #1006, opened 2019-05-31
Create execution api.

krhubert:
I don't like name `Exec` for service and also i've changed inputs and outputs to slice of byte as you cannot get `interface{}` from protobuf. Also in case of `bytes` type we can switch to a different format of input/output (for example msgpack).


antho1404:
I like the idea to keep bytes for inputs/outputs especially if we can do optimizations after with msgpack or other 👍 (if I remember cosmos is using something on top of protobuf for untyped data)

I also don't like the name `Exec`, it should be `Execution`.

Not really sure why you choose bit flags for the status, we will never have two status at the same time so if this is the reason it's not necessary and just a simple number could be enough.

Other than that it's great

protobuf/definition/execution.proto
```diff
@@ -0,0 +1,27 @@
...
+package definition;
+option go_package = "github.com/mesg-foundation/core/protobuf/definition";
+
+// Statys represents a status of execution.
+enum Status {
+  Unknown = 0;
+  Created = 0x01;
+  InProgress = 0x02;
+  Completed = 0x04;
+  Failed = 0x08;
```
NicolasMahe: why hexa?

krhubert: It's common for flags that you can perform `or`|`and` op. It's easier to see when you have more flags (0x01..0x10) rather then (1..16) but 1, 2, 4, 8, 16 are perfectly fine.

krhubert:
> Not really sure why you choose bit flags for the status, we will never have two status at the same time so if this is the reason it's not necessary and just a simple number could be enough.

I've should done this when implementing a filter status on execution package. The key point here is to allow filter set a bitmask for filtering statuses (in that case one integer instead of slice integer will be enough) and could be hidden on implementation level. It's better than sending a slice of statuses.

> I also don't like the name Exec, it should be Execution.

I don't like `Execution` ether, as it's already taken by `Executon` :) I found that some projects (like kubernetes) are adding suffix `Service` simply, so I propose `ExecutionService`.

protobuf/definition/execution.proto
antho1404: bitmask makes total sense for the filter but doesn't make much sense in the definition itself.
It's kind of confusing if we have it in the execution, it means that we could have 2 or more status at the same time. I'm ok to keep it but at least make it clear with comments

protobuf/execapi/api.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+import "protobuf/definition/execution.proto";
+
+package api;
+option go_package = "github.com/mesg-foundation/core/protobuf/execapi";
+
+service ExecutionService {
```
antho1404: I'm ok with that if we had only the ExecutionService but we will have the service so end up with:
```
service ServiceService {
```
I'm ok to keep just `Execution` here and anyway it will be in a different package so `execapi.Execution` and `definition.Execution`

protobuf/execapi/api.proto
```diff
@@ -0,0 +1,36 @@
+syntax = "proto3";
+
+import "protobuf/definition/execution.proto";
+
+package api;
+option go_package = "github.com/mesg-foundation/core/protobuf/execapi";
```
NicolasMahe: Could you move this file to `protobuf/api/execution.proto` like anthony suggest in https://forum.mesg.com/t/execution-db-and-api/280

protobuf/definition/execution.proto
```diff
@@ -0,0 +1,57 @@
...
+  string taskKey = 6;
+
+  // inputs is an inptut of execution.
+  bytes inputs = 7;
+
+  // outputs is an output of successful execution.
+  bytes outputs = 8;
+
+  // error is an output of a failed execution.
+  bytes error = 9;
```
NicolasMahe: why bytes for error? It should be a string as we don't have any define structure of what an error is.

krhubert: bytes is for outputs (which is output for success) so it makes sens to make error as bytes (which is output for failed)

protobuf/api/api.proto
```diff
@@ -0,0 +1,36 @@
...
+  // Get returns a single Execution specified in a request.
+  rpc Get(GetRequest) returns (definition.Execution) {}
+
+  // Stream returns a stream of exeuction that satisfy createria
+  // specified in a request.
+  rpc Stream(StreamRequest) returns (stream definition.Execution) {}
+}
+
+// GetRequest defines request to retrive a single execution.
+message GetRequest {
```
NicolasMahe: Need to add prefix or suffix `Execution` because its in package API

krhubert: Why do you want prefix or suffix here?

The standard practice is to use method name and Request suffix. If you want to add Execution to the name, then you have too add it to method too, so it will become `rpc GetExecution`.

NicolasMahe: The message are not namespaced by the service in protobuf..

The rpc method will be `api.Execution.Get` but the message will be `api.GetRequest`.
It will create conflict with other rpc method on other service. eg `api.Service.Get` with message `api.GetRequest`

protobuf/definition/execution.proto
NicolasMahe: error is not related to outputs anymore. It related to the state failed.
Error should be a string otherwise it's prone to JSON encode, or protobuf encode, or even dump the GO error struct into this field and NO client will be able to decode it.

protobuf/definition/execution.proto
NicolasMahe: @krhubert have you tried to use it with a client?
I'm pretty sure gRPC will not accept enum mask in a parameter of type `definition.Status`.
See line 31 of file protobuf/api/api.proto
https://github.com/mesg-foundation/engine/pull/1006/files#diff-759dbd0083a04481d938c7f1f0b9e403R31

protobuf/execapi/api.proto
NicolasMahe: The file is still name api.proto. it should be execution.proto

protobuf/api/api.proto
NicolasMahe: Conversation moved to https://github.com/mesg-foundation/engine/issues/1013

protobuf/execapi/api.proto
NicolasMahe: Conversation moved to https://github.com/mesg-foundation/engine/issues/1013

PR #1007, opened 2019-05-31
Implement execution service

krhubert:
dep #1006 

~~This is one intentional bug here.~~

~~there is one to one mapping of status in protobuf and execution.~~

~~To fix this I want to change execution status to be `or`ed and also replace slice of status in execution filter to just status int.~~

EDIT:

status value was change to one coresponding in execution package.


protobuf/definition/execution.proto
```diff
@@ -0,0 +1,57 @@
...
+  string taskKey = 6;
+
+  // inputs is an inptut of execution.
+  bytes inputs = 7;
+
+  // outputs is an output of successful execution.
+  bytes outputs = 8;
+
+  // error is an output of a failed execution.
+  bytes error = 9;
```
NicolasMahe: why bytes for error? It should be a string as we don't have any define structure of what an error is.

protobuf/execapi/api.proto
```diff
@@ -0,0 +1,36 @@
+syntax = "proto3";
+
+import "protobuf/definition/execution.proto";
+
+package api;
+option go_package = "github.com/mesg-foundation/core/protobuf/execapi";
```
NicolasMahe: Could you move this file to `protobuf/api/execution.proto` like anthony suggest in https://forum.mesg.com/t/execution-db-and-api/280

NicolasMahe:
@krhubert can you add the dependency to other PR
My comment are actually related to another PR.

protobuf/execapi/api.proto
NicolasMahe: https://github.com/mesg-foundation/engine/pull/1006/files#r289638747

protobuf/definition/execution.proto
NicolasMahe: Moved to https://github.com/mesg-foundation/engine/pull/1006/files#r289638767

NicolasMahe:
Could the new functions in sdk be moved to sdk/execution sub-package as describe in https://forum.mesg.com/t/execution-db-and-api/280 and https://forum.mesg.com/t/sdk-package-organisation/299 ?

NicolasMahe:
@krhubert todo related to these issue: https://github.com/mesg-foundation/engine/issues/1013

PR #1014, opened 2019-06-03
implement new deployment style

ilgooz:


sdk/service/service.go
```diff
@@ -0,0 +1,89 @@
...
+		container: c,
+		db:        db,
+		execDB:    execDB,
+	}
+}
+
+// Create creates a new service from definition.
+func (s *ServiceSDK) Create(srv *service.Service) error {
+	// download and untar service context into path.
+	path, err := ioutil.TempDir("", "mesg-"+uuid.NewV4().String())
```
NicolasMahe: `ioutil.TempDir` already create an unique folder. `uuid.NewV4().String()` is not necessary.

sdk/service/service.go
```diff
@@ -0,0 +1,89 @@
...
+		return err
+	}
+	defer os.RemoveAll(path)
+
+	resp, err := http.Get("http://ipfs.app.mesg.com:8080/ipfs/" + srv.Source)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != 200 {
+		return errors.New("service's ource code is not reachable")
```
NicolasMahe:
```suggestion
		return errors.New("service's source code is not reachable")
```

server/grpc/newcore/core.go
```diff
@@ -0,0 +1,27 @@
+package newcore
```
NicolasMahe: Can you move this file to `server/grpc/api/server.go` and the create function in `server/grpc/api/service.go` and renamed it `CreateService`

server/grpc/newcore/proto.go
```diff
@@ -0,0 +1,100 @@
+package newcore
```
NicolasMahe: Can you move the content of this file in `server/grpc/api/service.go`

protobuf/service/service.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
```
antho1404: maybe we can move this file in `protobuf/api/service.proto` (https://forum.mesg.com/t/protobuf-folder-structure/297)

sdk/service/service.go
antho1404:
```suggestion
		return errors.New("service's source code is not reachable")
```

sdk/service/service.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/mesg-foundation/core/database"
+	"github.com/mesg-foundation/core/service"
+	"github.com/mesg-foundation/core/service/manager"
+	"github.com/mesg-foundation/core/utils/dirhash"
+	"github.com/mr-tron/base58"
+	uuid "github.com/satori/go.uuid"
+)
+
+// ServiceSDK exposes service APIs of MESG.
+type ServiceSDK struct {
```
NicolasMahe: What about just naming it `Service`? the filepath already give the context of sdk.

ilgooz: or `SDK`?

NicolasMahe: No, it's only about the service

protobuf/definition/service.proto
```diff
@@ -14,6 +14,7 @@ message Service {
   repeated Event events = 6;                // The list of events this service can emit.
   repeated Dependency dependencies = 7;     // The container dependencies this service requires.
   string repository = 9;                    // Service's repository url.
+  string source = 13;                       // The hash id of service's source code on IPFS.
```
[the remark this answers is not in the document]
NicolasMahe: After confirmation with @antho1404, we will go with only 1 field containing directly the IPFS hash 👍  closing this comment.

protobuf/service/service.proto
ilgooz: done

sdk/service/service.go
ilgooz: done

server/grpc/newcore/proto.go
ilgooz: done

server/grpc/newcore/core.go
ilgooz: done

sdk/service/service.go
ilgooz: done

protobuf/service/service.proto
```diff
@@ -0,0 +1,19 @@
...
+import "protobuf/definition/service.proto";
+
+package service;
+option go_package = "github.com/mesg-foundation/core/protobuf/service";
+
+service Service {
+  rpc Create (CreateRequest) returns (CreateResponse) {}
+}
+
+message CreateRequest {
```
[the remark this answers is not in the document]
ilgooz: done

protobuf/service/service.proto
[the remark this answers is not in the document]
ilgooz: done

sdk/service/service.go
ilgooz: done

ilgooz: done

core/main.go
```diff
@@ -23,6 +24,7 @@ type dependencies struct {
 	executionDB database.ExecutionDB
 	container   container.Container
 	sdk         *sdk.SDK
+	serviceSDK  *servicesdk.ServiceSDK
```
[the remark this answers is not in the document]
ilgooz: done

protobuf/api/service.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+import "protobuf/definition/service.proto";
+
+package api;
+
+service ServiceX {
```
ilgooz: I put X at the end for now otherwise names conflicts for _api.Service_

sdk/sdk.go
```diff
@@ -17,6 +18,8 @@ import (
 
 // SDK exposes all functionalities of MESG core.
 type SDK struct {
+	ServiceSDK *servicesdk.Service
```
antho1404: rename in `Service` we are already in the `SDK` so we can access with `sdk.Service`

server/grpc/api/service.go
```diff
@@ -1,53 +1,53 @@
-package newcore
+package api
 
 import (
 	"github.com/mesg-foundation/core/protobuf/definition"
 	"github.com/mesg-foundation/core/service"
 )
 
-func fromProtoService(s *definition.Service) *service.Service {
+func FromProtoService(s *definition.Service) *service.Service {
```
antho1404: let's expose only this one, `fromProtoTasks`, `fromProtoEvents`... don't need to be public, they are internal functions only

server/grpc/server.go
```diff
@@ -107,12 +103,12 @@ func (s *Server) Close() {
 // register all server
 func (s *Server) register() error {
 	coreServer := core.NewServer(s.sdk)
-	newCoreServer := newcore.NewServer(s.serviceSDK)
+	coreServiceServer := NewServiceServer(s.sdk)
```
antho1404: could pass only the `s.sdk.Service` to reduce dependencies but it's fine ;)

protobuf/api/service.proto
antho1404: hum that's annoying. Ultimately the service api (with emitEvent, listenTask) will be removed in favor of the `execution/stream` api and a `core/emit` api. So I'm ok to keep it like that and rename it when we will be at this step.
Maybe if you can just add a TOFIX with a comment to make sure we don't forget about it

server/grpc/server.go
ilgooz: 👍 on forum proposal it says that sub sdk packages can use each other so this why chose to use sdk

sdk/sdk.go
ilgooz: done

server/grpc/api/service.go
ilgooz: done

protobuf/api/service.proto
ilgooz: done

sdk/service/service.go
```diff
@@ -0,0 +1,88 @@
...
+		return err
+	}
+
+	// calculate and apply hash to service.
+	dh := dirhash.New(path)
+	h, err := dh.Sum(nil)
+	if err != nil {
+		return err
+	}
+	srv.Hash = base58.Encode(h)
```
NicolasMahe: Careful here, the service hash should be calculated from the source (already done here) but also from the data of the structure service itself.

service/service.go
```diff
@@ -53,6 +53,9 @@ type Service struct {
 	// a Git host.
 	Repository string `hash:"name:7"`
 
+	// Source is the hash id of service's source code on IPFS.
+	Source string
```
```suggestion
	Source string `hash:"-"`
```
Source should be omitted from hash calculation until the cli can generate fully deterministic archive and get the same ifps hash all the time

sdk/service/service.go
NicolasMahe: I updated the post https://forum.mesg.com/t/service-compilation-deploy/288 as it didn't explicitly say that the hash calculation changed.

service/service.go
ilgooz: done

sdk/service/service.go
ilgooz: done

ilgooz:
@NicolasMahe let's merge this one, i'll send some fixes in a separate PR.

PR #1024, opened 2019-06-04
implement new service start style

ilgooz:
while using new instance db.

depends on https://github.com/mesg-foundation/engine/pull/1023.

@mesg-foundation/core please keep in mind that, since `listenTask()` only accepts service hashes, an instance created by `instancesdk.Create()` cannot start listening for tasks yet. I can make `listenTask()` to both accept service hash and instance hash if necessary but in another PR. Please request this feature if you think it's needed right now, otherwise i'll eventually do it after we merge everything related to new deployment/start style.


service/manager/manager.go
```diff
@@ -6,9 +6,11 @@ import (
 
 // Manager is responsible for managing Docker Containers of MESG services.
 // it can be implemented for any container orchestration tool.
+// TODO(ilgooz): discuss if these should accept service.Instance instead of service.Service.
```
ilgooz: @mesg-foundation/core please provide feedbacks about this.

sdk/instance/instance.go
```diff
@@ -0,0 +1,113 @@
...
+
+	srv.Hash = instanceHash
+	srv.Configuration.Image = imageHash
+	srv.Configuration.Env = xos.EnvMapToSlice(instanceEnv)
+	serviceIDs, networkID, err := i.m.Start(srv)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(ilgooz) this should be set by Manager.Start(), by the underlying
```
ilgooz: @mesg-foundation/core please provide feedbacks about this.

sdk/instance/instance.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	// save & start instance.
+	o := &service.Instance{
+		Sid:         srv.Sid,
+		Hash:        instanceHash,
+		ServiceHash: srv.Hash,
+		ImageHash:   imageHash,
+	}
+	if err := i.instanceDB.Save(o); err != nil {
```
ilgooz: it's better to also check for internal errors and return with after this nil err check. can be done in another PR too.

sdk/instance/instance.go
```diff
@@ -0,0 +1,116 @@
...
+
+	srv.Hash = instanceHash
+	srv.Configuration.Image = imageHash
+	srv.Configuration.Env = xos.EnvMapToSlice(instanceEnv)
+	serviceIDs, networkID, err := i.m.Start(srv)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO(ilgooz) this should be set by Manager.Start(), by the underlying
```
ilgooz: @mesg-foundation/core please provide feedbacks about this.

service/manager/manager.go
ilgooz: @mesg-foundation/core please provide feedbacks about this.

sdk/instance/instance.go
```diff
@@ -0,0 +1,117 @@
...
```
ilgooz: @mesg-foundation/core please provide feedbacks about this.

service/manager/manager.go
ilgooz: @mesg-foundation/core please provide feedbacks about this.

sdk/instance/instance.go
ilgooz: @mesg-foundation/core please provide feedbacks about this.

service/manager/manager.go
ilgooz: @mesg-foundation/core please provide feedbacks about this.

service/service.go
```diff
@@ -60,6 +60,16 @@ type Service struct {
 	DeployedAt time.Time `hash:"-"`
 }
 
+// Instance represents a running instance of service.
+type Instance struct {
+	Sid         string
```
antho1404: What's the purpose of this sid? is it to be able to run an instance based on the sid? If this is the case it should be removed and the api/sdk should do the resolution and Instance only have a service hash

service/manager/manager.go
antho1404: Yes it should accept instance and I think this function can be moved directly to the sdk under the instance sdk

sdk/instance/instance.go
antho1404: I would move the logic from the manager directly in this package.

service/service.go
ilgooz: It's for extra doc that we can send to cli on an instance create or delete action. I can move this and fetch it from service db. let's do it an another PR

service/manager/manager.go
ilgooz: I'll do this improvement in an another PR

ilgooz:
> from @antho1404 
I would move the logic from the manager directly in this package.

Any reason why? To me this should be handled by implementations of Manager. That way, the chosen Manager implementation can provide its own info easily. Otherwise we have to switch check to determine underlying orchestration tool used and maybe create different type of info for each (if in sdk/instance).

sdk/instance/instance.go
```diff
@@ -0,0 +1,117 @@
...
+	return &Instance{
+		m:          m,
+		container:  c,
+		db:         db,
+		instanceDB: instanceDB,
+	}
+}
+
+// Create creates a new service instance for service with id(sid/hash) and applies given env vars.
+func (i *Instance) Create(id string, env []string) (*service.Instance, error) {
```
```suggestion
func (i *Instance) Create(serviceHash string, env []string) (*service.Instance, error) {
```
Not blocking but I would rename the id variable to something way more explicit.

service/manager/manager.go
NicolasMahe: Yes it should accept instance instead of service.
Agree also to move the manager package directly inside sdk/instance.

service/service.go
NicolasMahe: Yes please remove Sid.

sdk/instance/instance.go
NicolasMahe: I agree with @antho1404. Moving the manager inside this package will simplify a lot.

sdk/instance/instance.go
```diff
@@ -0,0 +1,117 @@
...
+		ServiceHash: srv.Hash,
+		ImageHash:   imageHash,
+	}
+	if err := i.instanceDB.Save(o); err != nil {
+		return nil, err
+	}
+
+	srv.Hash = instanceHash
+	srv.Configuration.Image = imageHash
+	srv.Configuration.Env = xos.EnvMapToSlice(instanceEnv)
```
NicolasMahe: The above 3 lines have to me removed.
The service should not be modified when an instance is created.
The goal is actually to be able to create multiple instance for one service.
The instance have to directly contains those 3 info: instanceHash, imageHash and merge envs

NicolasMahe: The instanceHash and imageHash are alread saved in the instance struct.

The env could also be saved in the instance struct OR directly send to the function start of the container. Up to you.

service/service.go
NicolasMahe: Also move the struct instance in its own package `instance`.

antho1404:
> Any reason why? To me this should be handled by implementations of Manager. That way, the chosen Manager implementation can provide its own info easily. Otherwise we have to switch check to determine underlying orchestration tool used and maybe create different type of info for each (if in sdk/instance).

The orchestration tool used should be managed by the container package directly and the manager should anyway not be aware of that. 
The manager is actually doing some actions (kind of a controller) so it makes sense to me to put it in the package for all the actions (sdk).
Also, this manager does not manage services anymore but instances so it doesn't make sense to have it in the service package anyway.

Also, lint and test are failing

sdk/instance/instance.go
ilgooz: this will be done later in another pr.

sdk/instance/instance.go
ilgooz: Discussed with @antho1404, I'll change the name but still resolve the sid inside to service hash when sid is given in the argument.

antho1404:
This PR has been split into 
- https://github.com/mesg-foundation/engine/pull/1033
- https://github.com/mesg-foundation/engine/pull/1034
- https://github.com/mesg-foundation/engine/pull/1035
- https://github.com/mesg-foundation/engine/pull/1036

PR #1025, opened 2019-06-05
implement new service stop style

ilgooz:
while using new instance db.

depends on https://github.com/mesg-foundation/engine/pull/1024


sdk/instance/instance.go
```diff
@@ -115,3 +115,25 @@ func (i *Instance) Create(id string, env []string) (*service.Instance, error) {
...
+	}
+
+	// get instance's service from db.
+	srv, err := i.db.Get(o.ServiceHash)
+	if err != nil {
+		return nil, err
+	}
+
+	// stop the instance and delete it
+	srv.Hash = o.Hash
```
NicolasMahe: why changing the hash??

sdk/instance/instance.go
```diff
@@ -115,3 +115,25 @@ func (i *Instance) Create(id string, env []string) (*service.Instance, error) {
...
+// Delete deletes a running instance by instance hash.
+func (i *Instance) Delete(hash string) (*service.Instance, error) {
+	// get the instance from db.
+	o, err := i.instanceDB.Get(hash)
+	if err != nil {
+		return nil, err
+	}
+
+	// get instance's service from db.
+	srv, err := i.db.Get(o.ServiceHash)
```
NicolasMahe: why it's getting the service??
It should only stop the instance and delete it.

NicolasMahe:
should be close because of https://github.com/mesg-foundation/engine/pull/1035

PR #1026, opened 2019-06-05
sdk/service:delete: require service's instance been stopped before delete

ilgooz:
depends on https://github.com/mesg-foundation/engine/pull/1025.


database/instance_db.go
```diff
@@ -72,6 +75,28 @@ func (d *LevelDBInstanceDB) Get(hash string) (*service.Instance, error) {
 	return i, tx.Commit()
 }
 
+// GetByService retrives instance by its service's hash.
+func (d *LevelDBInstanceDB) GetByService(hash string) (*service.Instance, error) {
```
NicolasMahe: The relation between instance and service is many to one. many instance to one service.
This function should return an array of instance.

sdk/sdk.go
```diff
@@ -85,6 +86,16 @@ func (sdk *SDK) DeleteService(serviceID string, deleteData bool) error {
 	if err != nil {
 		return err
 	}
+
+	// check if service hash a running instance.
+	_, err = sdk.Instance.GetByService(s.Hash)
+	if err == nil {
+		return errors.New("stop service's instance before deleting")
```
NicolasMahe: Error should contains the instances' hash. Otherwise it will be very complex for the user to stop the instances.

database/instance_db.go
antho1404: to clarify: one service has many instances and one instance belongs to one service

NicolasMahe:
Should closed because of https://github.com/mesg-foundation/engine/pull/1037

PR #1032, opened 2019-06-06
Simplify configs package

NicolasMahe:
Closes https://github.com/mesg-foundation/engine/issues/1021

This PR simplify the config package by removing config that were used by the cli, flatten the Core struct, and make the Engine work with default value!


config/config.go
```diff
@@ -43,47 +38,25 @@ type Config struct {
...
 	c.Log.ForceColors = false
-	c.Core.Image = "mesg/engine:" + strings.Split(version.Version, " ")[0]
-	c.Core.Name = "engine"
-	c.Core.Path = filepath.Join(home, ".mesg")
-	c.Core.Database.ServiceRelativePath = filepath.Join("database", "services", serviceDBVersion)
-	c.Core.Database.ExecutionRelativePath = filepath.Join("database", "executions", executionDBVersion)
-	c.Docker.Core.Path = "/mesg"
-	c.Docker.Socket = "/var/run/docker.sock"
+	c.Name = "engine"
+	c.Path = filepath.Join("mesg")
```
krhubert:
```suggestion
	c.Path = "mesg"
```

or home is missing? also if mesg will be created in engine package, it is worth to add `mesg` to .gitignore

core/main.go
```diff
@@ -33,13 +33,13 @@ func initDependencies() (*dependencies, error) {
 	}
 
 	// init services db.
-	serviceDB, err := database.NewServiceDB(filepath.Join(config.Core.Path, config.Core.Database.ServiceRelativePath))
+	serviceDB, err := database.NewServiceDB(filepath.Join(config.Path, config.Database.ServiceRelativePath))
```
krhubert: config.Database.ServiceRelativePath should already contain config.Path

config/config.go
NicolasMahe: It was a mistake, I put back path to `/path`.
This path is only used inside the docker container.

core/main.go
NicolasMahe: At the end it makes sense to keep `config.Path` out of `config.Database.XXXXRelativePath` so by only modifiying `config.Path`, it will change all other path that use this folder.
The user doesn't need to know all sub-folder.

krhubert: What I want to avoid is the fact that other package needs to modify configs path. In that case, only config package manages paths etc... What you wrote is true also when config package joins paths.

To be clear I don't want a user to specify the full path for the database. I want move join into config package.

NicolasMahe:
I moved `config.Path` to `~/.mesg` so it's compatible within docker but also without docker! Really important in order to run test properly without having any hack.
The mesg folder should be mount inside docker in `/home/root/.mesg`. I will update JS CLI.

PR #1033, opened 2019-06-07
Create instance db

antho1404:
Add instance struct and instance DB


config/config.go
```diff
@@ -20,6 +20,7 @@ const (
 
 	serviceDBVersion   = "v2"
 	executionDBVersion = "v1"
+	instanceDBVersion  = "v1'"
```
NicolasMahe:
```suggestion
	instanceDBVersion  = "v1"
```

NicolasMahe: why is it resolved but not fixed?

database/instance_db.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// encode service
+	b, err := d.marshal(i)
+	if err != nil {
+		tx.Discard()
+		return err
+	}
+
+	// save instance with hash.
+	if err := tx.Put([]byte(i.Hash), b, nil); err != nil {
```
antho1404: @krhubert https://github.com/mesg-foundation/engine/pull/1037#discussion_r291853397
> the key can be saved with a single call without using transaction

database/instance_db.go
```diff
@@ -0,0 +1,104 @@
...
+	return &s, nil
+}
+
+// Get retrives instance by instance hash.
+func (d *LevelDBInstanceDB) Get(hash string) (*instance.Instance, error) {
+	tx, err := d.db.OpenTransaction()
+	if err != nil {
+		return nil, err
+	}
+	b, err := tx.Get([]byte(hash), nil)
```
antho1404: @krhubert https://github.com/mesg-foundation/engine/pull/1037#discussion_r291853412
> the key can be get with a single call without using transaction

database/instance_db.go
```diff
@@ -0,0 +1,104 @@
+package database
```
antho1404: @krhubert https://github.com/mesg-foundation/engine/pull/1037#discussion_r291853650
> There are no tests for this package

instance/instance.go
```diff
@@ -0,0 +1,9 @@
+package instance
+
+// Instance is an instance of a service.
+// This contains a reference to a service that is running.
+// Multiple instances can run for the same service as long as they have different configurations
+type Instance struct {
+	Hash        string `hash:"-"`
```
antho1404: @krhubert https://github.com/mesg-foundation/engine/pull/1037#discussion_r291853448
> instance tags are not needed here (structhash is not used on the struct)

ilgooz:
are you sure that writes are atomic without using transactions? it doesn't worth to rm transactions now since db is going to changed soon, there is no reason to introduce any instability in this period.

ilgooz:
@antho1404 this is blocking, can you give the last decision about transactions please?

antho1404:
I don't see the issue with the transaction, this is already used in the execution db so I don't see why there is an issue with that

PR #1034, opened 2019-06-07
Add sdk function to start an instance

antho1404:
dependency: https://github.com/mesg-foundation/engine/pull/1033

Start an instance based on the previous manager. Logic moved from service/manager to sdk/instance


ilgooz:
I get the following error but it starts the instance

```
Start service... Error: 2 UNKNOWN: Error response from daemon: rpc error: code = InvalidArgument desc = name must be valid as a DNS name component
Error: 2 UNKNOWN: Error response from daemon: rpc error: code = InvalidArgument desc = name must be valid as a DNS name component
    at Object.exports.createStatusError (/home/hey/data/src/github.com/mesg-foundation/cli/node_modules/grpc/src/common.js:91:15)
    at Object.onReceiveStatus (/home/hey/data/src/github.com/mesg-foundation/cli/node_modules/grpc/src/client_interceptors.js:1204:28)
    at InterceptingListener._callNext (/home/hey/data/src/github.com/mesg-foundation/cli/node_modules/grpc/src/client_interceptors.js:568:42)
    at InterceptingListener.onReceiveStatus (/home/hey/data/src/github.com/mesg-foundation/cli/node_modules/grpc/src/client_interceptors.js:618:8)
    at callback (/home/hey/data/src/github.com/mesg-foundation/cli/node_modules/grpc/src/client_interceptors.js:845:24)
root@m:/home/hey/data/src/github.com/mesg-foundation/cli# ./bin/run service:newstart J7EqrPcALm97MbiY6scYYBxoyEob4aR3DwegMPFoTQDS
Start service... Error: 2 UNKNOWN: service's instance is already running
Error: 2 UNKNOWN: service's instance is already running
    at Object.exports.createStatusError (/home/hey/data/src/github.com/mesg-foundation/cli/node_modules/grpc/src/common.js:91:15)
    at Object.onReceiveStatus (/home/hey/data/src/github.com/mesg-foundation/cli/node_modules/grpc/src/client_interceptors.js:1204:28)
    at InterceptingListener._callNext (/home/hey/data/src/github.com/mesg-foundation/cli/node_modules/grpc/src/client_interceptors.js:568:42)
    at InterceptingListener.onReceiveStatus (/home/hey/data/src/github.com/mesg-foundation/cli/node_modules/grpc/src/client_interceptors.js:618:8)
    at callback (/home/hey/data/src/github.com/mesg-foundation/cli/node_modules/grpc/src/client_interceptors.js:845:24)
root@m:/home/hey/data/src/github.com/mesg-foundation/cli# 
```

antho1404:
it seems that this issue is related to https://github.com/mesg-foundation/engine/pull/1023

sdk/instance/instance.go
```diff
@@ -0,0 +1,97 @@
...
+// Instance exposes service instance APIs of MESG.
+type Instance struct {
+	container  container.Container
+	serviceDB  database.ServiceDB
+	instanceDB database.InstanceDB
+}
+
+// New creates a new Instance SDK with given options.
+func New(c container.Container, serviceDB database.ServiceDB, instanceDB database.InstanceDB) *Instance {
+	return &Instance{
```
antho1404: @krhubert https://github.com/mesg-foundation/engine/pull/1037#discussion_r291853480
> sdk should use the manager as a bridge to container package

antho1404: I don't really understand the meaning of your comment

krhubert: so instance shouldn't take container as a parameter (rather a manager).

This is why `manager` struct exists to use it. container is lets say "low level" package and should be used only by `manager`.

If you start to use container it will be all over the code. This was one of the reason why manager was created - to have api hiding implementation details.

antho1404: agree but the goal here is to slowly migrate the manager (that is in the service package) to this part and only the instance SDK will have this logic to manage anything related to the container (because only the instance needs to know about that).
So basically the manager will disappear and move in instanceSDK.

Also here this is not the instance model but the instance SDK/controller, the model will not know anything about the container. Only this instanceSDK should know about that (and the global SDK to pass the dependency only)

PR #1036, opened 2019-06-07
Instance GRPC Server

antho1404:
dependency: https://github.com/mesg-foundation/engine/pull/1035
Add GRPC Server for instance with the create api


protobuf/api/instance.proto
```diff
@@ -0,0 +1,17 @@
...
+  rpc Create (CreateInstanceRequest) returns (CreateInstanceResponse) {}
+}
+
+message CreateInstanceRequest {
+  string serviceHash = 1;      // Service's sid or hash.
+  repeated string env = 2;     // Env vars to apply to service's instance on runtime.
+}
+
+message CreateInstanceResponse {
+  string hash = 2;        // Service's instance hash.
```
```suggestion
  string instanceHash = 2;        // Service's instance hash.
```
What about naming explicitly all hashes?
@krhubert and I worry that with so many hashes it will be really confused for anyone to know which hash is for what.
So we should explicitly named the hashes in the API but also in the codebase.

antho1404: The context should be enough, it doesn't really make sense to have to prefix all hashes and end up doing stuff like `instance.instanceHash` instead of `instance.hash`

NicolasMahe: it's the same logic as most of the gRPC api. the name of the parameter has just enough context to not need other extra.
Also in JS, the gRPC api has way less context than in Go because of the dynamic type..

NicolasMahe: Take a look at instance.proto and service.proto:
https://github.com/mesg-foundation/engine/blob/59f5069a0d93534fdf61ba7e143321a4c3ffe88d/protobuf/api/instance.proto
https://github.com/mesg-foundation/engine/blob/59f5069a0d93534fdf61ba7e143321a4c3ffe88d/protobuf/api/service.proto
I think that the payload are too similar and it will bring a lot of confusion: in `instance.proto` the service's hash is `serviceHash` but in service.proto it's simply `hash`.

antho1404: for me, `hash` is like an ID and an id is always called ID, if you have a model "user" you will have `user.id` and not `user.user_id` but you will keep the format for any reference/foreign key. Something like `purchase.user_id`.

I totally see your point but prefixing resources with their own name will be more confusing because it could look like a reference (like the execution parent hash). We could rename all hashes into ID to make it more "traditional" but I really feel it's against traditional development to prefix it.

antho1404: @krhubert https://github.com/mesg-foundation/engine/pull/1037#discussion_r291853262
> why starts from 2?

antho1404: because 2 is a good number... just fixed

PR #1037, opened 2019-06-07
Delete instance api

antho1404:
dependency: https://github.com/mesg-foundation/engine/pull/1036


ilgooz:
service:delete deletes the service even if an instance is running which is the opposite behavior in the proposal

protobuf/api/instance.proto
```diff
@@ -0,0 +1,26 @@
...
+  rpc Delete (DeleteInstanceRequest) returns (DeleteInstanceResponse) {}
+}
+
+message CreateInstanceRequest {
+  string serviceHash = 1;      // Service's sid or hash.
+  repeated string env = 2;     // Env vars to apply to service's instance on runtime.
+}
+
+message CreateInstanceResponse {
+  string hash = 2;        // Service's instance hash.
```
krhubert: why starts from 2?

protobuf/api/instance.proto
```diff
@@ -0,0 +1,26 @@
...
+  string serviceHash = 3; // Service's bare hash.
+}
+
+message DeleteInstanceRequest {
+  string hash = 1;      // Instance hash
+}
+
+message DeleteInstanceResponse {
+  string hash = 1;      // Instance hash
```
krhubert: why return instance hash in response if it sent in the request

database/instance_db.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// Close closes database.
+func (d *LevelDBInstanceDB) Close() error {
+	return d.db.Close()
+}
+
+// Delete deletes service from database.
+func (d *LevelDBInstanceDB) Delete(hash string) error {
+	tx, err := d.db.OpenTransaction()
```
krhubert: the key can be deleted with a single [delete](https://godoc.org/github.com/syndtr/goleveldb/leveldb#DB.Delete) call

database/instance_db.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// encode service
+	b, err := d.marshal(i)
+	if err != nil {
+		tx.Discard()
+		return err
+	}
+
+	// save instance with hash.
+	if err := tx.Put([]byte(i.Hash), b, nil); err != nil {
```
krhubert: the key can be saved with a single call without using transaction

database/instance_db.go
```diff
@@ -0,0 +1,120 @@
...
+	return &s, nil
+}
+
+// Get retrives instance by instance hash.
+func (d *LevelDBInstanceDB) Get(hash string) (*instance.Instance, error) {
+	tx, err := d.db.OpenTransaction()
+	if err != nil {
+		return nil, err
+	}
+	b, err := tx.Get([]byte(hash), nil)
```
krhubert: the key can be get with a single call without using transaction

instance/instance.go
```diff
@@ -0,0 +1,9 @@
+package instance
+
+// Instance is an instance of a service.
+// This contains a reference to a service that is running.
+// Multiple instances can run for the same service as long as they have different configurations
+type Instance struct {
+	Hash        string `hash:"-"`
+	ServiceHash string `hash:"serviceHash"`
```
krhubert: instance tags are not needed here (structhash is not used on the struct)

sdk/instance/instance.go
```diff
@@ -0,0 +1,109 @@
...
+type Instance struct {
+	container  container.Container
+	serviceDB  database.ServiceDB
+	instanceDB database.InstanceDB
+}
+
+// New creates a new Instance SDK with given options.
+func New(c container.Container, serviceDB database.ServiceDB, instanceDB database.InstanceDB) *Instance {
+	return &Instance{
+		container:  c,
```
krhubert: sdk should use the manager as a bridge to container package

database/instance_db.go
```diff
@@ -0,0 +1,120 @@
+package database
```
krhubert: There are no tests for this package

protobuf/api/instance.proto
antho1404: moved to the right PR

database/instance_db.go
antho1404: move to the right pr

database/instance_db.go
antho1404: fixed

protobuf/api/instance.proto
antho1404: I don't mind much about that, I can remove but the idea is that when we delete a resource we have the id of the resource delete in order to be able to have a full asynchrony application that can on one side send the request and on the other side just receive the result and know how to update a local store or anything.

But again I'm happy to delete it if needed

PR #1039, opened 2019-06-09
Separate unit tests and integration tests

NicolasMahe:
This PR separate integration test from unit test on CircleCI.
The unit test runs in only 8sec 🍾 
The code coverage is now only based on the unit tests. (that's why there is -5% of code coverage)

The ultimate thing to do about integration test is to transform it to a workflow that uses an application and a service to test many situation once.
I recommend to keep integration test until then.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,9 +65,9 @@ jobs:
       - run: go mod download
       - <<: *save_go_mod_cache
       - run: mkdir -p ~/.mesg
-      - <<: *restore_go_cache
+      # - <<: *restore_go_cache
       - run: go test -v -timeout 300s -p 1 -tags=integration -coverprofile=coverage.txt ./...
-      - <<: *save_go_cache
+      # - <<: *save_go_cache
       - run: bash <(curl -s https://codecov.io/bash)
 
   "lint":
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -64,9 +78,9 @@ jobs:
       - <<: *restore_go_mod_cache
       - run: go mod download
       - <<: *save_go_mod_cache
-      - <<: *restore_go_cache
+      # - <<: *restore_go_cache
```
krhubert: why this is not restored anymore?

krhubert: why go cache is commented here?

NicolasMahe: Nice catch

.circleci/config.yml
```diff
@@ -43,17 +43,31 @@ jobs:
...
       - run: docker swarm init
-      - run: docker build -t sleep docker-images/sleep/
-      - run: docker build -t http-server docker-images/http-server/
       - <<: *restore_go_mod_cache
       - run: go mod download
       - <<: *save_go_mod_cache
       - run: mkdir -p ~/.mesg
       - <<: *restore_go_cache
-      - run: go test -v -timeout 300s -p 1 -tags=integration -coverprofile=coverage.txt ./...
+      - run: go test -v -timeout 300s -coverprofile=coverage.txt ./...
```
antho1404: maybe we could even lower the timeout (but this might be a test for another time)

NicolasMahe: yes we can. all the test takes less that 8sec :)

NicolasMahe: but i'm lazy to change it...

PR #1043, opened 2019-06-10
Implement update method + fix for get data type

krhubert:


server/grpc/api/execution.go
```diff
@@ -24,7 +28,12 @@ func NewServer(sdk *sdk.SDK) *Server {
 
 // Get returns execution from given hash.
 func (s *Server) Get(ctx context.Context, req *api.GetExecutionRequest) (*definition.Execution, error) {
-	exec, err := s.sdk.GetExecution(req.Hash)
+	hash, err := hex.DecodeString(req.Hash)
```
antho1404: not sure why this is here? You changed the `GetExecutionRequest` hash to string so there is no decode needed

server/grpc/api/execution.go
```diff
@@ -57,6 +66,28 @@ func (s *Server) Stream(req *api.StreamExecutionRequest, resp api.Execution_Stre
 	return nil
 }
 
+// Update updates execution from given hash.
+func (s *Server) Update(ctx context.Context, req *api.UpdateExecutionRequest) (*api.UpdateExecutionResponse, error) {
+	hash, err := hex.DecodeString(req.Hash)
```
antho1404: same here, hash is a string

sdk/sdk.go
```diff
@@ -55,6 +55,11 @@ func (sdk *SDK) GetExecutionStream(f *ExecutionFilter) *ExecutionListener {
 	return l
 }
 
+// UpdateExecution udpate execution that matches given hash.
+func (sdk *SDK) UpdateExecution(execHash, outputs []byte, reterr error) error {
+	return sdk.SubmitResult(execHash, outputs, reterr)
```
antho1404: not blocking

maybe better to move the content of `SubmitResult` here and to make it use `UpdateExecution` as we will remove `SubmitResult`

server/grpc/api/execution.go
krhubert: yes it is, because execution.Execution uses bytes for store hash. grpc services are the only one using string for hashes so you need to decode it first.

sdk/sdk.go
krhubert: Could be that way too, but I left this transformation when we're gonna remove the submit result api.

PR #1047, opened 2019-06-12
Allow to run multiple instance of engine in dev mode

krhubert:
You can use this script same as before + start multiple engines

```
./dev

MESG_SERVICE_SERVER_ON=false MESG_SERVICE_NAME=engine-1 MESG_SERVER_PORT=50053 MESG_SERVER_ADDRESS=:50053 MESG_PATH=/root/.mesg/0 ./dev
```

### DIFF
--- a/dev
+++ b/dev
@@ -17,7 +17,6 @@ LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.EnvMarketplaceEndpoint=htt
 LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.EnvMarketplaceAddress=0xeCC1A867F871323350A1A89FcAf69629a2d5085e'"
 LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.EnvMarketplaceToken=0x5861b3dc52339d4f976b7fa5d80db6cd6f477f1b'"
 
-MESG_SERVICE_NAME=${MESG_SERVICE_NAME:-engine}
 MESG_SERVER_PORT=${MESG_SERVER_PORT:-50052}
 
 # if set to true the docker with system services will be started.
@@ -107,11 +106,11 @@ trap onexit EXIT INT
 
 function onexit {
   deleteServiceServer
-  docker service rm $MESG_SERVICE_NAME
+  docker service rm $MESG_NAME
 }
 
 docker service create \
-  --name $MESG_SERVICE_NAME \
+  --name $MESG_NAME \
   --tty \
   --label com.docker.stack.namespace=engine --label com.docker.stack.image=mesg/engine:$VERSION \
   --env MESG_NAME=$MESG_NAME \
@@ -122,4 +121,4 @@ docker service create \
   --network engine --publish $MESG_SERVER_PORT:50052 \
   mesg/engine:$VERSION
 
-docker service logs --follow --raw $MESG_SERVICE_NAME
+docker service logs --follow --raw $MESG_NAME
--- a/dev
+++ b/dev
@@ -3,14 +3,28 @@
 set -e
 
 DIR=$(pwd)
-MESG_LOG_LEVEL=${MESG_LOG_LEVEL:-debug}
+
+# mesg config variables
+MESG_NAME=${MESG_NAME:-engine}
+MESG_PATH=${MESG_PATH:-$HOME/.mesg}
+MESG_SERVER_ADDRESS=${MESG_SERVER_ADDRESS:-:50052}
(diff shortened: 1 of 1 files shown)


dev
```diff
@@ -3,14 +3,28 @@
...
 LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.EnvMarketplaceAddress=0xeCC1A867F871323350A1A89FcAf69629a2d5085e'"
 LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.EnvMarketplaceToken=0x5861b3dc52339d4f976b7fa5d80db6cd6f477f1b'"
+
+# if set to true the docker with system services will be started.
+MESG_SERVICE_SERVER_ON=${MESG_SERVICE_SERVER_ON:-true}
+
+MESG_SERVICE_NAME=${MESG_SERVICE_NAME:-engine}
+
 MESG_SERVICE_SERVER="engine-service-server"
+MESG_SERVER_PORT=${MESG_SERVER_PORT:-50052}
```
NicolasMahe: This `MESG_SERVER_PORT ` config is not useful because it's changing the internal port of the engine server.

Same logic as the docker volume. The source can change, but not the destination:
```
source=$MESG_PATH,destination=/root/.mesg
```

dev
```diff
@@ -87,15 +108,21 @@ trap onexit EXIT INT
...
 
 docker service create \
-  --name engine \
+  --name $MESG_SERVICE_NAME \
   --tty \
   --label com.docker.stack.namespace=engine --label com.docker.stack.image=mesg/engine:$VERSION \
-  --mount type=bind,source=/var/run/docker.sock,destination=/var/run/docker.sock --mount type=bind,source=$HOME/.mesg,destination=/root/.mesg \
-  --network engine --publish 50052:50052 \
+  --env MESG_NAME=$MESG_NAME \
+  --env MESG_PATH=$MESG_PATH \
```
NicolasMahe:
```suggestion
```
The `MESG_PATH` must not be injected in the docker as its change the path in the Docker.
The destination path on line 124 is still `/root/.mesg`.

dev
```diff
@@ -3,14 +3,28 @@
 set -e
 
 DIR=$(pwd)
-MESG_LOG_LEVEL=${MESG_LOG_LEVEL:-debug}
+
+# mesg config variables
+MESG_NAME=${MESG_NAME:-engine}
+MESG_PATH=${MESG_PATH:-$HOME/.mesg}
+MESG_SERVER_ADDRESS=${MESG_SERVER_ADDRESS:-:50052}
```
NicolasMahe: This `MESG_SERVER_ADDRESS ` config is not useful because it's only changing the internal port of the engine server. Each engine will get a different ip address even on the same network, thus they can all expose the server on the same internal port.
Even worst, this variable is not injected in the docker service so it will never change the port of the server..

dev
```diff
@@ -87,15 +108,21 @@ trap onexit EXIT INT
...
 docker service create \
-  --name engine \
+  --name $MESG_SERVICE_NAME \
   --tty \
   --label com.docker.stack.namespace=engine --label com.docker.stack.image=mesg/engine:$VERSION \
-  --mount type=bind,source=/var/run/docker.sock,destination=/var/run/docker.sock --mount type=bind,source=$HOME/.mesg,destination=/root/.mesg \
-  --network engine --publish 50052:50052 \
+  --env MESG_NAME=$MESG_NAME \
+  --env MESG_PATH=$MESG_PATH \
+  --env MESG_SERVER_ADDRESS=$MESG_SERVER_ADDRESS \
```
NicolasMahe:
```suggestion
```
See comment https://github.com/mesg-foundation/engine/pull/1047/files#r294078619

dev
```diff
@@ -87,15 +107,19 @@ trap onexit EXIT INT
...
 }
 
 docker service create \
-  --name engine \
+  --name $MESG_SERVICE_NAME \
   --tty \
   --label com.docker.stack.namespace=engine --label com.docker.stack.image=mesg/engine:$VERSION \
-  --mount type=bind,source=/var/run/docker.sock,destination=/var/run/docker.sock --mount type=bind,source=$HOME/.mesg,destination=/root/.mesg \
-  --network engine --publish 50052:50052 \
+  --env MESG_NAME=$MESG_NAME \
```
NicolasMahe: `MESG_SERVICE_NAME` should be merged with `MESG_NAME`.
Both need to be set with the same value...

NicolasMahe:
🎉

PR #1048, opened 2019-06-13
introduce the first version of Workflow Object & VM

ilgooz:
`Secret` can be carried as an execution tag. This Workflow VM does not depend on internal packages so, it's very lightweight, simple and easy to test.

VM uses `Process(Event)` method to receive(I of I/O) new event(/result) and uses `ExecutionRequests` channel to send(O of I/O) execution requests for new or in progress workflow lifecycles that match with state of `Event`.


workflow/vm/chain.go
```diff
@@ -0,0 +1,53 @@
...
+func (q *executionChain) Next(e Event) (ok bool, exec *Execution) {
+	if q.IsDone() || // check if there are any more execution requests left in execution chain.
+		// check if Event points to the result of the processing execution request of this execution chain.
+		// -1 means it is a standard Event, there are no previous execution requests and the first task
+		// in the execution chain will be actually put into the processing state.
+		q.index != -1 && q.indexSecret != e.Secret {
+		return false, nil
+	}
+	q.index++
+	q.indexSecret = uuid.NewV4().String()
```
krhubert: Use [rand.Reader](https://golang.org/pkg/crypto/rand/#pkg-variables) to generate secure random secrets

workflow/vm/chain.go
```diff
@@ -0,0 +1,53 @@
...
+
+// IsDone checks whether is the execution chain can create more execution requests or not.
+func (q *executionChain) IsDone() bool {
+	return q.index == len(q.w.Tasks)-1
+}
+
+// Next creates the next execution requests by using Event as its accumulator.
+// Event can be a standard event or a result of previous execution request that in execution chain.
+func (q *executionChain) Next(e Event) (ok bool, exec *Execution) {
+	if q.IsDone() || // check if there are any more execution requests left in execution chain.
```
krhubert: This comment could be above if stmt, because conditions are splited

workflow/vm/vm_test.go
```diff
@@ -0,0 +1,82 @@
+package workflowvm
+
+import (
+	"testing"
+
+	"github.com/mesg-foundation/core/workflow"
+	"github.com/stretchr/testify/require"
+)
+
+func Test(t *testing.T) {
```
krhubert: Add Name for this test

workflow/vm/vm.go
```diff
@@ -0,0 +1,164 @@
...
+		Filter: workflow.Filter{
+			TaskKey: e.TaskKey,
+		},
+	}
+}
+
+// Execution is an execution request that should be performed by the caller.
+// Each received Event from the Process() may cause multiple Execution requests
+// to be emitted by the Workflow VM.
+type Execution struct {
```
krhubert: Why Execution is redefined here? Shound't we use `execution.Execution` ?

workflow/vm/vm.go
```diff
@@ -0,0 +1,164 @@
...
+// Shutdown immediately ends vm and closes ExecutionRequests channel.
+func (v *VM) Shutdown() {
+	v.m.Lock()
+	defer v.m.Unlock()
+	v.closing = true
+	close(v.ExecuctionRequests)
+}
+
+// Event represents a new event/result input to the workflow VM.
+type Event struct {
```
krhubert: Why Execution is redefined here? Shound't we use event.Event?

workflow/vm/vm.go
```diff
@@ -0,0 +1,164 @@
...
+	Data map[string]interface{}
+
+	// Secret is generated by the Workflow VM for each execution request in a chain to check which
+	// Event belongs to which Execution request's result.
+	Secret string
+}
+
+// createTrigger creates a new trigger schema from event which is used to match the
+// event with registered workflows in VM.
+func (e Event) createTrigger() workflow.Trigger {
```
krhubert: Why is createTrigger method under Event?

It should be as a function `newTriggerFromEvent` or something simmilar

workflow/vm/vm.go
```diff
@@ -0,0 +1,164 @@
...
+	v.m.Lock()
+	defer v.m.Unlock()
+	v.closing = true
+	close(v.ExecuctionRequests)
+}
+
+// Event represents a new event/result input to the workflow VM.
+type Event struct {
+	// InstanceHash that event is emitted from.
+	InstanceHash string
```
krhubert:
```suggestion
	InstanceHash []byte
```

workflow/vm/vm.go
```diff
@@ -0,0 +1,164 @@
...
+// VM is a virtual machine that runs workflows.
+type VM struct {
+	// ExecuctionRequests keeps execution requests, caller should read this chan
+	// and perform executions.
+	ExecuctionRequests chan *Execution
+
+	m               sync.Mutex // protects following.
+	workflows       []workflow.Workflow
+	executionChains []*executionChain
+	closing         bool
```
krhubert:
```suggestion
	closed          bool
```

workflow/vm/chain.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// IsDone checks whether is the execution chain can create more execution requests or not.
+func (q *executionChain) IsDone() bool {
+	return q.index == len(q.w.Tasks)-1
+}
+
+// Next creates the next execution requests by using Event as its accumulator.
+// Event can be a standard event or a result of previous execution request that in execution chain.
+func (q *executionChain) Next(e Event) (ok bool, exec *Execution) {
```
krhubert: use error (io.EOF) that is used by read/next functions to represent end of data in general.

workflow/vm/vm.go
```diff
@@ -0,0 +1,164 @@
...
+				break
+			}
+		}
+	}
+	// event can also be a start of some new lifecycles for many workflows.
+	// create a full trigger spec for the event and find out the matching workflows.
+	t := e.createTrigger()
+	for _, w := range v.workflows {
+		// check if workflow matches with trigger created from event.
+		if ok := w.Match(t); ok {
```
krhubert:
```suggestion
		if w.Match(t) {
```

workflow/vm/vm.go
```diff
@@ -0,0 +1,164 @@
...
+
+	// TaskKey of execution.
+	TaskKey string
+
+	// Inputs of the execution.
+	Inputs map[string]interface{}
+
+	// Secret is generated by the Workflow VM for each execution request in a chain to check which
+	// Event belongs to which Execution request's result.
+	Secret string
```
krhubert: If secret is used to identify Execution why not to Use execution hash and keep map of this in vm, instead add o new variable. 

Also if we want to add new variable then it should be called something with ID or Hash, as secret is rather for passwords,passpharse etc...

workflow/workflow.go
```diff
@@ -0,0 +1,42 @@
+package workflow
+
+import (
+	"github.com/mesg-foundation/core/x/xstrings"
+)
+
+// Workflow represents a new workflow.
+type Workflow struct {
+	// Hash of the workflow.
+	Hash string
```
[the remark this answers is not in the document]
ilgooz: I'll do this string to byte slice changes but after we actually change them to byte slice in where we create them first

workflow/workflow.go
```diff
@@ -0,0 +1,42 @@
...
+	// Trigger of the workflow.
+	Trigger Trigger
+
+	// Tasks of the workflow.
+	Tasks []Task
+}
+
+// Match checks if the workflow's trigger matches with given trigger spec.
+func (w *Workflow) Match(trigger Trigger) bool {
+	return xstrings.SliceContains([]string{"*", "", trigger.InstanceHash}, w.Trigger.InstanceHash) &&
```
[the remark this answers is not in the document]
ilgooz: we can stop using "*", having it is nice but I'm ok to rm.

we don't have not-equal comparison for now, this is something that may come in future.

workflow/workflow.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Trigger is a pattern to match first in order to start workflow's lifecycle.
+type Trigger struct {
+	InstanceHash string
+	EventKey     string
+	Filter       Filter
+}
+
+// Filter keeps optional trigger conditions.
+type Filter struct {
```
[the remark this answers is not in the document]
ilgooz: if we have multiple filter types, I'll make the name more explicit like this

workflow/vm/chain.go
ilgooz: maybe in future, I don't want to make initial the code longer, uuid is simple and enough for now

workflow/vm/vm.go
ilgooz: this is not same as `execution.Execution`. this is actually an execution request. i may change the name to `ExecutionRequest`.

also, we cannot create a new `execution.Execution` in `workflowvm` because creation should be made one by the `execution.Create()` func which is actually called by the `sdk.ExecuteTask()`. so if you want to reuse execute task sdk, you have to pass the params it needs not `execution.Execution`. It is also nice to have `workflowvm` not fully aware of these types for simplicity.

workflow/vm/vm.go
ilgooz: https://github.com/mesg-foundation/engine/pull/1048#discussion_r293338121

also, event here is combination of Event & Result. `event.Event` just represents the Event itself. In future, we'll be merging Result to Event as discussed with @antho1404.

workflow/vm/vm.go
ilgooz: it's because we're creating a trigger spec from Event. I may change the name to `createTrigger Spec()`. the returned trigger contains all conditions to match with this event. if a workflow's trigger doesn't match at least one condition from this trigger spec, workflow won't match. empty or "*" conditions in workflow trigger considered as a match.

workflow/vm/vm.go
ilgooz: workflowvm does not and cannot now about execution ids. execution id is generated after execution is actually made. here, we're just creating the execution requests. this is the same logic why we've introduced execution tags since execution ids weren't just enough.

this Secret is not a new variable. it should be converted to an execution tag by the caller of workflowvm.

antho1404:
I don't understand this PR, it's not what has been discussed, there is no part that listens for the executions and the events, it doesn't create any execution and this chain is what was already developed with the execution db and the parent hash...

ilgooz:
> I don't understand this PR, it's not what has been discussed, there is no part that listens for the executions and the events, it doesn't create any execution and this chain is what was already developed with the execution db and the parent hash...

this is the first part of feature. listening and executing part will be done outside in core/main.go or sdk pkg.

what do you mean by execution db? it's a different subject it's only responsible for storing done executions. vm has to keep track of executions it has to made not the db.

PR #1049, opened 2019-06-13
add Workflow types to Service types

ilgooz:
* update service.Service type and definition.Service proto with Workflow type.
* create type bindings for definition.Workflow proto to service.Service.Workflow.

depends on https://github.com/mesg-foundation/engine/pull/1048


service/service.go
```diff
@@ -56,6 +57,9 @@ type Service struct {
 	// Source is the hash id of service's source code on IPFS.
 	Source string `hash:"name:9"`
 
+	// Workflows keeps the workflows of service and is optional.
+	Workflows []workflow.Workflow `hash:"name:10"`
```
krhubert: why service contains `workflows`? The workflow manages services so it should be another way around, or they should be next to each other.

antho1404: have a look at this post https://forum.mesg.com/t/integration-of-workflows-in-services-solves-service-composition/283

krhubert: I saw it, and I don't understand why workflows are one of the field in the service struct.

The workflow should be totally independent of the service. In fact, it doesn't need to know about service a all. As you wrote in workflow implementation it just a graph that get/recive data from the service. So as I see it, the workflow should be only focused on the flow of data. The other part will be ETL process from service into workflow. Otherwise again everything will be mixed together and tightly coupled which will lead to many problems (basic one is with tests):

- in order to test workflow package service must be used
- in order to load data into workflow the service must be used

And it could be done with simple unit tests, but it can't be so tightly coupled.

PR #1050, opened 2019-06-14
Marketplace backward compatibility

antho1404:
just by making the type attribute optional, both version of services from 0.9 and 0.10 can be returned from the service


antho1404:
failing tests are not related to this PR
Another PR fixes these tests https://github.com/mesg-foundation/engine/pull/1051

PR #1060, opened 2019-06-18
sdk/instancesdk: [fix] pass env vars to service and its dependencies on create

### LINKED ISSUE #1059: Issue with the Instance create
Here is my service.json (got form the compiled command)

```json
{"sid":"testservice","name":"test service","tasks":[{"key":"taskx","inputs":[{"key":"foo","type":"String","object":[]},{"key":"bar","type":"String","object":[]}],"outputs":[{"key":"result","type":"String",
"object":[]}]}],"events":[{"key":"event","data":[{"key":"x","type":"Any","object":[]}]}],"dependencies":[{"key":"nginx","image":"nginx","volumes":["/etc/nginx"]}],"configuration":{"volumes":["/volume/test/
"],"volumesFrom":["nginx"],"env":["ENVA=do_not_override","ENVB=override"]},"source":"Qmeuu7A7A8DERuRe8apL3kaHPaPtMrUrATcSDNuB2LW1bF"}
```

The deploy and start seems working
```bash
mesg-cli-dev service:newdeploy "$(cat service.json)"
mesg-cli-dev service:newstart testservice --env ENVB=1
```

but:

Logs on the service are not accessible because the engine cannot access the dependency.
```bash
mesg-cli-dev service:logs testservice

/Users/antho/prog/MESG/cli/node_modules/grpc/src/common.js:91
  let error = new Error(message);
              ^
Error: 2 UNKNOWN: Error response from daemon: service engine-d0a725b9672ae318983612c4913cade9fb58dfd9-nginx not found
```

The service is not running because the env variable (ENVB) needs to be changed when starting the instance (which should be done by mesg-cli-dev service:newstart testservice --env ENVB=1)
```bash
docker service logs mjddhxljgrv9

engine-c4f9f16d407ab5b6afd78ab052e221a988cb14a1-service.1.khfnci8o88nl@mesg-dev    | /app/src/index.ts:42
engine-c4f9f16d407ab5b6afd78ab052e221a988cb14a1-service.1.khfnci8o88nl@mesg-dev    | assert.notEqual(process.env.ENVB, 'override')
engine-c4f9f16d407ab5b6afd78ab052e221a988cb14a1-service.1.khfnci8o88nl@mesg-dev    |        ^
engine-c4f9f16d407ab5b6afd78ab052e221a988cb14a1-service.1.khfnci8o88nl@mesg-dev    | AssertionError [ERR_ASSERTION]: 'override' != 'override'
```

ilgooz:
and add env vars field to instance.Instance to keep runtime env vars.


sdk/instance/start.go
```diff
@@ -31,6 +31,7 @@ func (i *Instance) start(inst *instance.Instance) (serviceIDs []string, err erro
 	}
 	_, port, _ := xnet.SplitHostPort(conf.Server.Address)
 	endpoint := conf.Name + ":" + strconv.Itoa(port)
+	env := xos.EnvMapToSlice(xos.EnvMergeMaps(xos.EnvSliceToMap(srv.Configuration.Env), xos.EnvSliceToMap(inst.Env)))
```
krhubert:
```suggestion
	env := append(srv.Configuration.Env, inst.Env...)
```

or just use EnvMergeSlices

NicolasMahe:
I would really like to avoid to store envs in instances to avoid any sensitive data to leak on the future network.
The goal is for the instance db to be sync across the network. And we don't want to sync sensitive data..
The function `instance.Start` should as parameters the envs and only pass them to Docker.

sdk/instance/instance.go
```diff
@@ -65,7 +65,7 @@ func (i *Instance) Create(id string, env []string) (*instance.Instance, error) {
 		return nil, err
 	}
 
-	// overwrite default env vars with user defined ones.
+	// calculate env vars by overwriting user defined one's with defaults.
 	instanceEnv := xos.EnvMergeMaps(xos.EnvSliceToMap(srv.Configuration.Env), xos.EnvSliceToMap(env))
```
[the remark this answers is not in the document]
ilgooz: I chose to not optimise this in this PR since it's not the subject of this one, and I believe that env vars type should be changed to a map instead a slice so there might be some small refactoring coming in future.

sdk/instance/start.go
ilgooz: https://github.com/mesg-foundation/engine/pull/1060#discussion_r295254438

antho1404:
Valid but this doesn't totally solve the #1059

PR #1065, opened 2019-06-19
Add hash filter to execution proto service

krhubert:


protobuf/api/execution.proto
```diff
@@ -31,6 +31,9 @@ message StreamExecutionRequest{
     // Note: to filter multiple statues or flags, eg:
     // Failed | Completed
     definition.Status status = 1;
+
+    // hash is a hash for filtering execution by instance hash.
+    string hash = 2;
```
NicolasMahe: can you rename to `instanceHash`.
We don't know if it executionHash, or whatever

sdk/execution/execution_listener.go
```diff
@@ -18,6 +21,11 @@ func (f *Filter) Match(e *execution.Execution) bool {
 	if f == nil {
 		return true
 	}
+
+	if len(f.Hash) != 0 && !bytes.Equal(e.Hash, f.Hash) {
```
NicolasMahe: It should filter against the InstanceHash/ServiceHash, not the ExecutionHash.

sdk/execution/execution_listener.go
```diff
@@ -4,20 +4,27 @@ import (
...
+	Tags        []string
 }
 
 // Match matches execution.
 func (f *Filter) Match(e *execution.Execution) bool {
 	if f == nil {
 		return true
 	}
+
+	if len(f.ServiceHash) != 0 && base58.Encode(f.ServiceHash) == e.ServiceHash {
```
NicolasMahe:
```suggestion
	if len(f.ServiceHash) != 0 && base58.Encode(f.ServiceHash) != e.ServiceHash {
```

NicolasMahe:
I approved this PR but I think a new PR should be done to make all filter totally optional!

If serviceHash is not set, an error is returned.
If filter.status is not set, then the default value 0 (Unknown) is used that will NEVER match any event...

NicolasMahe:
Also, to test grpc api, you can use grpcurl:
```
grpcurl -plaintext -d '{"filter": {"serviceHash": "47JaNMMKeT14W9hafyPQqb7gWw24rk4yT9yT4dcELFcR", "status": "3"}}' localhost:50052 api.Execution/Stream
```

sdk/execution/execution_listener.go
krhubert: Done

krhubert:
> I approved this PR but I think a new PR should be done to make all filter totally optional!

But the filters are optional - if serviceHash is not set or filter.status is not set then they are not used.

NicolasMahe:
> > I approved this PR but I think a new PR should be done to make all filter totally optional!
> 
> But the filters are optional - if serviceHash is not set or filter.status is not set then they are not used.

Yes inside the sdk, wrong on the gRPC API implementation: https://github.com/mesg-foundation/engine/pull/1065/files#diff-aa6e9a2e558612f93397e9a80d219d26R31
```go
func (s *Server) Stream(req *api.StreamExecutionRequest, resp api.Execution_StreamServer) error {
	hash, err := base58.Decode(req.Filter.ServiceHash)
```
`base58.Decode` is returning an error if `req.Filter.ServiceHash` is empty.
And similar problem with status:
```go
Statuses:    []execution.Status{execution.Status(req.Filter.Status)},
```
A condition on default value of both `req.Filter.ServiceHash` and `req.Filter.Status` should be added.

PR #1066, opened 2019-06-19
Add env to proto definition of service dependencies

NicolasMahe:


protobuf/definition/service.proto
```diff
@@ -63,4 +64,5 @@ message Dependency {
   repeated string ports = 4;        // List of ports the container exposes.
   repeated string args = 6;         // Args to pass to the container.
   string command = 5;               // Command to run the container.
+  repeated string env = 9;          // Default env vars to apply to service's instance on runtime.
```
ilgooz:
```suggestion
  repeated string env = 9;          // Default env vars to apply to dependency on runtime.
```

PR #1067, opened 2019-06-19
Add get service new API

NicolasMahe:
Dependant on https://github.com/mesg-foundation/engine/pull/1066

Add the new api get service.

Monday task: https://mesg.monday.com/boards/231078139/pulses/256524680


server/grpc/api/service.go
```diff
@@ -100,3 +103,101 @@ func fromProtoDependencies(deps []*definition.Dependency) []*service.Dependency
 	}
 	return ds
 }
+
+// ToProtoService converts an internal service struct to the protobuf definition
+// TODO: should not be public. Need to move server/grpc/service.go to server/grpc/api/service.go
+func ToProtoService(s *service.Service) *definition.Service {
```
krhubert: This function is already defined in `server/grpc/core/proto.go`, 

make it public or create a function in definition package.

server/grpc/service.go
```diff
@@ -3,7 +3,9 @@ package grpc
 import (
 	"context"
 
+	protoApi "github.com/mesg-foundation/core/protobuf/api"
```
krhubert: do not use camelCase in imports.

protobuf/api/service.proto
```diff
@@ -8,6 +8,9 @@ package api;
 // right now we have conflicts with api.Service.
 service ServiceX {
   rpc Create (CreateServiceRequest) returns (CreateServiceResponse) {}
+
+  // Get returns a single Service specified in a request.
+  rpc Get(GetServiceRequest) returns (definition.Service) {}
```
ilgooz: it's better to apply same pattern of storing return values in a wrapper. otherwise it'll not be consistent for all apis.

e.g.
```
message GetServiceResponse {
 definition.Service service = 1;
}
```

protobuf/api/service.proto
```diff
@@ -18,3 +21,8 @@ message CreateServiceResponse {
   string sid = 1;
   string hash = 2;
 }
+
+// GetServiceRequest defines request to retrieve a single service.
+message GetServiceRequest {
+  string hash = 1;
```
ilgooz: personally, hashOrSid is a better name

sdk/service/service.go
```diff
@@ -87,3 +87,8 @@ func (s *Service) Create(srv *service.Service) error {
 
 	return s.db.Save(srv)
 }
+
+// Get returns the service that matches given hash.
+func (s *Service) Get(hash string) (*service.Service, error) {
```
ilgooz: personally, hashOrSid is a better name

server/grpc/api/service.go
NicolasMahe: Good catch! I searched for those function but I didn't find them...
I removed to "old" ones in favor of these one because they are in the right place. The `server/grpc/core` will be deleted in the future.

server/grpc/service.go
NicolasMahe: Actually this import is useless. It was already imported as `protobuf_api` 😂

protobuf/api/service.proto
NicolasMahe: Actually this is to be consistent with the new API.
Check new Execution grpc api ;)

protobuf/api/service.proto
NicolasMahe: `Sid` will be replaced by a dedicated system in the future OR will be managed by CLI.

sdk/service/service.go
NicolasMahe: Sid will be replaced by a dedicated system in the future OR will be managed by CLI.
All the API will use only hash as unique identifier of ANY resource.

PR #1070, opened 2019-06-19
Expose the execution gRPC service on the server

NicolasMahe:


server/grpc/server.go
```diff
@@ -106,11 +107,13 @@ func (s *Server) register() error {
 	coreServiceServer := NewServiceServer(s.sdk)
 	instanceServer := NewInstanceServer(s.sdk)
 	serviceServer := service.NewServer(s.sdk)
+	apiServer := api.NewServer(s.sdk)
```
antho1404:
```suggestion
	apiExecutionServer := api.NewServer(s.sdk)
```
the api should be `api.NewExecutionServer(s.sdk)` in the first place so could be good to be more explicit on the variables here otherwise it's really confusing

krhubert: We'll change it :)

PR #1071, opened 2019-06-19
add servicesdk.Delete() feature and its grpc API

ilgooz:


sdk/sdk.go
```diff
@@ -31,7 +31,7 @@ type SDK struct {
 func New(m manager.Manager, c container.Container, db database.ServiceDB, instanceDB database.InstanceDB, execDB database.ExecutionDB) *SDK {
 	ps := pubsub.New(0)
 	return &SDK{
-		Service:   servicesdk.New(m, c, db, execDB),
+		Service:   servicesdk.New(m, c, db, instanceDB, execDB),
```
antho1404: Instance can know service but service cannot know about instance. This is following the data structure. Instance depends on service but service doesn't depends on instance so let's not create this dependency here. 

The api (that has access to service and instance) should do this check.

ilgooz: done

server/grpc/service.go
```diff
@@ -26,3 +27,16 @@ func (s *ServiceServer) Create(ctx context.Context, request *protobuf_api.Create
...
+
+// Delete deletes service by hash or sid.
+func (s *ServiceServer) Delete(ctx context.Context, request *protobuf_api.DeleteServiceRequest) (*protobuf_api.DeleteServiceResponse, error) {
+	// first, check if service has any running instances.
+	instances, err := s.sdk.Instance.GetAllByService(request.HashOrSid)
+	if err != nil {
+		return nil, err
+	}
+	if len(instances) > 0 {
+		return nil, errors.New("service has running instances. in order to delete the service, stop its instances first")
```
ilgooz: @NicolasMahe you said before about listing running instance hashes here so devs can stop them. i think it's not relevant here, devs should do list running instances of service by using the cli

server/grpc/service.go
```diff
@@ -26,3 +27,20 @@ func (s *ServiceServer) Create(ctx context.Context, request *protobuf_api.Create
...
+	if err != nil {
+		return nil, err
+	}
+	// first, check if service has any running instances.
+	instances, err := s.sdk.Instance.GetAllByService(srv.Hash)
+	if err != nil {
+		return nil, err
+	}
+	if len(instances) > 0 {
+		return nil, errors.New("service has running instances. in order to delete the service, stop its instances first")
```
ilgooz: @NicolasMahe you said before about listing running instance hashes here so devs can stop them. i think it's not relevant here, devs should do list running instances of service by using the cli

server/grpc/service.go
```diff
@@ -26,3 +27,20 @@ func (s *ServiceServer) Create(ctx context.Context, request *protobuf_api.Create
...
 	return &protobuf_api.CreateServiceResponse{Sid: srv.Sid, Hash: srv.Hash}, nil
 }
+
+// Delete deletes service by hash or sid.
+func (s *ServiceServer) Delete(ctx context.Context, request *protobuf_api.DeleteServiceRequest) (*protobuf_api.DeleteServiceResponse, error) {
+	srv, err := s.sdk.GetService(request.HashOrSid)
+	if err != nil {
+		return nil, err
+	}
+	// first, check if service has any running instances.
```
ilgooz: not a fan of having this check here. high level sdk pkg can be a good place to put the service deletion logic. because it has access to both instancesdk and servicesdk. with this improvements api will not have any extra logic which is a good thing if there'll be other type of apis in future

PR #1075, opened 2019-06-19
added instancesdk.Get() feature and its gRPC API

ilgooz:
depends on https://github.com/mesg-foundation/engine/pull/1074


protobuf/api/instance.proto
```diff
@@ -1,12 +1,23 @@
 syntax = "proto3";
 
+import "protobuf/definition/instance.proto";
+
 package api;
 
 service Instance {
+  rpc Get (GetInstanceRequest) returns (GetInstanceResponse) {}
```
```suggestion
  rpc Get (GetInstanceRequest) returns (definition.Instance) {}
```

I really recommend to directly returning the Instance definition. That's the directly on the new API. Simpler is better.
Check:
https://github.com/mesg-foundation/engine/blob/acc3e0877af3eecb98a6b441997ff35b11fbd6b4/protobuf/api/execution.proto#L9-L15

NicolasMahe: Also, `CreateInstanceResponse ` should be replaced also by the instance proto definition:
https://github.com/mesg-foundation/engine/blob/acc3e0877af3eecb98a6b441997ff35b11fbd6b4/protobuf/api/instance.proto#L15-L18


monday task: https://mesg.monday.com/boards/231078139/pulses/231079102/posts/361694557

ilgooz: > I really recommend to directly returning the Instance definition. That's the directly on the new API. Simpler is better.
> Check:
> https://github.com/mesg-foundation/engine/blob/acc3e0877af3eecb98a6b441997ff35b11fbd6b4/protobuf/api/execution.proto#L9-L15

I don't agree with this one as Discussed before but I can commit the changes as you like, if you prefer that way.

NicolasMahe: Let's reach consensus on the forum: https://forum.mesg.com/t/standardisation-of-the-grpc-apis/320

ilgooz: > Also, `CreateInstanceResponse ` should be replaced also by the instance proto definition:
> https://github.com/mesg-foundation/engine/blob/acc3e0877af3eecb98a6b441997ff35b11fbd6b4/protobuf/api/instance.proto#L15-L18
> 
> monday task: https://mesg.monday.com/boards/231078139/pulses/231079102/posts/361694557

done

PR #1077, opened 2019-06-21
 Implement List for Service

krhubert:
Dep on: https://github.com/mesg-foundation/engine/pull/1072

### DIFF
--- a/server/grpc/api/service.go
+++ b/server/grpc/api/service.go
@@ -63,9 +63,24 @@ func (s *ServiceServer) Get(ctx context.Context, req *protobuf_api.GetServiceReq
 		return nil, err
 	}
 
-	exec, err := s.sdk.Service.Get(hash)
+	service, err := s.sdk.Service.Get(hash)
 	if err != nil {
 		return nil, err
 	}
-	return ToProtoService(exec), nil
+	return ToProtoService(service), nil
+}
+
+// List returns all services.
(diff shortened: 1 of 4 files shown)


antho1404:
why using a stream here. We don't have any pub/sub system for services yet so there is no reason to have a stream here. Let's simplify and return a list of services directly and let grpc/protobuf handle that

NicolasMahe:
@krhubert I agree with Anthony. A stream seems overkill for now.. Let's develop the simplest api we can.

server/grpc/api/service.go
```diff
@@ -63,9 +63,24 @@ func (s *ServiceServer) Get(ctx context.Context, req *protobuf_api.GetServiceReq
...
+// List returns all services.
+func (s *ServiceServer) List(ctx context.Context, req *protobuf_api.ListServiceRequest) (*protobuf_api.ListServiceResponse, error) {
+	services, err := s.sdk.Service.List()
+	if err != nil {
+		return nil, err
+	}
+
+	resp := &protobuf_api.ListServiceResponse{}
+	for _, service := range services {
+		resp.Services = append(resp.Services, ToProtoService(service))
```
NicolasMahe: I think there is a `ToProtoService` function. Let's use it or remove it.

antho1404: `mapping.go` > `ToProtoServices`

PR #1087, opened 2019-06-24
System service instance

antho1404:
Dependency: https://github.com/mesg-foundation/engine/pull/1085

- [x] Load system services based on service definition
- [x] Use new `service.Create` and `instance.Create` functions
- [x] Update `dev` script
- [x] Inject full service definition
  - [x] Add compiled version of the service in the sources
  - [x] Load compiled version in the binary with `go-assets` or something like that
- [ ] Fix Ethwallet with new API

fix https://github.com/mesg-foundation/engine/issues/1115


antho1404:
@NicolasMahe can you take care of ethwallet service

config/services.go
```diff
@@ -1,17 +1,22 @@
...
+// The service name should be the name of the directory inside `systemservices`
 // example:
 // 		var (
-// 			BarURL string
+// 			barCompiled string
 // 		)
 var (
-	EthwalletURL   string
-	MarketplaceURL string
+	ethwalletCompiled   string
```
antho1404:
```suggestion
	// ethwalletCompiled   string
```

dev
```diff
@@ -19,51 +20,22 @@ LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.EnvMarketplaceToken=0x5861
...
-    LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.${varname}URL=http://$MESG_SERVICE_SERVER/$file.tar.gz'"
-    cd ../..
+# from scripts/build-core.sh
+# upload system services
+for s in systemservices/* ; do
+  if [ -d "$s" ]; then
+    pushd $s > /dev/null
+    name=$(basename "$s")
+    varname="${name}"
+    LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.${varname}Compiled=$(cat compiled.json)'"
```
NicolasMahe:
```suggestion
    LDFLAGS+=" -X 'github.com/mesg-foundation/engine/config.${varname}Compiled=$(cat compiled.json)'"
```

dev
```diff
@@ -19,51 +20,22 @@ LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.EnvMarketplaceToken=0x5861
...
-for i in systemservices/* ; do
-  if [ -d "$i" ]; then
-    echo "Creating archive for $i..."
-    cd $i
-    file=$(basename "$i")
-    tar -czf ../../tmp-systemservices/$file.tar.gz --exclude-from=.dockerignore .
-    varname="$(tr '[:lower:]' '[:upper:]' <<< ${file:0:1})${file:1}"
-    LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.${varname}URL=http://$MESG_SERVICE_SERVER/$file.tar.gz'"
-    cd ../..
+# from scripts/build-core.sh
```
NicolasMahe:
```suggestion
# from scripts/build-engine.sh
```

dev
NicolasMahe: where the compiled.json is from?

scripts/build-core.sh
```diff
@@ -13,14 +13,8 @@ for s in systemservices/* ; do
...
-    archive="$name.tar.gz"
-
-    id=$(
-      tar -czf - --exclude-from=.dockerignore . |
-      curl -s -F "file=@-;filename=${archive}" http://$MESG_SERVICE_SERVER:5001/api/v0/add |
-      jq -r .Hash
-    )
-    LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.${varname}URL=http://$MESG_SERVICE_SERVER:8080/ipfs/$id'"
+    mesg-cli service:compile | jq -c . > compiled.json
+    LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.${varname}Compiled=$(cat compiled.json)'"
```
NicolasMahe:
```suggestion
    LDFLAGS+=" -X 'github.com/mesg-foundation/core/engine.${varname}Compiled=$(cat compiled.json)'"
```

sdk/instance/instance.go
```diff
@@ -148,3 +149,12 @@ func (i *Instance) Delete(hash hash.Hash, deleteData bool) error {
...
 	return i.instanceDB.Delete(hash)
 }
+
+// AlreadyExistsError is an not found error.
+type AlreadyExistsError struct {
+	Hash hash.Hash
+}
+
+func (e *AlreadyExistsError) Error() string {
+	return fmt.Sprintf("instance already exists")
```
```suggestion
	return fmt.Sprintf("instance %q already exists", e.Hash.ToString())
```

sdk/service/service.go
```diff
@@ -99,3 +100,12 @@ func (s *Service) Get(hash hash.Hash) (*service.Service, error) {
...
 	return s.serviceDB.All()
 }
+
+// AlreadyExistsError is an not found error.
+type AlreadyExistsError struct {
+	Hash hash.Hash
+}
+
+func (e *AlreadyExistsError) Error() string {
+	return fmt.Sprintf("service already exists")
```
NicolasMahe:
```suggestion
	return fmt.Sprintf("service %q already exists", e.Hash.ToString())
```

server/grpc/core/core.go
```diff
@@ -25,12 +25,16 @@ func (s *Server) Info(ctx context.Context, request *coreapi.InfoRequest) (*corea
...
+	}
+	services := make([]*coreapi.InfoReply_CoreService, len(servicesFromConfig))
+	for i, s := range servicesFromConfig {
 		services[i] = &coreapi.InfoReply_CoreService{
-			Sid:  s.Sid,
-			Hash: s.Hash.String(),
-			Url:  s.URL,
+			Sid:  s.Definition.Sid,
+			Hash: s.Instance.Hash.String(),
+			Url:  s.Definition.Source,
```
NicolasMahe: Is it useful to return the Source?
I think it will be nice to also have ServiceHash along side InstanceHash.
But anyway `Hash = InstanceHash` is the most important one 👍

dev
```diff
@@ -10,6 +10,7 @@ MESG_PATH=${MESG_PATH:-$HOME/.mesg}
 MESG_LOG_FORMAT=${MESG_LOG_FORMAT:-text}
 MESG_LOG_FORCECOLORS=${MESG_LOG_FORCECOLORS:-false}
 MESG_LOG_LEVEL=${MESG_LOG_LEVEL:-debug}
+MESG_SERVICE_SERVER=ipfs.app.mesg.com
```
NicolasMahe:
```suggestion
```
this is not used

dev
NicolasMahe: also, why not compiling the service also in this script? like scripts/build-engine.sh is doing? optimization?

antho1404: yes because of optimization but not only, we should only have the compiled version of the service and if we have the compiled version there is no need of having the sources of the system service in the engine and in that case, the compiled version is a snapshot of the exact version we want from this system service that we can manually update with something like 

```mesg-cli service:compile https://github.com/mesg-foundation/service-marketplace > systemservices/marketplace.json```

File: server/grpc/core/core.go (hunk 7 above)
Comment (antho1404) [reply server/grpc/core/core.go]: just didn't want to break the api and return what was initially returned. Let's remove it if needed in another PR

File: sdk/instance/instance.go [hunk 9]
```diff
@@ -148,3 +149,12 @@ func (i *Instance) Delete(hash hash.Hash, deleteData bool) error {
...
 	return i.instanceDB.Delete(hash)
 }
+
+// AlreadyExistsError is an not found error.
+type AlreadyExistsError struct {
+	Hash hash.Hash
+}
+
+func (e *AlreadyExistsError) Error() string {
+	return fmt.Sprintf("instance %q already exists", e.Hash.ToString())
```
```suggestion
	return fmt.Sprintf("instance %q already exists", e.Hash.String())
```

sdk/service/service.go
```diff
@@ -99,3 +100,12 @@ func (s *Service) Get(hash hash.Hash) (*service.Service, error) {
...
 	return s.serviceDB.All()
 }
+
+// AlreadyExistsError is an not found error.
+type AlreadyExistsError struct {
+	Hash hash.Hash
+}
+
+func (e *AlreadyExistsError) Error() string {
+	return fmt.Sprintf("service %q already exists", e.Hash.ToString())
```
NicolasMahe:
```suggestion
	return fmt.Sprintf("service %q already exists", e.Hash.String())
```

PR #1092, opened 2019-06-24
Add comments on the new gRPC APIS

NicolasMahe:
Dependant on https://github.com/mesg-foundation/engine/pull/1085

### DIFF
--- a/protobuf/api/execution.proto
+++ b/protobuf/api/execution.proto
@@ -23,32 +23,38 @@ service Execution {
 }
 
 // GetExecutionRequest defines request to retrieve a single execution.
-// **Example**	
-// TODO: add JSON example
 message GetExecutionRequest {
-  string hash = 1;  // Execution's hash to fetch.
+  // Execution's hash to fetch.
+  string hash = 1;
 }
 
 // StreamExecutionRequest defines request to retrieve a stream of executions.
-// **Example**	
-// TODO: add JSON example
 message StreamExecutionRequest{
   // Filter contains filtering criteria.
   message Filter {
-    definition.Status status = 1; // Status to filter executions.
-    string serviceHash = 2;       // Service's hash to filter executions.
+    // Status to filter executions.
+    definition.Status status = 1;
+
+    // Service's hash to filter executions.
+    string serviceHash = 2;
   }
-  Filter filter = 1;  // Filter used to filter a stream of executions.
+
+  // Filter used to filter a stream of executions.
+  Filter filter = 1;
 }
 
 // UpdateExecutionRequest defines request for execution update.
-// **Example**	
-// TODO: add JSON example
 message UpdateExecutionRequest {
-  string hash = 1;      // Hash represents execution.
-  oneof result {        // result pass to execution
-    string outputs = 2; // outputs is a success result.
-    string error = 3;   // error is an error result.
+  // Hash represents execution.
+  string hash = 1;
+
+  // result pass to execution
+  oneof result {
+    // outputs is a success result.
+    string outputs = 2;
+
+    // error is an error result.
+    string error = 3;
   }
 }
 
--- a/protobuf/api/execution.proto
+++ b/protobuf/api/execution.proto
@@ -22,14 +22,14 @@ service Execution {
   rpc Update(UpdateExecutionRequest) returns (UpdateExecutionResponse) {}
 }
 
-// GetExecutionRequest defines request to retrieve a single execution.//
+// GetExecutionRequest defines request to retrieve a single execution.
 // **Example**	
 // TODO: add JSON example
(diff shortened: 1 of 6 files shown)


protobuf/api/execution.proto
```diff
@@ -18,38 +22,33 @@ service Execution {
   rpc Update(UpdateExecutionRequest) returns (UpdateExecutionResponse) {}
 }
 
-// GetExecutionRequest defines request to retrieve a single execution.
+// GetExecutionRequest defines request to retrieve a single execution.//
```
antho1404:
```suggestion
// GetExecutionRequest defines request to retrieve a single execution.
```

protobuf/api/execution.proto
```diff
@@ -18,38 +22,33 @@ service Execution {
...
+// **Example**	
+// TODO: add JSON example
 message GetExecutionRequest {
-  string hash = 1;
+  string hash = 1;  // Execution's hash to fetch.
 }
 
-// StreamExecutionRequest defines request to retrieve a stream of executions.
+// StreamExecutionRequest defines request to retrieve a stream of executions.//
```
antho1404:
```suggestion
// StreamExecutionRequest defines request to retrieve a stream of executions.
```

protobuf/api/execution.proto
```diff
@@ -18,38 +22,33 @@ service Execution {
...
+    string serviceHash = 2;       // Service's hash to filter executions.
   }
-
-  // Filter used to filter a stream of executions.
-  Filter filter = 1;
+  Filter filter = 1;  // Filter used to filter a stream of executions.
 }
 
-// UpdateExecutionRequest defines request for execution update.
+// UpdateExecutionRequest defines request for execution update.//
```
antho1404:
```suggestion
// UpdateExecutionRequest defines request for execution update.
```

protobuf/api/execution.proto
```diff
@@ -19,37 +23,32 @@ service Execution {
...
   // Filter contains filtering criteria.
   message Filter {
-    // Status for filtering execution by status.
-    // Note: to filter multiple statues or flags, eg:
-    // Failed | Completed
-    definition.Status status = 1;
-
-    // serviceHash is a hash for filtering execution by service hash.
-    string serviceHash = 2;
+    definition.Status status = 1; // Status to filter executions.
```
krhubert: I feel putting a comment in one line make it less readable. 
Also if the comment is a bit longer (like before) there is no option to make it readable at all.

protobuf/api/execution.proto
```diff
@@ -19,37 +23,32 @@ service Execution {
 }
 
 // GetExecutionRequest defines request to retrieve a single execution.
+// **Example**	
+// TODO: add JSON example
```
krhubert: About JSON example I think it's a waste of our time. someone who will use it will know how protobuf is mapped to json.

protobuf/api/execution.proto
antho1404: Nico explained to me about that, it's for 2 things:
- Not sure the generation of the documentation works with comments on top (but seems that it works https://github.com/pseudomuto/protoc-gen-doc#writing-documentation)
- It looks nicer when we have a list of attributes with everything aligned

protobuf/api/execution.proto
antho1404: The json example is good in the documentation, it the proto it is not really useful.
I was thinking about this part could be directly injected in the documentation and we could in the file include the `grpcurl` command to test the api and get the data. This way on the proto we have a clear example of how to use it and when we generate the documentation we could actually execute this command to generate with real inputs/outputs and use that as "integration test"

protobuf/api/execution.proto
krhubert: Still I think it's better to have clean docs (long and explaining in some cases) rather than aligned fields. And remember it's our docs for Public API! So it should be good and here I feel like it was shrink just to keep a small line length.

protobuf/api/execution.proto
krhubert:
```go
package main

import (
  "os"

  "github.com/golang/protobuf/jsonpb"
  "github.com/mesg-foundation/core/protobuf/api"
)

func main() {
  (&jsonpb.Marshaler{EmitDefaults: true, Indent: "\t"}).Marshal(os.Stdout, &api.GetExecutionRequest{})
}
```

Just found the Marshaler, maybe it could be integrated into docs creation pipeline if we want to have it generated.

NicolasMahe: ok i will remove json example

protobuf/api/execution.proto
NicolasMahe: ok guys!

PR #1096, opened 2019-06-24
Sdk dependency rearrangement

antho1404:
dependency: https://github.com/mesg-foundation/engine/pull/1085

A SDK should only have access to its own database (eg: executionsdk have access to executiondb only). All the other data access should go through different sdks (servicesdk, instancesdk)


NicolasMahe:
@antho1404 can you fix conflict

PR #1097, opened 2019-06-24
Fix service API and add comment

NicolasMahe:
Based on https://github.com/mesg-foundation/engine/pull/1092

### DIFF
--- a/sdk/service/service.go
+++ b/sdk/service/service.go
@@ -68,12 +68,11 @@ func (s *Service) Create(srv *service.Service) (*service.Service, error) {
 		return nil, errors.New("service is already deployed")
 	}
 
-	// build service's Docker image and apply to service.
-	imageHash, err := s.container.Build(path)
+	// build service's Docker image.
+	_, err = s.container.Build(path)
 	if err != nil {
 		return nil, err
 	}
-	srv.Configuration.Image = imageHash
 	// TODO: the following test should be moved in New function
 	if srv.Sid == "" {
 		// make sure that sid doesn't have the same length with id.
--- a/server/grpc/api/mapping.go
+++ b/server/grpc/api/mapping.go
@@ -1,7 +1,6 @@
 package api
 
 import (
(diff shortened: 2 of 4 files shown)


server/grpc/api/mapping.go
```diff
@@ -1,21 +1,14 @@
...
 // FromProtoService converts a the protobuf definition to the internal service struct
 // TODO: should not be public. Need to move server/grpc/service.go to server/grpc/api/service.go
 func FromProtoService(s *definition.Service) (*service.Service, error) {
-	hash, err := hash.Decode(s.Hash)
-	if err != nil {
-		return nil, err
-	}
-
 	return &service.Service{
-		Hash:          hash,
```
NicolasMahe: I remove the mapping of the hash here because it's only use in the api service.create that accept service that should NOT have any hash.

sdk/service/service.go
```diff
@@ -70,12 +70,11 @@ func (s *Service) Create(srv *service.Service) (*service.Service, error) {
 		return nil, errors.New("service is already deployed")
 	}
 
-	// build service's Docker image and apply to service.
-	imageHash, err := s.container.Build(path)
+	// build service's Docker image.
+	_, err = s.container.Build(path)
```
NicolasMahe: We are still building the image for test purpose but not saving the image hash here. It will be saved in instance.create.

server/grpc/api/mapping.go
antho1404: `service.Create` should not take a definition in the first place. It should be definition minus hash and configuration image.

sdk/service/service.go
antho1404: still valid to build to test the validity of the service

NicolasMahe:
> I'm not 100% sure about the example in the file. The only purpose is for the documentation website and it's hard to maintain this kind of documentation where it can be generated based on the commands grpcurl of the api that contains request and response all the time

I remove all example 👍

PR #1098, opened 2019-06-24
Remove xstructhash

### LINKED ISSUE #1093: Use Hash.Dump instead of xstructhash or structhash
- [ ] Function that need to hash a struct should use `Hash.Dump` instead of `xstructhash` or `structhash` (but keep `structhash` under the hood)
- [ ] `xstructhash` should be removed
- [ ] Also, it would make sense that the `Dump` function return `Digest` instead of `Hash`, so more data can be added to the `Digest`. @krhubert what do you think?

krhubert:
close #1093


sdk/instance/helpers.go
```diff
@@ -77,9 +76,9 @@ func extractVolumesFrom(s *service.Service, d *service.Dependency) ([]container.
...
-	return hex.EncodeToString(xstructhash.Hash([]string{
-		s.Sid,
-		dependency,
-		volume,
-	}, 1))
+	d := sha1.New()
+	d.Write([]byte(s.Sid))
+	d.Write([]byte(dependency))
+	d.Write([]byte(volume))
+	return hex.EncodeToString(d.Sum(nil))
```
NicolasMahe: Did you try to use Hash.Hash?
A sha256 base58 encoded should be small enough for docker?

sdk/service/service.go
```diff
@@ -57,7 +56,7 @@ func (s *Service) Create(srv *service.Service) (*service.Service, error) {
 
 	// calculate and apply hash to service.
 	dh := dirhash.New(path)
```
NicolasMahe: not blocking: could you also move dirhash in Hash package?

sdk/instance/helpers.go
krhubert: No, I have not tested, but I can try. I thought it would be too much.

krhubert:
```
        "Mounts": [
            {
                "Type": "volume",
                "Name": "8RBsoeyoRwajj86MZfZE6gMDJQVYGYcdSfx1zxqxNHbr",
                "Source": "/var/lib/docker/volumes/8RBsoeyoRwajj86MZfZE6gMDJQVYGYcdSfx1zxqxNHbr/_data",
```
It's working

PR #1099, opened 2019-06-24
Add validation service

### LINKED ISSUE #1088: Add back validation on service struct
We need to add back the validation on the service, this time directly on the service.Service structure instead of the import.Service!

krhubert:
I've added back service validation but there is one pbm:

If we don't have "main" dependency as before how do we address it in `volumeFrom`? [here](https://github.com/mesg-foundation/engine/compare/fix/service-validation?expand=1#diff-fdf7d56c1eb9f4c5fa9f1276efde04e9R78)

related to #1088

### DIFF
--- a/service/service_validate.go
+++ b/service/service_validate.go
@@ -1,0 +1,117 @@
+package service
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/go-playground/locales/en"
+	ut "github.com/go-playground/universal-translator"
+	"github.com/mesg-foundation/core/x/xerrors"
+	"github.com/mesg-foundation/core/x/xvalidator"
+	validator "gopkg.in/go-playground/validator.v9"
+	en_translations "gopkg.in/go-playground/validator.v9/translations/en"
+)
+
+const namespacePrefix = "service."
+
+var validate, translator = newValidator()
+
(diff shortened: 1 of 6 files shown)


service/service_validate.go
```diff
@@ -0,0 +1,117 @@
...
+	for _, dep := range service.Dependencies {
+		for _, depVolumeKey := range dep.VolumesFrom {
+			found := false
+			for _, s := range service.Dependencies {
+				if s.Key == depVolumeKey {
+					found = true
+					break
+				}
+			}
+			if !found && depVolumeKey != "main" {
```
NicolasMahe: why "main"? it should be "service", no? There is a const somewhere that define it.

NicolasMahe:
> If we don't have "main" dependency as before how do we address it in `volumeFrom`? [here](https://github.com/mesg-foundation/engine/compare/fix/service-validation?expand=1#diff-fdf7d56c1eb9f4c5fa9f1276efde04e9R78)

I'm not sure to exactly understand, but: if there is no volume defined in the configuration ("main" dependency), then not volume will be mounted when use the `volumeFrom`.
`volumeFrom` only tell which dependency's volumes to mount from, not explicitly the volume themself. So if not volume are specified, then `volumeFrom` will mount 0 volume. No error.

PR #1100, opened 2019-06-25
[DRAFT] Autogenerate grpc requests and responses for Service API

NicolasMahe:
Dependant on https://github.com/mesg-foundation/engine/pull/1097

Here is a first draft of using grpcurl to auto generate the grpc docs.

The only required file is `service-definition.json`. All other files are generated by the script.

@mesg-foundation/mesg-core-team tell me what you think ;)


antho1404:
I like it, I just would not call it integration test but maybe manual test or something like that also this might make more in a wiki maybe not sure in the codebase

krhubert:
For me integration test is really e2e test, so such dir should be created.

Also we shouldn't use bash for e2e testing, but rather go protobuf client.

NicolasMahe:
closing this PR because it's outdated

PR #1102, opened 2019-06-25
Fix instance start API

NicolasMahe:
Based on https://github.com/mesg-foundation/engine/pull/1097


sdk/instance/instance.go
```diff
@@ -102,16 +106,17 @@ func (i *Instance) Create(serviceHash hash.Hash, env []string) (*instance.Instan
...
 	}
-	if err := i.instanceDB.Save(o); err != nil {
+	if err := i.instanceDB.Save(inst); err != nil {
 		return nil, err
 	}
 
-	_, err = i.start(o, xos.EnvMapToSlice(instanceEnv))
-	return o, err
+	_, err = i.start(inst, imageHash, xos.EnvMapToSlice(instanceEnv))
+	fmt.Println("err start", err)
```
```suggestion
```

server/grpc/api/mapping.go
```diff
@@ -1,13 +1,18 @@
...
+	"fmt"
+
 	"github.com/mesg-foundation/core/protobuf/definition"
 	"github.com/mesg-foundation/core/service"
 )
 
 // FromProtoService converts a the protobuf definition to the internal service struct
 // TODO: should not be public. Need to move server/grpc/service.go to server/grpc/api/service.go
 func FromProtoService(s *definition.Service) (*service.Service, error) {
+	if s == nil {
```
krhubert: Do not check for nil, as it only hides some real issue with nil pointers. It should just do conversion without returning an error.

server/grpc/api/mapping.go
```diff
@@ -175,7 +182,7 @@ func toProtoParameters(params []*service.Parameter) []*definition.Parameter {
 
 func toProtoConfiguration(configuration *service.Dependency) *definition.Configuration {
 	if configuration == nil {
```
krhubert: same here I guess, it shouldn't' check for nil, as this function is for mapping, not nic catching :)

server/grpc/api/mapping.go
NicolasMahe: This check should exist somewhere. I guess it could be move in the `service.Create` api function to keep the mapping simple 👍

server/grpc/api/mapping.go
NicolasMahe: Same answer, this have to be managed somewhere but i agree that it's not the best place here.

server/grpc/api/mapping.go
```diff
@@ -65,7 +65,9 @@ func fromProtoParameters(params []*definition.Parameter) []*service.Parameter {
 
 func fromProtoConfiguration(configuration *definition.Configuration) *service.Dependency {
 	if configuration == nil {
-		return nil
+		return &service.Dependency{
```
NicolasMahe: This will be updated in a future PR that i'm working one ;) so please don't block this PR because of this one.

NicolasMahe: PR in question: https://github.com/mesg-foundation/engine/pull/1110

PR #1109, opened 2019-06-25
Fix service connection with instance

antho1404:
- [x] fix issue with emit event api
- [x] fix issue with listen task api
- [x] fix issue with execute task api


server/grpc/core/core.go
```diff
@@ -70,6 +71,15 @@ func (s *Server) ExecuteTask(ctx context.Context, request *coreapi.ExecuteTaskRe
 	if err != nil {
 		return nil, err
 	}
+
+	instances, err := s.sdk.Instance.List(&instancesdk.Filter{ServiceHash: hash})
+	if err != nil {
+		return nil, err
+	}
+	if len(instances) != 1 {
```
krhubert: Is this check necessary? Is there be a situation that two or more instances will be deployed on the same host?

PR #1110, opened 2019-06-25
Add more validation to service struct

krhubert:
close https://github.com/mesg-foundation/engine/issues/1088

### DIFF
--- a/service/parameter.go
+++ b/service/parameter.go
@@ -22,5 +22,5 @@ type Parameter struct {
 	Repeated bool `hash:"name:6"`
 
 	// Definition of the structure of the object when the type is object
-	Object []*Parameter `hash:"name:7"`
+	Object []*Parameter `hash:"name:7" validate:"unique,dive,required"`
 }
--- a/service/service_validate.go
+++ b/service/service_validate.go
@@ -35,6 +35,10 @@ func ValidateService(service *Service) error {
 		}
 	}
 
+	if err := isServiceKeysUnique(service); err != nil {
+		errs = append(errs, err)
+	}
+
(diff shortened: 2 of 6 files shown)


service/parameter.go
```diff
@@ -22,5 +22,5 @@ type Parameter struct {
 	Repeated bool `hash:"name:6"`
 
 	// Definition of the structure of the object when the type is object
-	Object []*Parameter `hash:"name:7"`
+	Object []*Parameter `hash:"name:7" validate:"unique,dive,required"`
```
NicolasMahe: why this one is unique?
does it actually works?

service/service_validate.go
```diff
@@ -82,6 +86,59 @@ func ValidateService(service *Service) error {
...
+	exist = make(map[string]bool)
+	for _, event := range service.Events {
+		if exist[event.Key] {
+			errs = append(errs, fmt.Errorf("events[%s] already exist", event.Key))
+		}
+		exist[event.Key] = true
+
+		existparam := make(map[string]bool)
+		for _, param := range event.Data {
+			if existparam[param.Key] {
```
NicolasMahe: params.Object keys are not tested (task inputs, task outputs and event)
maybe we could use the validation lib with something like:
```
Object []*Parameter `hash:"name:7" validate:"dive,(key)unique,required"`
```

NicolasMahe: otherwise a new custom validation like `PortMapping`

service/parameter.go
krhubert: it won't be used here, but shouldn't exist

PR #1112, opened 2019-06-26
Add service validation

NicolasMahe:
- calls `service.ValidateService` in `service.Create` API
- simplify the function `fromProtoConfiguration` by returning `nil` if config is `nil`
- add required validation on `service.Configuration`
- split function `ValidateService` to return error if the service structure validation returns an error


server/grpc/api/service.go
```diff
@@ -24,6 +25,9 @@ func NewServiceServer(sdk *sdk.SDK) *ServiceServer {
 // Create creates a new service from definition.
 func (s *ServiceServer) Create(ctx context.Context, request *protobuf_api.CreateServiceRequest) (*protobuf_api.CreateServiceResponse, error) {
 	definition := fromProtoService(request.Definition)
+	if err := service.ValidateService(definition); err != nil {
```
NicolasMahe: Should the Engine calls `service.ValidateService` in api package or directly in sdk package?

NicolasMahe: PR https://github.com/mesg-foundation/engine/pull/1113 add the validation in the serviceSDK.create function but only before the save, so it doesn't actually check the inputs before reading them.
So both make sense. Let's improve this later.

service/service_validate.go
```diff
@@ -34,6 +34,9 @@ func ValidateService(service *Service) error {
 			errs = append(errs, fmt.Errorf("%s%s", namespace, e.Translate(translator)))
 		}
 	}
+	if errs.ErrorOrNil() != nil {
```
antho1404: why this pre-exit?
The rest of the code is appending the errors so we don't need to exit here, that sounds like an early optimization or (and I think it's the case) to exit when the struct has an error (like config missing).
In that case maybe this function should only return an error and exit on the first error that would make things way easier

server/grpc/api/service.go
krhubert: In sdk package. The server package should handle validation on `definition.Service` level , validation for request and response transformation

service/service_validate.go
NicolasMahe: because the rest of the tests will panic as they access pointer that could be nil.
Check line 46: https://github.com/mesg-foundation/engine/pull/1112/files/70e9a6664e42aeab17e1687f2357506ea0d2a76c#diff-fdf7d56c1eb9f4c5fa9f1276efde04e9R46

NicolasMahe: This function is actually doing too many things:
1. check the integrity of the struct
2. check stuff on the data of the struct

if #1 has error, it should return them. that what this line is doing

NicolasMahe: I splitted the function `ValidateService` in 2 functions, one to check the struct integrity, the other one to check the data.

server/grpc/api/service.go
NicolasMahe: I removed the validation from here

NicolasMahe:
@mesg-foundation/mesg-core-team ready to review again

PR #1120, opened 2019-06-26
Split database error to service and instance error

NicolasMahe:
Dependant on https://github.com/mesg-foundation/engine/pull/1106

- Split database error to service and instance error


database/instance_db.go
```diff
@@ -120,7 +126,31 @@ func (d *LevelDBInstanceDB) Close() error {
...
+type ErrInstanceNotFound struct {
+	hash hash.Hash
+}
+
+func (e *ErrInstanceNotFound) Error() string {
+	return fmt.Sprintf("database: instance %q not found", e.hash)
+}
+
+// InstanceDecodeError represents a resource impossible to decode.
+type InstanceDecodeError struct {
```
krhubert: Do not create dedicated errors unless you want to use them.

Look comment below on decode error as well.

database/service_db.go
```diff
@@ -73,7 +72,7 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
 		if err != nil {
 			// NOTE: Ignore all decode errors (possibly due to a service
 			// structure change or database corruption)
-			if decodeErr, ok := err.(*DecodeError); ok {
+			if decodeErr, ok := err.(*ServiceDecodeError); ok {
```
krhubert: looking at the code this error will always be a type of DecodeError.

database/instance_db.go
```diff
@@ -120,7 +126,31 @@ func (d *LevelDBInstanceDB) Close() error {
...
 }
 
-// Delete deletes service from database.
+// Delete deletes an instance from database.
 func (d *LevelDBInstanceDB) Delete(hash hash.Hash) error {
 	return d.db.Delete(hash, nil)
 }
+
+// ErrInstanceNotFound is an not found error.
+type ErrInstanceNotFound struct {
```
krhubert: What is the reason for splitting this and extending api? In that case we end up with err not found for every type.

can't we merge those errors into one like:

```go
type ErrNotFound struct {
  hash hash.Hash
  label string
}
```

func (e *ErrNotFound) Error() string {
  return fmt.Sprintf("database: %s %q not found", e.label, e.hash)
}
```

NicolasMahe: That's what I start doing it first but then I revert to create new types..

NicolasMahe:
@krhubert could you fix this PR or create a new one that improve the database error? I need to work on something else.

database/instance_db.go
antho1404: not sure we even need a label, we should have it based on the context and the api we call

krhubert:
@NicolasMahe I want to finish api (event/logs) first, then I will work on this

database/instance_db.go
NicolasMahe: the context of the apu is not enough, a call to instance APIs can fetch a service and thus return an instance error (because of context) instead of a service.

krhubert:
closing in favor of https://github.com/mesg-foundation/engine/pull/1129

PR #1122, opened 2019-06-26
Use instance hash in event

krhubert:


event/event.go
```diff
@@ -1,28 +1,21 @@
...
+	"github.com/mesg-foundation/core/hash"
 )
 
 // Event stores all informations about Events.
 type Event struct {
-	Service   *service.Service
-	Key       string
-	Data      map[string]interface{}
-	CreatedAt time.Time
+	InstanseHash hash.Hash
```
NicolasMahe:
```suggestion
	InstanceHash hash.Hash
```
typo!

krhubert:
This will be done in next PR when event api will be created

PR #1124, opened 2019-06-26
Make execution filter optional

### LINKED ISSUE #1123: All filters on gRPC APIs should be optional
All filter on the APIs should be optional and work with default values.
If they are not set, then any value should match.
This is the list of all filters, some may already be optional (if it the case, tick the box)
- [ ] `StreamExecutionRequest.Filter.status`. default should match any status
- [ ] `StreamExecutionRequest.Filter.instanceHash`
- [x] `ListInstancesRequest.serviceHash`

krhubert:
close #1123


server/grpc/api/execution.go
```diff
@@ -44,13 +44,19 @@ func (s *ExecutionServer) Get(ctx context.Context, req *api.GetExecutionRequest)
...
-	if err != nil {
+	if req.Filter.InstanceHash != "" && err != nil {
 		return err
 	}
 
+	var statuses []execution.Status
+	if req.Filter.Status != definition.Status_Unknown {
+		statuses = []execution.Status{execution.Status(req.Filter.Status)}
+	}
+	// []execution.Status{execution.Status(req.Filter.Status)},
```
NicolasMahe:
```suggestion
```

PR #1126, opened 2019-06-26
Create event api

krhubert:


protobuf/api/event.proto
```diff
@@ -0,0 +1,49 @@
...
+    // key is the key of the event.
+    string key = 3;
+  }
+
+  // Filter used to filter a stream of events.
+  Filter filter = 1;
+}
+
+// CreateEventRequest defines request for execution update.
+message CreateEventRequest {
```
NicolasMahe: you could change the message `CreateEventRequest` to:
```
message CreateEventRequest {
  // instanceHash is hash of instance that can proceed an execution.
  string instanceHash = 1;

   // key is the key of the event.
  string key = 2;

   // data is the data for the event.
  string data = 3;
}
```
So there is not need to check for a not empty hash in `server/grpc/api/event.go`

Also I think it's more logic to have different messages to create the resource and to return it.
The same modification should be done (but in a future time) on ServiceX/Create. Reusing the definition.service causes more problem than it solves.

server/grpc/api/event.go
```diff
@@ -0,0 +1,108 @@
...
+	return &EventServer{sdk: sdk}
+}
+
+// Create creates a new event.
+func (s *EventServer) Create(ctx context.Context, req *api.CreateEventRequest) (*api.CreateEventResponse, error) {
+	if req.Event.Hash != "" {
+		return nil, errors.New("create event: hash not allowed")
+	}
+
+	if req.Event.Key != "" {
```
NicolasMahe:
```suggestion
	if req.Event.Key == "" {
```

sdk/event/event.go
```diff
@@ -31,26 +31,25 @@ func New(ps *pubsub.PubSub, service *servicesdk.Service, instance *instancesdk.I
 }
 
 // Emit emits a MESG event eventKey with eventData for service token.
-func (e *Event) Emit(instanceHash hash.Hash, eventKey string, eventData map[string]interface{}) error {
+func (e *Event) Emit(instanceHash hash.Hash, eventKey string, eventData map[string]interface{}) (*event.Event, error) {
```
NicolasMahe:
```suggestion
func (e *Event) Emit(evnt *event.Event) (error) {
```
what about change the parameters of this function to directly accept the event struct?
I think it will make this function a bit simpler as well as the api package.

protobuf/api/event.proto
```diff
@@ -0,0 +1,49 @@
...
+  // The event definition to use to create the Event.
+  definition.Event event = 1;
+}
+
+// CreateEventResponse defines response for execution update.
+message CreateEventResponse {
+  // Hash represents execution.
```
antho1404:
```suggestion
  // Hash represents event.
```

protobuf/api/event.proto
```diff
@@ -0,0 +1,49 @@
...
+service Event {
+  // Create creates event with data.
+  rpc Create(CreateEventRequest) returns (CreateEventResponse) {}
+
+  // Stream returns a stream of events that satisfy criteria
+  // specified in a request.
+  rpc Stream(StreamEventRequest) returns (stream definition.Event) {}
+}
+
+// StreamEventRequest defines request to retrieve a stream of executions.
```
antho1404:
```suggestion
// StreamEventRequest defines request to retrieve a stream of events.
```

sdk/event/event.go
antho1404: It's fine like that, otherwise, we will still have the confusion between the actual inputs and the resource itself which is not good. We will have the same problem that we have with the service.create

protobuf/api/event.proto
krhubert:
```
// The request's data for the `Create` API.
  definition.Service definition = 1; // The service definition to use to create the Service.
}
```

We accept service in servcie api, so I wanted to be consistent. And I prefer to check if hash is empty rather then redefine all event fields (same for `CreateServiceRequest`). It's more clear that CreateEventRquest expects event as argument.

sdk/event/event.go
krhubert: I'm ok with both - just another thing we can make consistent accross source code.

protobuf/api/event.proto
NicolasMahe: The following issue is actually to change `CreateServiceRequest`: https://github.com/mesg-foundation/engine/issues/1111

sdk/event/event.go
NicolasMahe: The function create in serviceSDK accept directly the service:
```
func (s *Service) Create(srv *service.Service) (*service.Service, error) {
```

krhubert: Can I upadte this when I'm going to remove coreapi methods?

PR #1128, opened 2019-06-27
clean proto service

krhubert:


server/grpc/api/event.go
```diff
@@ -44,9 +44,8 @@ func (s *EventServer) Create(ctx context.Context, req *api.CreateEventRequest) (
 	if err := json.Unmarshal([]byte(req.Event.Data), &data); err != nil {
 		return nil, fmt.Errorf("create event: data %s", err)
 	}
-
-	event, err := s.sdk.Event.Emit(instanceHash, req.Event.Key, data)
-	if err != nil {
+	event := event.Create(instanceHash, req.Event.Key, data)
+	if err := s.sdk.Event.Create(event); err != nil {
```
antho1404: This is weird stuff and one of the reasons why I don't like this pattern to pass the resource when we actually want to create the resource. We call 2 `event.create` this is so confusing

sdk/event/event.go
```diff
@@ -31,25 +29,23 @@ func New(ps *pubsub.PubSub, service *servicesdk.Service, instance *instancesdk.I
 }
 
 // Emit emits a MESG event eventKey with eventData for service token.
-func (e *Event) Emit(instanceHash hash.Hash, eventKey string, eventData map[string]interface{}) (*event.Event, error) {
-	instance, err := e.instance.Get(instanceHash)
+func (e *Event) Create(event *event.Event) error {
+	instance, err := e.instance.Get(event.InstanceHash)
```
antho1404: I don't really like this pattern where we send an initialized data to create this data... 
Any create should have the following signature `Create(inputs WhateverStructOrListOfInputs) (ResourceType, error)`

protobuf/coreapi/api.proto
```diff
@@ -13,78 +13,13 @@ option go_package = "github.com/mesg-foundation/core/protobuf/coreapi";
 //
 // The source file of this API is hosted on [GitHub](https://github.com/mesg-foundation/core/blob/master/protobuf/coreapi/api.proto).
 service Core {
-  // Subscribe to a stream that listens for events from a service.
-  rpc ListenEvent (ListenEventRequest) returns (stream EventData) {}
-
-  // Execute a service's task through Core.
-  rpc ExecuteTask (ExecuteTaskRequest) returns (ExecuteTaskReply) {}
```
antho1404: This API needs to stay, for now, we don't have any binding for that one

krhubert: I think this api should be moved to execution where it belongs

PR #1129, opened 2019-06-27
Add resource type to database error + remove decode error

krhubert:


database/service_db.go
```diff
@@ -69,16 +68,13 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
...
 		if err != nil {
 			// NOTE: Ignore all decode errors (possibly due to a service
 			// structure change or database corruption)
-			if decodeErr, ok := err.(*DecodeError); ok {
-				logrus.WithField("service", decodeErr.hash.String()).Warning(decodeErr.Error())
-				continue
-			}
-			iter.Release()
-			return nil, err
+			logrus.WithField("service", hash.String()).Warning(err.Error())
```
antho1404: This is a change of behavior and now all the unmarshal error are just as a warning and will never be returned.
Is there any reason for this change of behavior?

PR #1131, opened 2019-06-27
Rename to engine in docs and tools

krhubert:


dev
```diff
@@ -12,10 +12,10 @@ MESG_LOG_FORCECOLORS=${MESG_LOG_FORCECOLORS:-false}
 MESG_LOG_LEVEL=${MESG_LOG_LEVEL:-debug}
 
 VERSION=local
-LDFLAGS="-X 'github.com/mesg-foundation/core/version.Version=$VERSION'"
-LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.EnvMarketplaceEndpoint=https://ropsten.infura.io/v3/6115ae2531f64c04a9a392cf500e5fbe'"
-LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.EnvMarketplaceAddress=0xeCC1A867F871323350A1A89FcAf69629a2d5085e'"
-LDFLAGS+=" -X 'github.com/mesg-foundation/core/config.EnvMarketplaceToken=0x5861b3dc52339d4f976b7fa5d80db6cd6f477f1b'"
+LDFLAGS="-X 'github.com/mesg-foundation/engine/version.Version=$VERSION'"
```
antho1404: use core here, this will probably not work

go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/mesg-foundation/core
+module github.com/mesg-foundation/engine
```
antho1404: This breaks everything because it's not in sync with what Github has or something like that, @NicolasMahe can explain more but I this that's why everything fails 
```suggestion
module github.com/mesg-foundation/core
```

scripts/build-engine.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -e
+
+MESG_SERVICE_SERVER=ipfs.app.mesg.com
+
+LDFLAGS="-X 'github.com/mesg-foundation/engine/version.Version=$version'"
```
antho1404: core here too

dev
krhubert: how it won't work? can you  explain?

go.mod
krhubert: @NicolasMahe how it brakes,  what is not sync I don't get it.

NicolasMahe: I will check this afternoon ;)

NicolasMahe: it was because the `go.mod` in github.com/mesg-foundation/engine is still with `module github.com/mesg-foundation/core`.
So `go mod install` was returning the error `go.mod path is different than source` or something like this.

NicolasMahe: I guess same problem as docker face by renaming to moby. All the go sources are still under docker repo to not break previous version i think

NicolasMahe: I will check in depth in a few hours

krhubert: oh ok, i guess import names was never changed, i'll update this and check

go.mod
```diff
@@ -31,19 +26,15 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware v0.0.0-20180522105215-e9c5d9645c43
 	github.com/huandu/xstrings v1.2.0 // indirect
 	github.com/imdario/mergo v0.3.7 // indirect
-	github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99 // indirect
 	github.com/kelseyhightower/envconfig v1.3.0
 	github.com/kevinburke/ssh_config v0.0.0-20180830205328-81db2a75821e // indirect
-	github.com/leodido/go-urn v1.1.0 // indirect
 	github.com/lyft/protoc-gen-validate v0.1.0 // indirect
+	github.com/mesg-foundation/core v0.10.1
```
NicolasMahe: dw

krhubert: sorry

NicolasMahe:
I vote to merge this PR even if it breaking the build on go software that depend on the Engine.
The earlier we do this update, the less software/devs are impacted.

antho1404:
ok for me too to merge, please fix conflicts first and merge when tests are green and let's cross fingers 🤞

PR #1140, opened 2019-06-28
Delete Log service

krhubert:
Dep #1131 

done:
- api logs removal
- move config.Name to version constance
- change docker namespace parameter from string slice to string

The last two was started because of new logs api which was dropped. Thus they are still a good thing because they simplify code and docker api (I don't know why namespace was slice in the first place)


config/config.go
```diff
@@ -27,7 +27,6 @@ var (
 
 // Config contains all the configuration needed.
 type Config struct {
-	Name string
```
NicolasMahe: Don't we need this to allow multiple Engine to run on the same docker?

krhubert: Nope, we pass this to service, but in fact, we used $MESG_NAME to have a different docker service name. This one was constant all the time.

krhubert: Retested, working fine!

NicolasMahe: What about the address of the Engine? It is used to create the ENDPOINT env in the services.

NicolasMahe:
This is a huge PR...
Could you move "move config.Name to version constance" and "change docker namespace parameter from string slice to string" to other PR even if they depend on each other.

NicolasMahe:
Slice of namespace were used to be able to create namespace composed of many subnamespace and let the docker package handler the concatenation. Maybe it's was over engineered and because we are now hashing them it's not needed anymore.

krhubert:
Please merge #1131 before then the PR has only

```
 35 files changed, 190 insertions(+), 841 deletions(-)
```

where 585 lines were removed in 0417c62b26e8c71aff6e3495628594b57bd4d54d and next 81 lines in 0417c62b26e8c71aff6e3495628594b57bd4d54d

So this PR is rather small because it has a lot of deletion old api, but we need to merge #1131 before.

There will be more work with splitting this PR then view it commit by commit. I can help with reviewing if you want.

config/config.go
antho1404: this should not be removed, this is the part that is inserted in the docker service. Now based on the rest of the pull request you inject this `version.Name` but there is no way to change this name where on the config we can change it with `MESG_NAME` (`envconfig.MustProcess(envPrefix, c)`)

With this change we cannot run multiple engine on the same computer. Of course run multiple instances will work but when we will start services then we will have a problem. I don't see any reason to change that behavior.

sdk/instance/helpers.go
```diff
@@ -9,14 +10,14 @@ import (
 	"github.com/mesg-foundation/engine/service"
 )
 
-// instanceNamespace returns the namespace of the service.
-func instanceNamespace(hash hash.Hash) []string {
-	return []string{hash.String()}
+// InstanceNamespace returns the namespace of the service.
+func InstanceNamespace(hash fmt.Stringer) string {
```
antho1404: why this is exposed? maybe I'm missing something but I don't see any external package that uses it

config/config.go
krhubert: This makes sens , i've reverted the commit + update a branch

sdk/instance/helpers.go
krhubert: It was done for service logs, reverting.

PR #1141, opened 2019-07-01
Update create service&event proto request data layout.

### LINKED ISSUE #1111: Update protobuf struct CreateXXXRequest
Currently the protobuf struct `CreateServiceRequest` is accepting a `definition.Service` as first parameter.
This is not the best as the data send to `CreateServiceRequest` are not exactly a `definition.Service`, the hash should not be present as it is calculated by the Engine, it makes the API a bit confusing.
Also, we cannot pass directly the output of the command `mesg-cli service:compile` as it needs to be inside the key `definition` in `CreateServiceRequest`.

My suggestion is to update `CreateServiceRequest` to:
```proto
  string sid = 1;                                      // Service's sid.
  string name = 2;                                     // Service's name.
  string description = 3;                              // Service's description.
  definition.Configuration configuration = 4;          // Configurations related to the service
  repeated definition.Task tasks = 5;                  // The list of tasks this service can execute.
  repeated definition.Event events = 6;                // The list of events this service can emit.
  repeated definition.Dependency dependencies = 7;     // The container dependencies this service requires.
  string repository = 8;                               // Service's repository url.
  string source = 9;                                   // The hash id of service's source code on IPFS.
}
```

Make sure that other `CreateXXXRequest` messages are following the same pattern.

krhubert:
close #1111


server/grpc/api/service.go
```diff
@@ -22,8 +22,18 @@ func NewServiceServer(sdk *sdk.SDK) *ServiceServer {
 }
 
 // Create creates a new service from definition.
-func (s *ServiceServer) Create(ctx context.Context, request *protobuf_api.CreateServiceRequest) (*protobuf_api.CreateServiceResponse, error) {
-	definition := fromProtoService(request.Definition)
+func (s *ServiceServer) Create(ctx context.Context, req *protobuf_api.CreateServiceRequest) (*protobuf_api.CreateServiceResponse, error) {
+	definition := fromProtoService(&types.Service{
```
NicolasMahe: Why not directly updating `fromProtoService` to accept the struct `CreateServiceResponse`?
`fromProtoService` is only used here. You could also renamed it to something more explicit.

NicolasMahe: Actually, what did we say about sdk also? To creates a new struct if there are too many parameters, no?
`CreateServiceParameters`?

krhubert: I feel like passing one struct is easier here. I understand why api requirement was to pass exact data that create (or any other method) expect, but internally `Service` struct could be used. 

We can merge this to finish stabilize api and think how we want to approach this pbm.

NicolasMahe: > We can merge this to finish stabilize api and think how we want to approach this pbm.
👍

NicolasMahe: Issue created: https://github.com/mesg-foundation/engine/issues/1147

PR #1144, opened 2019-07-01
Rename event id to event hash

### LINKED ISSUE #1142: Rename eventId to eventHash in Execution
The `eventID` parameter should be renamed by `eventHash` in the Execution API and  also in Execution struct. The type should also be changed to `hash.Hash`.

krhubert:
close #1142


hash/hash_test.go
```diff
@@ -25,6 +27,27 @@ func TestInt(t *testing.T) {
 	assert.Len(t, Int(1), size)
 }
 
+func TestRnadom(t *testing.T) {
```
NicolasMahe:
```suggestion
func TestRandom(t *testing.T) {
```

hash/hash.go
```diff
@@ -58,6 +59,21 @@ func Int(h int) Hash {
 	return hash
 }
 
+// Random returns a new random hash.
+func Random() (Hash, error) {
```
NicolasMahe: Should use this function is way more test ;)

execution/execution.go
```diff
@@ -36,7 +36,7 @@ func (s Status) String() (r string) {
 type Execution struct {
 	Hash         hash.Hash              `hash:"-"`
 	ParentHash   hash.Hash              `hash:"name:parentHash"`
-	EventID      string                 `hash:"name:eventID"`
+	EventHash    hash.Hash              `hash:"name:eventID"`
```
NicolasMahe:
```suggestion
	EventHash    hash.Hash              `hash:"name:eventHash"`
```

hash/hash.go
krhubert: Int is fine :P

hash/hash_test.go
```diff
@@ -25,6 +27,27 @@ func TestInt(t *testing.T) {
...
+			return false
+		}
+
+		if seen[hash.String()] {
+			return false
+		}
+		seen[hash.String()] = true
+		return true
+	}
+	require.NoError(t, quick.Check(f, nil))
```
antho1404: didn't know about this quick method 👍

PR #1148, opened 2019-07-01
Update eth wallet to use new client

krhubert:
~~Waiting for:~~

~~https://github.com/mesg-foundation/engine/pull/1146~~

Client has:

- access to all apis via grpc
- (Un)Marshal functions to not exposed internal input format
- execution runner that is a nice wrapper that could be used by services 

TODO:

- [ ] Test and find a better name for ExecutionRunner and ExeuctionRun.

Problems:

The client has to support both inputs and errors (same for the services). Now service only support valid inputs, but with workflow, an error output must be supported as well. We need to think of good client api for that :

```go
func (input []byte, errOut error) ([]byte, error)
```

I don't like it, maybe some type, or enum, don't know.


client/execution_runner.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func (r *ExecutionRunner) Add(name string, fn ExecutionRun) {
+	r.defs[name] = fn
+}
+
+func (r *ExecutionRunner) Run() error {
+	stream, err := r.client.ExecutionClient.Stream(context.Background(), &pb.StreamExecutionRequest{
+		Filter: &pb.StreamExecutionRequest_Filter{
+			InstanceHash: r.client.InstanceHash,
```
antho1404: as a service, you should only listen to execution with the status in-progress

NicolasMahe:
> * [ ]  Test and find a better name for ExecutionRunner and ExeuctionRun.

typo

krhubert:
close #1157

NicolasMahe:
Tested with CLI 👍

PR #1151, opened 2019-07-02
Change namespace from string slice to string

krhubert:
- Use namespace prefix in container instead of accessing config.
- Change namespace from string slice to string

closes https://github.com/mesg-foundation/engine/issues/495


antho1404:
let me time to test some edge cases on that one before merge

PR #1163, opened 2019-07-03
Do not require service configuration

krhubert:


antho1404:
Why is this a hight priority? I don't understand what problem does it solve. Please provide details of the reason of this change

krhubert:
```
./bin/run service:dev ../test-service
Create service... !
Error: 2 UNKNOWN: Service.Configuration is a required field
```

for given mesg.yaml.

```
name: "test-service"
events:
  foo:
    data:
      bar:
        type: String
```

So, the configuration could be left empty.

antho1404:
ok, nice catch.

The validation on the data should stay and the creation of the default config should not be on the validation but on the creation of the resource.

Just move this
```
if srv.Configuration == nil {
  srv.Configuration = &service.Dependency{}
}
```
to `sdk/service.go#Create` and that should be the only modification needed

antho1404:
I created another PR that can replace this one https://github.com/mesg-foundation/engine/pull/1165

PR #1172, opened 2019-07-07
Workflow engine foundations

antho1404:
Basic implementation of workflow.
For now, workflows are limited to only event and a single task (final implementation will be event or result and multiple tasks)

Testing:

service.json 
```json
{"sid":"test-workflow","name":"Test workflow","tasks":[{"key":"taskX","inputs":[{"key":"foo","type":"String","object":[]},{"key":"bar","type":"String","object":[]}],"outputs":[{"key":"res","type":"Any","object":[]}]}],"events":[{"key":"eventX","data":[{"key":"foo","type":"String","object":[]},{"key":"bar","type":"String","object":[]}]}],"dependencies":[],"source":"QmQvRzJPFDhyBGK2rQP5mAeMrgp1XsTB8WYK2c7FHvyAB8"}
```

```bash
grpcurl -plaintext -d "$(cat service.json)" localhost:50052 api.Service/Create
grpcurl -plaintext -d "{\"serviceHash\":\"C8Go4Dmuc6hJaxQZzii6Jcp8ym2fhynULabKf4VowpCx\"}" localhost:50052 api.Instance/Create
grpcurl -plaintext -d "{\"filter\": {\"statuses\": [\"Completed\"]}}" localhost:50052 api.
Execution/Stream
```

🎉 you have new executions created because of the event from the service that is triggered by [this workflow](https://github.com/mesg-foundation/engine/pull/1172/files#diff-a242ceb5eae6259bbc1db81deef03041R81)

### DIFF
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -34,10 +34,10 @@ func (w *Workflow) Start() error {
 		return fmt.Errorf("workflow engine already running")
 	}
 	w.eventStream = w.event.GetStream(nil)
-	for {
-		event := <-w.eventStream.C
+	for event := range w.eventStream.C {
 		go w.processEvent(event)
 	}
+	return nil
 }
 
 func (w *Workflow) processEvent(event *event.Event) {
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -1,0 +1,92 @@
+package workflow
+
+import (
+	"fmt"
+
+	"github.com/mesg-foundation/engine/event"
+	"github.com/mesg-foundation/engine/hash"
+	eventsdk "github.com/mesg-foundation/engine/sdk/event"
+	executionsdk "github.com/mesg-foundation/engine/sdk/execution"
+)
+
+// Workflow exposes functions of the workflow
+type Workflow struct {
+	event       *eventsdk.Event
+	eventStream *eventsdk.Listener
+
+	execution *executionsdk.Execution
+
+	ErrC chan error
+}
+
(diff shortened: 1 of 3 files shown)


workflow/workflow.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/mesg-foundation/engine/event"
+	"github.com/mesg-foundation/engine/hash"
+	eventsdk "github.com/mesg-foundation/engine/sdk/event"
+	executionsdk "github.com/mesg-foundation/engine/sdk/execution"
+)
+
+// Workflow exposes functions of the workflow
+type Workflow struct {
+	event       *eventsdk.Event
+	eventStream *eventsdk.Listener
```
krhubert: eventStream is used only in start method so it can be removed here

workflow/workflow.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+}
+
+// Start the workflow engine
+func (w *Workflow) Start() error {
+	if w.eventStream != nil {
+		return fmt.Errorf("workflow engine already running")
+	}
+	w.eventStream = w.event.GetStream(nil)
+	for {
```
krhubert:
```suggestion
	for event := range w.eventStream.C {
```

+ remove event below

workflow/workflow.go
antho1404: This is used to check that the workflow engine is not started already and we should keep it in order to close the listener when we stop the workflow engine (not yet implemented)

PR #1173, opened 2019-07-07
Update docker labels

krhubert:
close https://github.com/mesg-foundation/engine/issues/1152

> Let's change it with the following:
```
mesg.engine: name of the engine
mesg.hash: hash of the instance
mesg.service: hash of the service
mesg.dependency: name of the dependency
mesg.sid: sid of the service
```

I've change it a bit to be consistent (s/mesg.hash/mesg.instance/)

```
mesg.engine: name of the engine
mesg.instance: hash of the instance
mesg.service: hash of the service
mesg.dependency: name of the dependency
mesg.sid: sid of the service
```

### DIFF
--- a/sdk/instance/start.go
+++ b/sdk/instance/start.go
@@ -35,13 +35,6 @@ func (i *Instance) start(inst *instance.Instance, imageHash string, env []string
 	// After solving this by docker, switch back to deploy in parallel
 	configs := make([]container.ServiceOptions, 0)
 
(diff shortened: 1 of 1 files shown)


antho1404:
let's merge this when we finish all the releases

antho1404:
This has a BREAKING CHANGE, now the CLI needs to update the logs based on the `mesg.instance` flag

PR #1174, opened 2019-07-07
Do not use config inside instance sdk

krhubert:


sdk/instance/instance.go
```diff
@@ -21,14 +22,20 @@ type Instance struct {
...
 	instanceDB database.InstanceDB
+
+	endpoint   string
+	engineName string
 }
 
 // New creates a new Instance SDK with given options.
-func New(c container.Container, service *servicesdk.Service, instanceDB database.InstanceDB) *Instance {
+func New(c container.Container, service *servicesdk.Service, instanceDB database.InstanceDB, endpoint string) *Instance {
+	engineName, _, _ := xnet.SplitHostPort(endpoint)
```
antho1404: I don't really like the idea to decompose the name based on the endpoint. It would be better to have the name and port and recompose the endpoint with that in the instance/start file

sdk/sdk.go
```diff
@@ -19,10 +19,10 @@ type SDK struct {
 }
 
 // New creates a new SDK with given options.
-func New(c container.Container, serviceDB database.ServiceDB, instanceDB database.InstanceDB, execDB database.ExecutionDB) *SDK {
+func New(c container.Container, serviceDB database.ServiceDB, instanceDB database.InstanceDB, execDB database.ExecutionDB, endpoint string) *SDK {
```
antho1404: not blocking but this one start to be big, maybe we could create a struct like `SDKConfig` to pass all this

krhubert: Yes, but this will be changed with cosmossdk

antho1404:
@NicolasMahe was this merged in the last release? Otherwise, we should put it in the next milestone

PR #1175, opened 2019-07-09
Attach event to execution

antho1404:
- Allow SDK to accept event (and not create the even internally)
- Attach the event to executions from the workflow engine

Before merging, we need to have the best solution for the `Execution/Create` API. Every execution should be triggered by an event so the API is creating an event but there are 2 ways to do that, please check the comment below and let me know what you think.

https://github.com/mesg-foundation/engine/blob/082f506a618b56c4e2129d334b87a945833121b1/server/grpc/api/execution.go#L44-L59

### DIFF
--- a/sdk/execution/execution.go
+++ b/sdk/execution/execution.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/cskr/pubsub"
 	"github.com/mesg-foundation/engine/database"
+	"github.com/mesg-foundation/engine/event"
 	"github.com/mesg-foundation/engine/execution"
 	"github.com/mesg-foundation/engine/hash"
 	instancesdk "github.com/mesg-foundation/engine/sdk/instance"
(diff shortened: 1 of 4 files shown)


antho1404:
but executions need to have the reference of the event and this removes the hack that we had about this random event that was here to wait for the workflow to come and populate this event.

A solution for that that we already talked about it to have these system-events and one of them would be this one from the API that skips the validation (or have its own validation). This will be also useful for the event of the end of execution.
But it means that we have specific names of event that are restricted for the user.

krhubert:
So I don't get we can't create an event in SDK like before? 

As I see it there are two solutions:

- put a reference (+getter method) inside `execution` struct and create event while calling `execution.New`
- split `Execute` into `Create` and `Execute` and create should return `Execution` and `Event` attached to it.


For sure creation of the event shouldn't be done in server pkg, or any other package should handle the creation of random event attached to execution.

antho1404:
Merging this one, I did the creation of engine events in this PR https://github.com/mesg-foundation/engine/pull/1178

PR #1178, opened 2019-07-14
Even system

antho1404:
Creation of an "engine event" used for the execution/create api

### DIFF
--- a/event/engine.go
+++ b/event/engine.go
@@ -1,0 +1,28 @@
+package event
+
+import (
+	"github.com/mesg-foundation/engine/hash"
+)
+
+// a 44 length constant hash
+const engineEventHash = "EngineEventEngineEventEngineEventEngineEvent"
+
+// EngineEventType type to describe engine events
+// Engine events doesn't need the same validation that normal events need
+// These events are used internally only and cannot be emitted by services
+type EngineEventType string
+
+const (
+	// EngineAPIExecution This event is triggered when the `execution/create` API is called
+	EngineAPIExecution EngineEventType = "engine-api-execution"
+)
+
+// EngineEvent creates an engine event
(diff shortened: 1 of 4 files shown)


event/engine.go
```diff
@@ -0,0 +1,17 @@
+package event
+
+// EngineEventType type to describe engine events
+// Engine events doesn't need the same validation that normal events need
+// These events are used internally only and cannot be emitted by services
+type EngineEventType string
```
krhubert: I don't like the fact type is prefixed with the project name, I prefer to use `InternalEvent` prefix.

The engine shouldn't limit names for a service event.  These internal events should be published on internal pubsub and shouldn't be mixed with events from services. 

A flag could be added to `Event` + method `IsInternal` rather than use magic event names

NicolasMahe: I agree that the using prefix Engine is bad.
We did the same before with Core and we end up calling everything CoreXXX.

Internal is also not the best as we would like to create a bunch of pre-defined events that also Services could emit (eg: logs or health check).

What about `SystemEvents`?

event/engine.go
```diff
@@ -0,0 +1,17 @@
...
+
+const (
+	// EngineAPIExecution This event is triggered when the `execution/create` API is called
+	EngineAPIExecution EngineEventType = "mesg:engine-api-execution"
+)
+
+// EngineEvent creates an engine event.
+// Engine event does not have any instance hash and are prefixed by mesg:
+func EngineEvent(evtType EngineEventType, data map[string]interface{}) *Event {
```
NicolasMahe: This function should be prefixed with `New`, shouldn't it?
```suggestion
func New EngineEvent(evtType EngineEventType, data map[string]interface{}) *Event {
```

antho1404:
closing in favor of https://github.com/mesg-foundation/engine/pull/1182

PR #1180, opened 2019-07-14
End of execution event

antho1404:
dependency: https://github.com/mesg-foundation/engine/pull/1178

Creates an event at the end of execution with all the details of this execution. This will be useful for the workflow system that can now connect normal events and events from any end of execution.

These events are published on the event stream and can be filtered too like any other events.

### DIFF
--- a/server/grpc/api/execution.go
+++ b/server/grpc/api/execution.go
@@ -41,7 +41,7 @@ func (s *ExecutionServer) Create(ctx context.Context, req *api.CreateExecutionRe
 		return nil, fmt.Errorf("cannot parse execution's inputs (JSON format): %s", err)
 	}
 
-	evt := event.EngineEvent(event.EngineAPIExecution, inputs)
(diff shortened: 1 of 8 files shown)


server/grpc/api/execution.go
```diff
@@ -41,22 +41,7 @@ func (s *ExecutionServer) Create(ctx context.Context, req *api.CreateExecutionRe
...
-		Data:         inputs,
-	}
-	// Or the following but it needs to create a hack to skip the validation
-	// because the event is not present in the service. We could have an additional
-	// parameter `skipValidation` but I'm not a huge fan
-	// evt, err := s.sdk.Event.Create(hash, "api-call", inputs)
-	// if err != nil {
-	// 	return nil, err
-	// }
+	evt := s.sdk.Event.CreateEngineEvent(event.EngineAPIExecution, inputs)
```
krhubert: Based on #1178 here we should have internal pubsub

antho1404:
closing in favor of https://github.com/mesg-foundation/engine/pull/1182

PR #1181, opened 2019-07-14
Workflow filters

antho1404:
Add possibility to filter event data on the workflow.
Workflow can have a list of filters. All filters of a workflow should match to trigger a task.

Filters have 3 attributes:
- Key: the key we want to filter
- Predicate: the predicate we want to use (for now only EQ (equal) is implemented)
- Value: the expected value

**Test:**
Follow the steps in https://github.com/mesg-foundation/engine/pull/1172 but make sure to listen the stream before starting the service.
The service will emit events that auto increment and when the event has the output `world-2` the execution will be created and only at this moment.
The counter is saved in a volume so make sure to stop the instance with the `delete-data` attribute

### DIFF
--- a/workflow/type.go
+++ b/workflow/type.go
@@ -1,9 +1,19 @@
 package workflow
 
-import "github.com/mesg-foundation/engine/hash"
+import (
+	"github.com/mesg-foundation/engine/event"
+	"github.com/mesg-foundation/engine/hash"
+)
 
 // These structs are temporary and will be part of the service definition
 // TODO: Move to service struct
+
+type predicate uint
+
+// Possible status for services.
+const (
+	EQ predicate = iota
+)
 
 type workflow struct {
 	Trigger trigger
@@ -15,8 +25,43 @@ type task struct {
 	TaskKey      string
 }
 
+type filter struct {
+	Key       string
+	Predicate predicate
+	Value     interface{}
+}
+
 // Trigger is an event that triggers a workflow
 type trigger struct {
 	InstanceHash hash.Hash
 	EventKey     string
+	Filters      []*filter
+}
+
+func (t *trigger) Match(evt *event.Event) (bool, error) {
+	instanceHash := evt.InstanceHash
+	if !t.InstanceHash.Equal(instanceHash) {
+		return false, nil
+	}
+
+	if t.EventKey != evt.Key {
+		return false, nil
+	}
+
+	for _, filter := range t.Filters {
+		if !filter.Match(evt.Data) {
+			return false, nil
+		}
+	}
+
+	return true, nil
+}
+
+func (f *filter) Match(inputs map[string]interface{}) bool {
+	switch f.Predicate {
+	case EQ:
+		return inputs[f.Key] == f.Value
+	default:
+		return false
+	}
 }
--- a/workflow/type.go
+++ b/workflow/type.go
@@ -38,23 +38,22 @@ type trigger struct {
 	Filters      []*filter
 }
 
-func (t *trigger) Match(evt *event.Event) (bool, error) {
-	instanceHash := evt.InstanceHash
-	if !t.InstanceHash.Equal(instanceHash) {
-		return false, nil
+func (t *trigger) Match(evt *event.Event) bool {
+	if !t.InstanceHash.Equal(evt.InstanceHash) {
+		return false
 	}
 
 	if t.EventKey != evt.Key {
-		return false, nil
+		return false
 	}
 
 	for _, filter := range t.Filters {
 		if !filter.Match(evt.Data) {
-			return false, nil
+			return false
 		}
 	}
 
-	return true, nil
+	return true
 }
 
 func (f *filter) Match(inputs map[string]interface{}) bool {
--- a/workflow/type_test.go
+++ b/workflow/type_test.go
@@ -1,0 +1,76 @@
+package workflow
+
+import (
+	"testing"
+
+	"github.com/mesg-foundation/engine/event"
+	"github.com/mesg-foundation/engine/hash"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMatch(t *testing.T) {
+	var tests = []struct {
+		trigger *trigger
+		event   *event.Event
+		match   bool
+	}{
+		{ // matching event
+			&trigger{InstanceHash: hash.Int(1), EventKey: "xx"},
+			&event.Event{InstanceHash: hash.Int(1), Key: "xx"},
+			true,
+		},
+		{ // not matching instance
+			&trigger{InstanceHash: hash.Int(1)},
+			&event.Event{InstanceHash: hash.Int(2)},
+			false,
+		},
+		{ // not matching event
+			&trigger{InstanceHash: hash.Int(1), EventKey: "xx"},
+			&event.Event{InstanceHash: hash.Int(1), Key: "yy"},
+			false,
+		},
+		{ // matching filter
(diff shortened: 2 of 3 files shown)


workflow/type_test.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+func TestMatch(t *testing.T) {
+	var tests = []struct {
+		trigger *trigger
+		event   *event.Event
+		match   bool
+	}{
+		{ // matching event
+			&trigger{InstanceHash: hash.Int(1), EventKey: "xx"},
```
krhubert: Just a small fix, but do not use magic names/values etc in test (like xx :)), rather put `-` or `(none)`  for other developers

workflow/workflow.go
```diff
@@ -33,6 +34,11 @@ func (w *Workflow) Start() error {
 	if w.eventStream != nil {
 		return fmt.Errorf("workflow engine already running")
 	}
+	go func() {
+		for err := range w.ErrC {
+			logrus.Error(err)
```
krhubert: Do not log errors in package, `ErrC` is exposed and errors should be logged in the main.go or package for collecting and logging all errors.

workflow/type.go
```diff
@@ -1,10 +1,20 @@
...
-import "github.com/mesg-foundation/engine/hash"
+import (
+	"github.com/mesg-foundation/engine/event"
+	"github.com/mesg-foundation/engine/hash"
+)
 
 // These structs are temporary and will be part of the service definition
 // TODO: Move to service struct
 
+type predicate uint
```
krhubert:
```suggestion
type operator uint
```

and then EQ rename OP_EQ as a starndard notation

workflow/type.go
```diff
@@ -15,8 +25,46 @@ type task struct {
 	TaskKey      string
 }
 
+type filter struct {
+	Key       string
+	Predicate predicate
+	Value     interface{}
```
krhubert: Unsued value, what for?

workflow/type.go
```diff
@@ -15,8 +25,46 @@ type task struct {
...
+}
+
 // Trigger is an event that triggers a workflow
 type trigger struct {
```
krhubert: rename trigger to hum EventTrigger?

workflow/type.go
```diff
@@ -15,8 +25,46 @@ type task struct {
 	TaskKey      string
 }
 
+type filter struct {
```
krhubert: rename to eventFilter

Plus logic to match te events could be implemented directrly in the event package

workflow/type.go
antho1404: `Value` is used in the `Match` function of the filter

workflow/type.go
antho1404: do you mean having something like
```
event.Match(filter)
```
in that case why not but we need to have some kind of standard filter. This filter is really specific to workflows only.

Also for the renaming, let's not worry too much now as this will move.
https://github.com/mesg-foundation/engine/blob/260f7fb35f07fca05a15b7d30ddd834770cb31c2/workflow/type.go#L8-L9

workflow/type.go
antho1404: Is `OP_EQ` really standard? It doesn't follow the camelCase naming rule

workflow/type.go
NicolasMahe: I really like to have a dedicated filter package that event, execution, instance and services could implement ;)

workflow/type.go
NicolasMahe: I vote to keep the name `predicate` and to use explicit names closer to SQL:
```
equals (=)
greater than (>)
less than (<)
not equal to (<>)
less than or equal to (<=)
greater than or equal to (>=)
```

We can keep `operator` for `AND`, `OR` and `NOT` operators than can combine predicates.

http://www.cs.toronto.edu/~nn/csc309-20085/guide/pointbase/docs/html/htmlfiles/dev_searchcondandpredicates.html
https://en.wikipedia.org/wiki/Where_(SQL)
https://docs.microsoft.com/en-us/sql/t-sql/queries/search-condition-transact-sql?view=sql-server-2017

NicolasMahe: or otherwise, something closer to noSQL:

Name | Description
-- | --
$eq | Matches values that are equal to a specified value.
$gt | Matches values that are greater than a specified value.
$gte | Matches values that are greater than or equal to a specified value.
$in | Matches any of the values specified in an array.
$lt | Matches values that are less than a specified value.
$lte | Matches values that are less than or equal to a specified value.
$ne | Matches all values that are not equal to a specified value.
$nin | Matches none of the values specified in an array.

https://docs.mongodb.com/manual/reference/operator/query/

workflow/type_test.go
antho1404: in fact, just "-" or "(none)" doesn't cover all the test cases and it is even harder to read for the complex tests like the ones with the filters.
If it's not blocking I would keep it like that or otherwise, we need at least 4 or 6 magic but conventional values...

workflow/type.go
antho1404: I'm ok to keep operator for AND and OR it makes more sense.

For the naming of the predicates, I like the simplicity of noSQL without the $ so EQ, GT, GTE...

workflow/type.go
antho1404: let's not have this dedicated filter now, also this means that the logic of filtering is inverted.

Instead of `filter.Match(resource)` we have `resource.Match(filter)`. This could have some undesired implications

workflow/type.go
antho1404: will be moved in the service package so let's discuss that at this time

workflow/workflow.go
antho1404: done

workflow/type.go
```diff
@@ -15,8 +25,43 @@ type task struct {
...
+}
+
 // Trigger is an event that triggers a workflow
 type trigger struct {
 	InstanceHash hash.Hash
 	EventKey     string
+	Filters      []*filter
+}
+
+func (t *trigger) Match(evt *event.Event) (bool, error) {
```
NicolasMahe:
```suggestion
func (t *trigger) Match(evt *event.Event) (bool) {
```
No error is returned.

antho1404: good catch, fixed

workflow/type.go
krhubert: > Is OP_EQ really standard? It doesn't follow the camelCase naming rule

I can find more but here is the first example from golang source code
https://github.com/golang/go/blob/master/src/go/token/token.go

If you write some tokenizer/parser/lexer you don't use camelCase and this is one of the exceptions.

> We can keep operator for AND, OR and NOT operators than can combine predicates.

As you post [here](http://www.cs.toronto.edu/~nn/csc309-20085/guide/pointbase/docs/html/htmlfiles/dev_searchcondandpredicates.html)
predicates are more complex, including BETWEEN, LIKE function etc...:

```
PointBase supports the following types of predicates:

comparison (=, <>, <, >, <=, >=, !=)
BETWEEN
LIKE
EXITS | NOT EXISTS
IN | NOT IN
NULL
//....
```

but AND (&), OR (|) EQ(==), GT(>=) etc are all operators and I don't see the point to split AND, OR, NOT operators and EQ, ...

Also for the first implementation of workflow (there is no info about more complex syntax), we need only simple matching and there is no need to overengineering this part of the code at this point. We can start with operators because we need only them by now.

The question is where we want to go next. You want some DSL lang for it, or relay only on configuration syntax.

PR #1182, opened 2019-07-14
Add support of results for workflow triggers

antho1404:
The workflow is now listening to the following streams
- Events
- Executions (filtered by completed executions only)

If it receives a matching even, it creates an execution with the event hash and the event data
If it receives a matching execution, it creates an execution with the parent execution hash and the output of the parent execution

To test:

Current workflow implemented:
https://github.com/mesg-foundation/engine/blob/f35871959db3ed4925f4565b8007e3af66bd1a99/workflow/workflow.go#L95-L103

Service [here](https://github.com/mesg-foundation/engine/blob/f35871959db3ed4925f4565b8007e3af66bd1a99/workflow/workflow.go#L89)

- Create the service
- Listen for executions (do not start the service yet)
- Start the instance (with `--env EVENT_INTERVAL=0` to remove useless logs)

You should see 2 executions, one with the evenHash (event start) and the other with a parentHash (result of taskA)

You can trigger another one with
```mesg-cli service:execute test-workflow taskA --data a=hello```

And see again 2 executions, one from the CLI (with an eventHash) the second from the workflow (with an parentHash)

### DIFF
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -3,7 +3,6 @@ package workflow
 import (
 	"fmt"
 
-	"github.com/mesg-foundation/engine/event"
 	"github.com/mesg-foundation/engine/execution"
 	"github.com/mesg-foundation/engine/hash"
 	eventsdk "github.com/mesg-foundation/engine/sdk/event"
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by NicolasMahe** [comment]
I really think all the Engine should be based on event and it will be simpler to have only one type of data to pass consensus on the network.

File: workflow/workflow.go [hunk 1]
```diff
@@ -30,18 +32,35 @@ func New(event *eventsdk.Event, execution *executionsdk.Execution) *Workflow {
...
+		select {
+		case event := <-w.eventStream.C:
+			go w.processEvent(event, Event)
+		case execution := <-w.executionStream.C:
+			eventHash, err := hash.Random()
+			if err != nil {
+				w.ErrC <- err
+			}
+			event := &event.Event{
+				Hash:         eventHash,
```
NicolasMahe:
```suggestion
				Hash:         execution.Hash,
```

NicolasMahe: Event Hash = Instance Hash + Data (outputs + execution hash)

antho1404:
Based on the last meeting, we continue with this implementation but we simplify it.
- Executions don't need an event, either they have an event hash OR a parent hash
- Removing the event from the exection.Create function and pass inputs directly

PR #1188, opened 2019-07-16
Workflow service

antho1404:
This PR is moving the temporary types from the workflow package to the service package and update the service type/api to accept/expose the information of the workflow.

This lets you now create services with workflows like the following:

```
+----------+              +----------+
| ServiceA |              | ServiceB |
+-----+----+              +-----^----+
      |                         |
 when eventX               call taskX
 when result                    |
      |     +----------+        |
      +---->+ ServiceC +--------+
            +----------+
            |workflow1 |
            |workflow2 |
            +----------+

```

**Service A**
- task: `testEvent()` => emits the eventX event with `foo: string, bar: bool`
- task: `testResult()` => returns `foo: string, bar: bool`
```json
{"sid":"service-a","name":"serviceA","tasks":[{"key":"testEvent","inputs":[],"outputs":[]},{"key":"testResult","inputs":[],"outputs":[{"key":"foo","type":"String","object":[]},{"key":"bar","type":"Boolean","object":[]}]}],"events":[{"key":"eventX","data":[{"key":"foo","type":"String","object":[]},{"key":"bar","type":"Boolean","object":[]}]}],"dependencies":[],"source":"QmWf9yTChsvy6HL2SiLhbRNSwRqXFSoFZ4PDikjRRwxncR"}
```

_Deploy and start the service with the API or MESG CLI_

**Service B**
- task: `taskX(foo: string, bar: bool)`
```json
{"sid":"service-b","name":"serviceB","tasks":[{"key":"taskX","inputs":[{"key":"foo","type":"String","object":[]},{"key":"bar","type":"Boolean","object":[]}],"outputs":[{"key":"message","type":"String","object":[]}]}],"events":[],"dependencies":[],"source":"QmSZaGhQX2HRGrQahQy9LsxA2sr3doS9MtoSBBwTcPL6SP"}
```

_Deploy and start the service with the API or MESG CLI_

**Service C**
```json
{
  "sid": "service-c",
  "name": "serviceC",
  "tasks": [],
  "events": [],
  "dependencies": [],
  "source": "QmUwVgi4LuuDCqmzWvGgMHRABwg7T74ngUNxASbNYvJidL"
}
```

In **service C** add the following workflows:
- When `ServiceA` emits `eventX` with `foo = hello` => call `ServiceB#taskX`
- When `ServiceA#testResult` => call `ServiceB#taskX`
```json
"workflows": [
    {
      "trigger": {
        "type": "Event",
        "instanceHash": "7vMG8ac7Xd9z4LfEnS6PfUKM3B7x9hFbaf4LM3JTwrBY",
        "key": "eventX",
        "filters": [
          {
            "key": "foo",
            "predicate": "EQ",
            "value": "hello"
          }
        ]
      },
      "task": {
        "instanceHash": "5fzKqHcXDfRvka3KgHqt2yNWyzCFapFY5BqmfrvMdkLZ",
        "taskKey": "taskX"
      }
    },
    {
      "trigger": {
        "type": "Result",
        "instanceHash": "7vMG8ac7Xd9z4LfEnS6PfUKM3B7x9hFbaf4LM3JTwrBY",
        "key": "testResult",
        "filters": [
          {
            "key": "foo",
            "predicate": "EQ",
            "value": "hello"
          }
        ]
      },
      "task": {
        "instanceHash": "5fzKqHcXDfRvka3KgHqt2yNWyzCFapFY5BqmfrvMdkLZ",
        "taskKey": "taskX"
      }
    }
  ]
```

_Create the service with the API, **not the CLI**_


service/type.go
```diff
@@ -161,3 +181,30 @@ type Parameter struct {
...
+type WorkflowTask struct {
+	InstanceHash hash.Hash `hash:"name:1" validate:"required"`
+	TaskKey      string    `hash:"name:2" validate:"printascii"`
+}
+
+// WorkflowTrigger is an event that triggers a workflow
+type WorkflowTrigger struct {
+	InstanceHash hash.Hash                `hash:"name:1" validate:"required"`
+	Key          string                   `hash:"name:2" validate:"printascii,printascii"`
+	Type         TriggerType              `hash:"name:3"`
```
antho1404: the `required` validation doesn't work I guess because TriggerType can be 0. It could be good to add validation tho if you have any idea

service/type.go
```diff
@@ -161,3 +181,30 @@ type Parameter struct {
...
+	Filters      []*WorkflowTriggerFilter `hash:"name:4" validate:"dive,required"`
+}
+
+// WorkflowTriggerFilter is the filter definition that can be applied to a workflow trigger
+type WorkflowTriggerFilter struct {
+	Key       string            `hash:"name:1" validate:"required,printascii"`
+	Predicate WorkflowPredicate `hash:"name:2"`
```
antho1404: Same comment here for the validation of the enum

server/grpc/api/mapping.go
```diff
@@ -189,3 +252,50 @@ func toProtoDependencies(deps []*service.Dependency) []*types.Service_Dependency
...
+	for i, filter := range filters {
+		var predicate types.Service_Workflow_Trigger_Filter_Predicate
+		// switch filter.Predicate {
+		if filter.Predicate == service.EQ {
+			predicate = types.Service_Workflow_Trigger_Filter_EQ
+		}
+		fs[i] = &types.Service_Workflow_Trigger_Filter{
+			Key:       filter.Key,
+			Predicate: predicate,
+			Value:     filter.Value.(string),
```
antho1404: Can we have filter's value with any type?

protobuf/types/service.proto
```diff
@@ -54,7 +54,48 @@ message Service {
...
+      message Filter {
+
+        // Type of condition available to compare the values.
+        enum Predicate {
+          EQ = 0; // Equal
+        }
+
+        string key = 1;           // Key to check.
+        Predicate predicate = 2;  // Type of condition to apply.
+        string value = 3;         // Value of the filter
```
antho1404: Now filters are only possible with string value

protobuf/types/service.proto
```diff
@@ -65,6 +106,5 @@ message Service {
   repeated Dependency dependencies = 7;     // The container dependencies this service requires.
   string repository = 9;                    // Service's repository url.
   string source = 13;                       // The hash id of service's source code on IPFS.
+  repeated Workflow workflows = 14;         // List of workflows implemented for this service.
```
NicolasMahe:
```suggestion
  repeated Workflow workflows = 14;         // List of workflows defined by this service.
```

service/type.go
NicolasMahe: You need to set a not used default value on the enum.
Something like:
```
const (
        UNKNOWNTRIGGERTYPE TriggerType = iota
	EVENT
	RESULT
)
```
https://github.com/mesg-foundation/engine/pull/1188/files#diff-fd2977ccac84eb22a5975801a69cf64cR40

antho1404: yes but with that, we lose the property that by default we are listening to events.
I could use the `oneof` validation but then it would be `oneof=0 1` which might be a bit weird. Or we could just keep it like that.

protobuf/types/service.proto
```diff
@@ -54,7 +54,48 @@ message Service {
...
+
+        string key = 1;           // Key to check.
+        Predicate predicate = 2;  // Type of condition to apply.
+        string value = 3;         // Value of the filter
+      }
+
+      TriggerType type = 1;         // Type of trigger.
+      string instanceHash = 2;      // Hash of the instance that triggers the workflow.
+      string key = 3;               // Key that triggers the workflow (event key or task key).
+      repeated Filter filters = 4;  // List of filters to apply on the data of the event/result.
```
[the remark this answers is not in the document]
antho1404: Filters are really important if we want to do workflows that are useful. Also, this is developed already, the part here is really the integration in the API with protobuf that doesn't support any non-typed random data.

The filter definitely needs more work. I will write something on the workflow post that could describe the structure of the filter and open the discussion to improve it. But except if there any good reason, I don't see the point to remove it now. It's just a first version of the filter that will have new features but is still valid.

service/type.go
NicolasMahe: why do you want to listen to events by default?

protobuf/types/service.proto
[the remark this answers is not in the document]
NicolasMahe: I'm ok to keep it for test but we have to standardize filters so they can be used on anything before any official release.
Also, a filter can be done by a task of a service that will filter the inputs (maybe with the JS service)

service/type.go
antho1404: It seems to me like a good default, if you put nothing it's because you want to react to an event but i'm ok not to have any default and have an error if not set and this will resolve the required part. I will do the fixes

protobuf/types/service.proto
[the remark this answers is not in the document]
antho1404: I was thinking about a filter with another service but in that case, we lose the access of the initial data so this doesn't work.
Also, I'm totally fine to have this generic filter system but let's not block anything because of that. We already have different filters in the codebase right now (execution stream, event stream) and it didn't block any release.

service/type.go
antho1404: done

protobuf/types/service.proto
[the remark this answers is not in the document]
antho1404: For the filter, there is already a post on the forum https://forum.mesg.com/t/filter-events-and-results-by-data/149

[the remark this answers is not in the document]
ilgooz: Following might give some insights:
https://github.com/ilgooz/service-logic
https://github.com/ilgooz/service-objects

protobuf/types/service.proto
```diff
@@ -54,7 +54,48 @@ message Service {
...
-  
+
+  // A workflow is a configuration to trigger a specific task when certains conditions of a trigger are valid.
+  message Workflow {
+
+    // Trigger is the configuration that will trigger a workflow.
+    message Trigger {
+
+      // Type of trigger available. Event to react to an event. Result to react to a result of an execution.
+      enum TriggerType {
```
krhubert: Can we rename it to `Type`? It's already in the scope of Trigger message

protobuf/types/service.proto
```diff
@@ -54,7 +54,48 @@ message Service {
...
+        Event = 0;  // React to event.
```
krhubert: Here types are enumerated from 0, but in go code from 1.

antho1404: proto doesn't allow enum to start with 0
In Go, it starts with 1 to make sure that we have a value and then be able to do validation without having to manage empty values https://github.com/mesg-foundation/engine/pull/1188#discussion_r304224737

protobuf/types/service.proto
antho1404: done

protobuf/types/service.proto
antho1404: Let's add a new value `UNKNOWN = 0` to be consistent with the go struct

PR #1192, opened 2019-07-22
Replace json format with protobuf Struct

krhubert:
fix https://github.com/mesg-foundation/engine/issues/1176


antho1404:
I don't understand why you don't directly use `import "google/protobuf/struct.proto";`
Now because you use your own struct this is not recognized by proto and it will not do any processing on the message but if you use the one from google you should have pretty much nothing to do.

If I compile with the official struct I have in my generated file a new type imported `_struct "github.com/golang/protobuf/ptypes/struct"`

Also, please description, labels, project related on the pull requests ;)

krhubert:
> I don't understand why you don't directly use import "google/protobuf/struct.proto";

```
google/protobuf/struct.proto: File not found.
protobuf/types/event.proto:3:1: Import "google/protobuf/struct.proto" was not found or had errors.
```

We have to pass path to struct when building (flag `--proto_path`) and we don't keep protobuf definitions in repo. We have to copy it somewhere (so I did it with just few modifications)

> Now because you use your own struct this is not recognized by proto and it will not do any processing on the message but if you use the one from google you should have pretty much nothing to do.

What do you mean "is not recognized by proto"/"will not do any processing on the message" - any more info.

> If I compile with the official struct I have in my generated file a new type imported _struct "github.com/golang/protobuf/ptypes/struct"

Yes, the import path is different (if this is what you mean) because the go_package in the officaial file is different. But they are pretty much the same after generation.

https://github.com/golang/protobuf/blob/master/ptypes/struct/struct.pb.go

>Also, please description, labels, issue solved, project related on the pull requests ;)

Sure :)

krhubert:
fix https://github.com/mesg-foundation/engine/issues/1176

wip: need to fix ethwallet as well.

systemservices/ethwallet/go.mod
```diff
@@ -8,7 +8,9 @@ require (
 	github.com/deckarep/golang-set v1.7.1 // indirect
 	github.com/ethereum/go-ethereum v1.8.27
 	github.com/go-stack/stack v1.8.0 // indirect
-	github.com/mesg-foundation/engine v0.5.1-0.20190702155243-a2826b3949a7
+	github.com/mesg-foundation/core v0.10.1 // indirect
```
NicolasMahe: is it needed?

krhubert:
Ok I think I'm done here, only tests left but for tests I need cli updated @NicolasMahe @antho1404 can you prepare an update of cli so I can test eth wallet and marketplace?

antho1404:
can't this be tested just with the api? Also is this ready to review? there is still the wip label? Please also add details on this PR as discussed

krhubert:
> can't this be tested just with the api? 

It can but we will need mesg-cli anyway, so why to test it twice, if this could be done with mesg-cli.

> Also is this ready to review? there is still the wip label? 

It's ready for review, but it's not tested this is why I left `wip` label.

antho1404:
This would actually make `grpcurl` easier because we don't have to json stringify the inputs and just pass them in a map directly.

example:
```
grpcurl -plaintext -d "{\"instanceHash\":\"6LfmaGAKcpGT93z6fMp1PVqraCbXstWwaSwM7HdvfgPp\", \"taskKey\":\"taskX\", \"inputs\":{\"foo\": \"hello\", \"bar\":\"world\"}}" localhost:50052 api.Execution/Create
```
instead of something like
```
grpcurl -plaintext -d "{\"instanceHash\":\"6LfmaGAKcpGT93z6fMp1PVqraCbXstWwaSwM7HdvfgPp\", \"taskKey\":\"taskX\", \"inputs\":\"{\\\"foo\\\": \\\"hello\\\", \\\"bar\\\":\\\"world\\\"}\"}" localhost:50052 api.Execution/Create
```

Also yes cli and lib needs to be updated when this development is finished and tested

protobuf/api/event.proto
```diff
@@ -45,7 +46,7 @@ message CreateEventRequest {
   string key = 2;
 
   // data is the data for the event.
-  string data = 3;
+  google.protobuf.Struct data = 3;
```
NicolasMahe: As the type changed, we should also change the number:
```suggestion
  google.protobuf.Struct data = 4;
```
Need to be done everywhere else

protobuf/api/execution.proto
```diff
@@ -29,7 +30,7 @@ service Execution {
 message CreateExecutionRequest {
   string instanceHash = 1;
   string taskKey = 2;
-  string inputs = 3;
+  google.protobuf.Struct inputs = 3;
```
NicolasMahe:
```suggestion
  google.protobuf.Struct inputs = 5;
```

protobuf/api/execution.proto
```diff
@@ -74,7 +75,7 @@ message UpdateExecutionRequest {
   // result pass to execution
   oneof result {
     // outputs is a success result.
-    string outputs = 2;
+    google.protobuf.Struct outputs = 2;
```
NicolasMahe:
```suggestion
    google.protobuf.Struct outputs = 4;
```

protobuf/types/event.proto
```diff
@@ -15,5 +17,5 @@ message Event {
   string key = 3;
 
   // data is the data for the event.
-  string data = 4;
+  google.protobuf.Struct data = 4;
```
NicolasMahe:
```suggestion
  google.protobuf.Struct data = 5;
```

protobuf/types/execution.proto
```diff
@@ -44,10 +46,10 @@ message Execution {
   string taskKey = 6;
 
   // inputs data of the execution.
-  string inputs = 7;
+  google.protobuf.Struct inputs = 7;
```
NicolasMahe:
```suggestion
  google.protobuf.Struct inputs = 11;
```

protobuf/types/execution.proto
```diff
@@ -44,10 +46,10 @@ message Execution {
...
 
   // outputs are the returned data of successful execution.
-  string outputs = 8;
+  google.protobuf.Struct outputs = 8;
```
NicolasMahe:
```suggestion
  google.protobuf.Struct outputs = 12;
```

systemservices/ethwallet/client/encoding.go
```diff
@@ -1,13 +1,27 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var output map[string]interface{}
+	if err := json.Unmarshal(b, &output); err != nil {
+		return nil, err
+	}
+
+	return output, nil
```
NicolasMahe: why marshal and then unmarshal?

protobuf/api/event.proto
antho1404: anyway we don't support the string so the previous api breaks anyway so let's not add backward compatibility on the api when we don't support it in the software yet

antho1404:
I have a lot of issues testing with the `ethwallet` service (that is the only one updated with the new api)

Example:
### Invalid inputs for the list task
```bash
grpcurl -plaintext -d "{\"instanceHash\":\"Hsy6y2i7ZxrTMHFb39sm1bLrSB2QenjXEcGKcNjb7kbw\", \"taskKey\":\"list\", \"inputs\":{\"passphrase\":\"1\"}}" localhost:50052 api.Execution/Create
```
No errors in the engine, execution with the passphrase as input and still in pending.

### Generate a new address
```bash
grpcurl -plaintext -d "{\"instanceHash\":\"Hsy6y2i7ZxrTMHFb39sm1bLrSB2QenjXEcGKcNjb7kbw\", \"taskKey\":\"create\", \"inputs\":{\"passphrase\":\"1\"}}" localhost:50052 api.Execution/Create
```
Execution working but with an error `invalid character '\\x11' looking for beginning of value`

NicolasMahe:
> ### Invalid inputs for the list task
> ```shell
> grpcurl -plaintext -d "{\"instanceHash\":\"Hsy6y2i7ZxrTMHFb39sm1bLrSB2QenjXEcGKcNjb7kbw\", \"taskKey\":\"list\", \"inputs\":{\"passphrase\":\"1\"}}" localhost:50052 api.Execution/Create
> ```
> 
> No errors in the engine, execution with the passphrase as input and still in pending.

This is not related to this PR. The Engine only check for required data and their type, not for data that are not declared by the task.

antho1404:
sure but the execution should be completed and it is not but I agree, the parameter is irrelevant

protobuf/api/event.proto
krhubert: +1 for leaving api id as it is.

systemservices/ethwallet/client/encoding.go
krhubert: Just to convert `interface{}` to `map[stirng]interface{}`.

It's not the best client out there, but api is changin and now when it's kind of stable we can think of client api

PR #1204, opened 2019-08-03
Move workflow to its own database/api

antho1404:
Dependency https://github.com/mesg-foundation/engine/pull/1201

There is a lot of copy/paste from the existing service SDK/API. I didn't change the workflow engine, just moved it from `workflow` to `engine` package

- [x] package workflow with the definition
- [x] remove workflows from services
- [x] create protobuf type
- [x] create workflow SDK
- [x] create workflow API
- [x] update the workflow engine to use workflow SDK.

Now there is 4 new apis:
- `Workflow.Create`
- `Workflow.Delete`
- `Workflow.Get`
- `Workflow.List`

When we create the workflow, all the instances will be validated and once recorded in the database the workflow engine will start to process executions related to this workflow.
There is no need to deploy a service with no source code.


krhubert:
Why you want to move workflow package to the engine. It's rather a general name and also the name of main repo.

antho1404:
> Why you want to move workflow package to the engine. It's rather a general name and also the name of main repo.

We have now a model called `workflow` and the workflow engine that is connecting the workflow definition to the executions so this cannot be in the same package.

We could go with `workflow_engine` but https://blog.golang.org/package-names.

I'm open for suggestions on the name for this package or just keep it as it is

NicolasMahe:
> > Why you want to move workflow package to the engine. It's rather a general name and also the name of main repo.
> 
> We have now a model called `workflow` and the workflow engine that is connecting the workflow definition to the executions so this cannot be in the same package.
> 
> We could go with `workflow_engine` but https://blog.golang.org/package-names.
> 
> I'm open for suggestions on the name for this package or just keep it as it is

let's keep it like this for this PR and change it in a future PR.
I suggest to group all workflow stuff in one folder with multiple subpackage (lib approach):
```
workflow
|  types
|  engine
|  ....
```

workflow/type.go
```diff
@@ -0,0 +1,49 @@
...
+// Task describes the instructions for the workflow to execute a task
+type Task struct {
+	InstanceHash hash.Hash `hash:"name:1" validate:"required"`
+	TaskKey      string    `hash:"name:2" validate:"printascii"`
+}
+
+// Trigger is an event that triggers a workflow
+type Trigger struct {
+	InstanceHash hash.Hash        `hash:"name:1" validate:"required"`
+	Key          string           `hash:"name:2" validate:"printascii,printascii"`
```
NicolasMahe:
```suggestion
	Key          string           `hash:"name:2" validate:"printascii"`
```

antho1404:
Another PR created for the renaming https://github.com/mesg-foundation/engine/pull/1208
Let's validate/merge this one if it's ok and discuss on the other PR

krhubert:
@antho1404 just to remind you you can use `"github.com/golang/protobuf/jsonpb"` package to do conversion the same way as for pbstruct. Not necessary in this pr

PR #1206, opened 2019-08-07
Feature/cosmos integration nico

NicolasMahe:
To start the engine in validator mode: no new config 🎊 🍾 
```
MESG_LOG_LEVEL=info ./dev
```

To start a another engine that connect to a validator one:
```
rm -R ~/.mesg2; MESG_TENDERMINT_P2P_SEEDS=97b0a26ff96fcd17f103a941cc5f60c7fe55d705@10.0.98.2:26656 MESG_TENDERMINT_VALIDATORPUBKEY=9CAC089727C943E534D58472294D3157915058A60B6A800BC66A3E8B33D142DA MESG_NAME=engine-1 MESG_SERVER_PORT=50053 MESG_PATH=~/.mesg2 MESG_LOG_LEVEL=info ./dev
```
_don't forget to replace `MESG_TENDERMINT_VALIDATORPUBKEY` and `MESG_TENDERMINT_P2P_SEEDS` ;)_


tendermint/node.go
```diff
@@ -16,28 +17,25 @@ import (
...
 	cfg := config.DefaultConfig()
 	cfg.P2P.Seeds = seeds
 	cfg.SetRoot(root)
 
-	os.MkdirAll(root+"/data", os.FileMode(0755))
-	var me = privval.LoadOrGenFilePV(cfg.PrivValidatorKeyFile(), cfg.PrivValidatorStateFile())
-	nodeKey, err := p2p.LoadOrGenNodeKey(cfg.NodeKeyFile())
-	if err != nil {
-		return nil, err
+	nodeKey := &p2p.NodeKey{
```
NicolasMahe: @krhubert why removing `loadOrGenNodeKey`? The goal of this PR is to use a maximum of stuff already done by Cosmos SDK. We could implement our own system to store the node and validator private keys later on ;)

config/config.go
```diff
@@ -122,3 +123,20 @@ func (c *Config) Validate() error {
 	}
 	return nil
 }
+
+// PubKeyEd25519 is type used to parse value provided by envconfig.
+type PubKeyEd25519 ed25519.PubKeyEd25519
+
+func (key *PubKeyEd25519) Decode(value string) error {
```
NicolasMahe: Nice fix ;)

NicolasMahe: Maybe we could do a PR on tendermint repo to add this function ;)

krhubert: They use viper for receiving flags/args/envs and so I don't think so it would pass :/

tendermint/node.go
krhubert: You said that node key will be generated every time engine start, so why we want to load them? Using loadOrGen is misleading for other devs. 

By generating directly pub key it's clear that the key is changing on every start.

NicolasMahe: The goal of this PR is simply to start tendermint with the most similar logic and code than 
cosmos sdk. we can change the behavior later on.

Only the genesis is generated to avoid to add its hardcoded json version to the source code. But it brings some limitation and complicated when we add store to the cosmos app..

NicolasMahe:
Commit https://github.com/mesg-foundation/engine/pull/1206/commits/e32d91218029f68fb6db8c210fe72b7bd8b07e2c make it possible for a second tendermint node to connect to the validator and receive blocks 🎊

NicolasMahe:
closed in favor of https://github.com/mesg-foundation/engine/pull/1211

PR #1208, opened 2019-08-08
Rename engine package to scheduler

antho1404:
Based on https://github.com/mesg-foundation/engine/pull/1204#issuecomment-518088266

Renaming `engine` package to `scheduler`.
Engine was not really clear because we are already in a repository called engine...


NicolasMahe:
it was also breaking the release process because the CI was trying to build the engine with the name engine that was already taken by the engine package...

PR #1211, opened 2019-08-09
Cosmos integration

NicolasMahe:
This PR integrate Cosmos and mainly tendermint in the Engine 🍾 🎊 

There is 2 ways to run cosmos, as validator (will create the blocks) and as client (will receive blocks from a validator).
A working blockchain should have at least one validator.

To run a validator, execute:
```
./dev --validator
```

To run a client, execute:
```
./dev
```
If you want to run multiple Engine on the same computer, you'll need to set a different name, path and port:
```
MESG_NAME=engine2 MESG_SERVER_PORT=50053 MESG_PATH=~/.mesg2 ./dev
```

To override default, you can specificy the validator with the following envs:
```
MESG_TENDERMINT_VALIDATORPUBKEY=9731AA3C8A4D381397B17726308E441A4FD5028451C99CD7CD5A84BDE9E5D
288 
MESG_TENDERMINT_P2P_SEEDS=c291a80fbad7c95ac7c86251beeabf0e53d7a836@0.tcp.ngrok.io:26656
```

----

The logs from Cosmos are too much and will be configurable in another PR.

---

Closes https://github.com/mesg-foundation/engine/issues/1195


core/main.go
```diff
@@ -148,16 +146,39 @@ func stopRunningServices(sdk *sdk.SDK) error {
...
 	if err != nil {
 		logrus.Fatalln(err)
 	}
 
 	// init logger.
-	logger.Init(dep.config.Log.Format, dep.config.Log.Level, dep.config.Log.ForceColors)
+	logger.Init(cfg.Log.Format, cfg.Log.Level, cfg.Log.ForceColors)
+
+	// init app
+	db := db.NewMemDB()
```
antho1404: doesn't matter for this PR but just a reminder that this is in memory

core/main.go
```diff
@@ -167,9 +188,17 @@ func main() {
...
+			logrus.Fatalln(err)
+		}
+	}()
+
+	logrus.WithField("seeds", cfg.Tendermint.P2P.Seeds).Info("starting tendermint node")
+	go func() {
+		if err := node.Start(); err != nil {
 			logrus.Fatalln(err)
 		}
+		select {}
```
antho1404: why this select?

dev
```diff
@@ -4,26 +4,51 @@ set -e
...
+TENDERMINT_VALIDATOR_PORT=${TENDERMINT_VALIDATOR_PORT:-26656}
+TENDERMINT_VALIDATOR_PATH=${TENDERMINT_VALIDATOR_PATH:-".tendermint-validator"}
+
+# Init and read validator identity
+eval $(go run internal/tools/gen_tm_identity.go -name=$TENDERMINT_VALIDATOR_NAME -port=$TENDERMINT_VALIDATOR_PORT -path=$TENDERMINT_VALIDATOR_PATH)
+MESG_TENDERMINT_P2P_SEEDS=${MESG_TENDERMINT_P2P_SEEDS:-$NODE_PUBKEY}
+MESG_TENDERMINT_VALIDATORPUBKEY=${MESG_TENDERMINT_VALIDATORPUBKEY:-$VALIDATOR_PUBKEY}
+
+# Setup the validator private keys
+if [[ $* == *--validator* ]]; then
```
antho1404: @krhubert if you know a better way to read flag but definitely not blocking

tendermint/node.go
```diff
@@ -0,0 +1,75 @@
...
+		logger,
+	)
+}
+
+func genesisLoader(validator crypto.PubKey) func() (*types.GenesisDoc, error) {
+	return func() (*types.GenesisDoc, error) {
+		genesis := &types.GenesisDoc{
+			GenesisTime:     time.Date(2019, 8, 8, 0, 0, 0, 0, time.UTC),
+			ChainID:         "xxx",
+			ConsensusParams: types.DefaultConsensusParams(),
```
antho1404: using the default consensus here but a config with 10 sec. Don't we have any issue with that?

tendermint/node.go
```diff
@@ -0,0 +1,75 @@
...
+		node.DefaultMetricsProvider(cfg.Instrumentation),
+		logger,
+	)
+}
+
+func genesisLoader(validator crypto.PubKey) func() (*types.GenesisDoc, error) {
+	return func() (*types.GenesisDoc, error) {
+		genesis := &types.GenesisDoc{
+			GenesisTime:     time.Date(2019, 8, 8, 0, 0, 0, 0, time.UTC),
+			ChainID:         "xxx",
```
antho1404: chain id could be from the params and set in the config

go.mod
```diff
@@ -45,25 +50,31 @@ require (
...
+	golang.org/x/sys v0.0.0-20190712062909-fae7ac547cb7 // indirect
 	golang.org/x/text v0.3.2 // indirect
-	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4 // indirect
-	google.golang.org/genproto v0.0.0-20190620144150-6af8c5fc6601 // indirect
-	google.golang.org/grpc v1.19.0
-	gopkg.in/airbrake/gobrake.v2 v2.0.9 // indirect
-	gopkg.in/gemnasium/logrus-airbrake-hook.v2 v2.1.2 // indirect
+	golang.org/x/tools v0.0.0-20190712213246-8b927904ee0d // indirect
+	google.golang.org/genproto v0.0.0-20190708153700-3bdd9d9f5532 // indirect
+	google.golang.org/grpc v1.22.0
```
antho1404: many dependencies changed what's the reason? Is this really necessary for this PR and if it is we need to make sure it didn't break anything and test all that

dev
antho1404: @krhubert if you know a better way to read flag but definitely not blocking

core/main.go
antho1404: why this select?

antho1404:
there are also issues on codacy because of missing comments to fix

core/main.go
krhubert: node.Start doesn't block like similar http/grpc method server. so, in fact, it dosesn't need to be started it in a separate goroutine

dev
krhubert: yes, there is `getopts` tool (one of many examples https://sookocheff.com/post/bash/parsing-bash-script-arguments-with-shopts/)

tendermint/node.go
krhubert: ConsensusParams != config.Consensus

config.Consensus is stored in `config/config.toml`:

```toml
##### consensus configuration options #####
[consensus]

wal_file = "data/cs.wal/wal"

timeout_propose = "3s"
timeout_propose_delta = "500ms"
timeout_prevote = "1s"
timeout_prevote_delta = "500ms"
timeout_precommit = "1s"
timeout_precommit_delta = "500ms"
timeout_commit = "5s"
# ....
```

ConsensusParams is stored in `config/genesis.json`:

```json
  "consensus_params": {
    "block": {
      "max_bytes": "22020096",
      "max_gas": "-1",
      "time_iota_ms": "1000"
    },
    "evidence": {
      "max_age": "100000"
    },
    "validator": {
      "pub_key_types": [
        "ed25519"
      ]
    }
  },
```

tendermint/node.go
krhubert: we should rename it from xxx to something like `mesg-chain` or similar, but we don't need it to be configurable as it will be const all the time.

go.mod
krhubert: Dependency with cosmos-sdk is the reason.

core/main.go
NicolasMahe: @krhubert please improve

PR #1213, opened 2019-08-09
Print file and function name on logging

krhubert:
example:

```
INFO[0015]/root/go/src/github.com/mesg-foundation/engine/core/main.go:167 main.main() starting MESG Engine version local           
INFO[0015]/root/go/src/github.com/mesg-foundation/engine/core/main.go:175 main.main() starting workflow engine                     
INFO[0015]/root/go/src/github.com/mesg-foundation/engine/server/grpc/server.go:53 github.com/mesg-foundation/engine/server/grpc.(*Server).Serve() server listens on [::]:50052 
```

close https://github.com/mesg-foundation/engine/issues/1212


NicolasMahe:
It shows 2 times the folder path??
```
INFO[0015]/root/go/src/github.com/mesg-foundation/engine/server/grpc/server.go:53 github.com/mesg-foundation/engine/server/grpc.(*Server).Serve() server listens on [::]:50052 
```
That's not usable.
Can you improve it?

What about the JSON output?

---

And add the labels, projects and reviewers on the PR......

krhubert:
> It shows 2 times the folder path??

No, it shows file name and function/method name

> Can you improve it?

How do you want it to look like?

> And add the labels, projects and reviewers on the PR......

There is no project that can be used here (we have only specific projects like network

Label/reviewers added

antho1404:
I put a comment on the issue https://github.com/mesg-foundation/engine/issues/1212.

Basically not use the file, this will be too hard to read and having the file path in a production log doesn't make sense. I think to pass the logger as dependencies is the best option and for now that's what we are doing with cosmos and we just need to do that for the api too and then we can grep what we want on the module field

krhubert:
> and having the file path in a production log doesn't make sense.

We can make it optional and only set for dev purpose.

> I think to pass the logger as dependencies is the best option and for now that's what we are doing with cosmos

But you are talking about passing logger to every package. Let's say I want to print log in a package that dosesn't have any struct, what's then? Suppose I want to print log in event package then where I store the logger?

Here we have one line solution that works and grepping thought logs by package name is a very specific task. If you want to have something to grep you can always use `WithFiled("module_name", "xxx").Info(...)` syntax which is perfect here. 

We can disable cosmos logs by passing nop logger if we want to have less verbose logs. Or just use `logrus.Info` and grep by searching phrase, not package name.

There is also [CallerPrettyfier](https://godoc.org/github.com/sirupsen/logrus#TextFormatter) in the formater which can shorten the function or filename (or even don't display it if we want). I just need to know what format of file/function name do you expect, for example:

```
INFO[0005] main.go:116 main.deployCoreServices Instance started with hash "9SsJyoF4wuGVHMBpQnUma1VuXcvtMQnhxLLXZP5zQX2Y" 
INFO[0005] main.go:89 main.deployCoreServices Deploying service "ethwallet"                
INFO[0005] main.go:102 main.deployCoreServices Service "ethwallet" deployed with hash "AVgXTRVwWmFfs3jARsxCoi9zbpeB16jDTihoDbznA9Ty" 
INFO[0007] main.go:116 main.deployCoreServices Instance started with hash "CuF2egTJFfspiYeywQuWdgm9ZpXGAeGkq8Tj9t1o1GsE" 
INFO[0007] main.go:167 main.main starting MESG Engine version local           
INFO[0007] main.go:175 main.main starting workflow engine                     
INFO[0007] server.go:53 grpc.(*Server).Serve server listens on [::]:50052                 
```

which is already pushed.

NicolasMahe:
If vote to have something like:
```
INFO[0000] validator tx                                  module="cosmos"
```
using `WithFiled("module", "cosmos").Info(...)`

core/main.go
```diff
@@ -86,7 +86,7 @@ func initDependencies() (*dependencies, error) {
 
 func deployCoreServices(config *config.Config, sdk *sdk.SDK) error {
 	for _, serviceConfig := range config.SystemServices {
-		logrus.Infof("Deploying service %q", serviceConfig.Definition.Sid)
+		logrus.WithField("module", "main").Infof("Deploying service %q", serviceConfig.Definition.Sid)
```
NicolasMahe: It could be shortened by something like:
```go
var (
	logger = logrus.WithField("module", "main")
)
```
```suggestion
		logger.Infof("Deploying service %q", serviceConfig.Definition.Sid)
```

krhubert: I can do this when cosmos will be integrated and then will add also with filed in main.go (the one put when dev cosmos branch)

antho1404:
We should have documentation on how to now be able to filter the logs we need. This can be a wiki or just updating the description of this PR might be enough

PR #1214, opened 2019-08-12
Create struct service.Configuration

NicolasMahe:
Closes https://github.com/mesg-foundation/engine/issues/1104

This PR add a new struct `service.Configuration` that is also used by `service.Dependency` by inheritance and update the codebase to propagate the modif.


service/type.go
```diff
@@ -76,32 +76,38 @@ type Service struct {
...
+// Dependency represents a Docker container and it holds instructions about
+// how it should run.
+type Dependency struct {
+	*Configuration
+
+	// Key is the key of dependency.
+	Key string `hash:"name:d1" validate:"printascii"`
+
+	// Image is the Docker image.
+	Image string `hash:"name:d2" validate:"printascii"`
```
NicolasMahe: As it is using inheritance, I choose to name the hash with the prefix "d" so it doesn't override future properties added to Configuration struct.

krhubert: I've just tested structhash package and we don't need to prefix for hash. With struct composition when the two fields with the same name hash is different.
```go
package main

import (
    "fmt"

    "github.com/cnf/structhash"
)

type A struct {
    N string `hash:"name:1"`
    M string `hash:"name:2"`
    B
}

type B struct {
    M string `hash:"name:2"`
}

func main() {
    fmt.Printf("%x\n", structhash.Dump(A{N: "n", M: "m", B: B{M: "m"}}, 0))
    fmt.Printf("%x\n", structhash.Dump(A{N: "n", M: "m-", B: B{M: "m"}}, 0))

    fmt.Printf("%x\n", structhash.Sha1(A{N: "n", M: "m", B: B{M: "m"}}, 0))
    fmt.Printf("%x\n", structhash.Sha1(A{N: "n", M: "m-", B: B{M: "m"}}, 0))
}

```
➜  go run main.go
7b313a226e222c323a226d222c423a7b323a226d227d7d
7b313a226e222c323a226d2d222c423a7b323a226d227d7d
338eec5b413091c4db0e9aa057e861549ca45d9e
11ac72d3248a5dd00d65a1b1640a1fb6bc8326c3
```

File: service/type.go [hunk 2]
```diff
@@ -76,32 +76,38 @@ type Service struct {
...
+	Env []string `hash:"name:6" validate:"unique,dive,env"`
+}
+
+// Dependency represents a Docker container and it holds instructions about
+// how it should run.
+type Dependency struct {
+	*Configuration
+
+	// Key is the key of dependency.
+	Key string `hash:"name:d1" validate:"printascii"`
```
NicolasMahe:
```suggestion
	Key string `hash:"name:1" validate:"printascii"`
```

service/type.go
NicolasMahe:
```suggestion
	Image string `hash:"name:2" validate:"printascii"`
```

NicolasMahe: nice catch 👍  I removed the prefix

PR #1218, opened 2019-08-12
Set tm config in config package and pass it on node creation

krhubert:


config/config.go
```diff
@@ -81,6 +84,10 @@ func New() (*Config, error) {
 	c.Database.ExecutionRelativePath = filepath.Join("database", "executions", executionDBVersion)
 	c.Database.WorkflowRelativePath = filepath.Join("database", "workflows", workflowDBVersion)
 	c.Tendermint.Path = "tendermint"
+	c.Tendermint.Config = config.DefaultConfig()
```
antho1404: I like the idea but let's go 100% to that direction and remove the sub configs that can have conflicts.

So just `config.Tendermint` of type `*config.Config`

PR #1220, opened 2019-08-13
Remove regexp network filter from dev script

NicolasMahe:


antho1404:
what's the fix? and how to reproduce?

krhubert:
before the commit:

```
docker network create --driver overlay engine
./dev
# error: network already exists
```

after it's starting without an error

PR #1223, opened 2019-08-13
Workflow validation

antho1404:
Add workflow validation:
- Basic validation of the required attributes
- Graph structure validation:
  - Do not allow graph with cycles
  - Do not allow graph with multiple roots, (non-connected graphs)
  - Do not allow graph with nodes that contain more than one parent

With this validation, we are allowing only directed tree graphs which is a family of DAG.

You can use the test from this PR: https://github.com/mesg-foundation/engine/pull/1209


workflow/algorithm.go
```diff
@@ -0,0 +1,98 @@
...
+		if !w.isAcyclic(child, copyMap) {
+			return false
+		}
+	}
+	return true
+}
+
+// A connected graph is a graph where all the nodes are connected with each other through edges.
+// Warning: this function will have a stack overflow if the graph is not acyclic.
+func (w *Workflow) isConnected(node string) bool {
```
krhubert:
```suggestion
func (w *Workflow) isConnected() bool {
```

It should retruns if a graph is connected without specifying the starting node. the docs says that the method returns true if all the nodes are connected.

workflow/algorithm.go
```diff
@@ -0,0 +1,98 @@
...
+	visited := make(map[string]bool)
+	root := w.getRoot(node)
+	w.visitChildren(root, visited)
+	return len(visited) == len(w.Nodes)
+}
+
+// walk through all the children of a node and populate a map of visited children.
+func (w *Workflow) visitChildren(node string, visited map[string]bool) {
```
krhubert:
```suggestion
func (w *Workflow) dfs(node string, fn func (node string)) {
```

workflow/workflow.go
```diff
@@ -4,8 +4,42 @@ import (
...
+		return fmt.Errorf("null graph")
+	}
+	if !w.isAcyclic(w.Trigger.NodeKey, map[string]bool{}) {
+		return fmt.Errorf("non acyclic graph")
+	}
+	if !w.isConnected(w.Trigger.NodeKey) {
+		return fmt.Errorf("non connected graph")
+	}
+	if !w.isMonoParental() {
+		return fmt.Errorf("non mono-parental graph")
```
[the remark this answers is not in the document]
antho1404: This is not the right term, I couldn't find a better name. If you know any name that describes that each node should have only one parent I'm really happy to change this

workflow/algorithm.go
```diff
@@ -0,0 +1,98 @@
...
+
+func (w *Workflow) getRoot(node string) string {
+	parents := w.ParentIDs(node)
+	if len(parents) == 0 {
+		return node
+	}
+	if len(parents) > 1 {
+		panic("multiple parents is not supported")
+	}
+	return w.getRoot(parents[0])
```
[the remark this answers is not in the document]
antho1404: it returns a list to make sure we can validate the number of parents in `isMonoParental`

antho1404:
> With this PR a lot of workflow methods handle the graph and not workflow itself. As in previous PR I would love to see the part where the graph is a separate struct.

I agree that's why I started to move all the graph-related stuff in the algorithm file. I don't want to charge too much this PR and also I have to have a look at https://github.com/gonum/gonum/tree/master/graph that might replace all. So this will certainly change soon.

workflow/algorithm.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// An acyclic graph is a graph that doesn't contain a cycle. If you walk through the graph you will go maximum one time on each node.
+func (w *Workflow) isAcyclic(node string, discovered map[string]bool) bool {
+	if discovered[node] {
+		return false
+	}
+	discovered[node] = true
+	for _, child := range w.ChildrenIDs(node) {
+		copyMap := make(map[string]bool)
```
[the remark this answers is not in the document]
antho1404: I did a different version which is more like a backtrack (the one in the article you sent). It checks all the possible cycles from the descendant of a node without and this for each node but memorizing the one already processed

workflow/workflow.go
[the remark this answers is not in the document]
antho1404: changed the function to `maximumParentNode(max int)` that returns true when all the nodes have less than max parents.

workflow/algorithm.go
```diff
@@ -0,0 +1,124 @@
...
+		if node.Key == key {
+			return node, nil
+		}
+	}
+	return Node{}, fmt.Errorf("node %q not found", key)
+}
+
+// A null graph is a graph that contains no nodes
+func (w *Workflow) hasNodes() bool {
+	return len(w.Nodes) == 0
```
NicolasMahe: The bool return is opposite of the name of the function...
on empty graph: hasNodes true?. it should be false.
```suggestion
	return len(w.Nodes) > 0
```

workflow/algorithm_test.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+}
+
+func TestHasNodes(t *testing.T) {
+	var tests = []struct {
+		graph *Workflow
+		null  bool
+	}{
+		{graph: defaultGraph(), null: false},
+		{graph: &Workflow{}, null: true},
```
NicolasMahe: again about hasNodes...
```suggestion
		{graph: &Workflow{}, hasNodes: false},
```

workflow/algorithm_test.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+}
+
+func TestHasNodes(t *testing.T) {
+	var tests = []struct {
+		graph *Workflow
+		null  bool
+	}{
+		{graph: defaultGraph(), null: false},
```
NicolasMahe:
```suggestion
		{graph: defaultGraph(), hasNodes: true},
```

workflow/algorithm_test.go
```diff
@@ -0,0 +1,268 @@
...
+		{graph: &Workflow{}, null: true},
+	}
+	for _, test := range tests {
+		assert.Equal(t, test.graph.hasNodes(), test.null)
```
NicolasMahe:
```suggestion
		assert.Equal(t, test.graph.hasNodes(), test.hasNodes)
```

workflow/algorithm.go
```diff
@@ -0,0 +1,124 @@
...
+	for _, node := range w.Nodes {
+		if len(w.ParentIDs(node.Key)) > max {
+			return false
+		}
+	}
+	return true
+}
+
+func (w *Workflow) shouldBeDirectedTree() error {
+	if w.hasNodes() {
```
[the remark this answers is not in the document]
antho1404: my bad, before it was `isNullGraph` and I just renamed without changing the logic...

workflow/algorithm.go
```diff
@@ -0,0 +1,125 @@
...
+		panic("multiple parents is not supported")
+	}
+	return w.getRoot(parents[0])
+}
+
+// Return the maximum number of parent found in the graph.
+func (w *Workflow) maximumParents() int {
+	max := 0
+	for _, node := range w.Nodes {
+		if max < len(w.ParentIDs(node.Key)) {
```
krhubert:
```suggestion
		if l := len(w.ParentIDs(node.Key)); max < l {
```

antho1404: done

PR #1227, opened 2019-08-14
Possibility to enable/disable network when running the engine

antho1404:
Now the engine has a flag `experimental-network` that is `false` by default.
When this flag is enabled the tendermint node is created and started. If not the engine starts without any network.

The `dev` script runs the engine with the flag but by default, the network will not be started. If we want to run the network we need to manually run the docker image with the flag `experimental-network`

### DIFF
--- a/core/main.go
+++ b/core/main.go
@@ -166,7 +166,7 @@ func main() {
 		if err != nil {
 			logrus.WithField("module", "main").Fatalln(err)
 		}
-		logrus.WithField("seeds", cfg.Tendermint.P2P.Seeds).Info("starting tendermint node")
+		logrus.WithField("module", "main").WithField("seeds", cfg.Tendermint.P2P.Seeds).Info("starting tendermint node")
 		if err := node.Start(); err != nil {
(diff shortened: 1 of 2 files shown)


core/main.go
```diff
@@ -156,10 +160,16 @@ func main() {
...
-	node, err := tendermint.NewNode(cfg.Tendermint.Config, &cfg.Cosmos)
-	if err != nil {
-		logrus.Fatalln(err)
+	if *network {
+		// create tendermint node
+		node, err := tendermint.NewNode(cfg.Tendermint.Config, &cfg.Cosmos)
+		if err != nil {
+			logrus.WithField("module", "main").Fatalln(err)
+		}
+		logrus.WithField("seeds", cfg.Tendermint.P2P.Seeds).Info("starting tendermint node")
```
krhubert:
```suggestion
	logrus.WithField("module", "main").WithField("seeds", cfg.Tendermint.P2P.Seeds).Info("starting tendermint node")
```

PR #1241, opened 2019-08-21
Workflow input resolution

antho1404:
fix https://github.com/mesg-foundation/engine/issues/1238

This pull request adds the support of inputs mapping for the tasks executions.
It is now possible to connect a specific output of a task to a specific input of the task to execute.
This is done for each task's connections (on the edge).

Every edge has the following:
- A list of inputs identified by their key
- A reference value for each input that links to the output key of a specific node

A reference can reference any output of an execution (except the trigger) from the execution chain. It will recursively iterate over all the past executions until it matches the one with the output needed.

TO TEST:
Need to be tested with the API. The CLI doesn't support this feature yet.
Steps:
- Create a workflow definition with at least 2 tasks
- Create an edge that connects 2 of the tasks
- Add the input binding

```json
"edges": [
   { ...
       "inputs": [
        {
          "key": "inputA",
          "ref": {
            "nodeKey": "node-0",
            "key": "value"
          }
        }
      ]
   }
]
```

full example:
```json
{"key":"Event","trigger":{"instanceHash":"6hDyE3NurqLtxgV4AtXMpLemNMEzvEiTgDDwtxjZzdPR","eventKey":"eventX","nodeKey":"node-0"},"nodes":[{"key":"node-0","taskKey":"taskX","instanceHash":"6hDyE3NurqLtxgV4AtXMpLemNMEzvEiTgDDwtxjZzdPR"},{"key":"node-1","taskKey":"taskY","instanceHash":"2iLAvYeJu78wi4Y4sHBKmVYnM3PVVqP8gpCocUnnLJBH"},{"key":"node-2","taskKey":"taskY","instanceHash":"2iLAvYeJu78wi4Y4sHBKmVYnM3PVVqP8gpCocUnnLJBH"}],"edges":[{"src":"node-0","dst":"node-1","inputs":[{"key":"inputA","ref":{"nodeKey":"node-0","key":"value"}}]},{"src":"node-1","dst":"node-2","inputs":[{"key":"inputA","ref":{"nodeKey":"node-0","key":"value"}}]}]}
```

based on:
```json
{"name":"serviceA","tasks":[{"inputs":[{"type":"String","key":"message"}],"outputs":[{"type":"String","key":"value"}],"key":"taskX"},{"key":"emits"}],"events":[{"data":[{"type":"String","key":"message"}],"key":"eventX"}],"configuration":{},"hash":"F8v5pwfTcFK8BDUXAbRUGyjJZ88Xd2d8J7p4RgKZdczg","sid":"serviceA","source":"QmT4DzLGD8GHsvJkfQuzmabpbDCVZAA7NgeRYz3y485kW9"}
```
and 
```json
{"name":"serviceB","tasks":[{"inputs":[{"type":"String","key":"inputA"}],"outputs":[{"type":"String","key":"res"}],"key":"taskY"}],"configuration":{},"hash":"7GJiYp6XRtZGgQKbQrnvTvDWeTPhAmjSMs9T5S83aPiY","sid":"serviceB","source":"QmRzco9eDNjyyjagd5qd1c1nDJ6eZupg71fYXdyU9JXVpL"}
```

and can be triggered with 
```
mesg-cli service:execute serviceA emits
```


protobuf/types/workflow.proto
```diff
@@ -40,8 +40,20 @@ message Workflow {
...
-    string src = 1; // Source of the edge.
-    string dst = 2; // Destination of the edge.
+    message Input {
+      message Reference {
+        string nodeKey = 1; // Key of the node in the graph. If empty, will be using the src of the edge.
+        string key = 2;     // Data from the references node key as defined in the service definition.
+      }
+
+      string key = 1;       // Key of the input (as defined in the the service's task definition).
+      oneof value {
```
antho1404: I'm using a oneof here because I will add the support of constants in it. This will be done later in another PR.

protobuf/types/workflow.proto
```diff
@@ -40,8 +40,20 @@ message Workflow {
   }
 
   message Edge {
-    string src = 1; // Source of the edge.
-    string dst = 2; // Destination of the edge.
+    message Input {
+      message Reference {
+        string nodeKey = 1; // Key of the node in the graph. If empty, will be using the src of the edge.
+        string key = 2;     // Data from the references node key as defined in the service definition.
```
NicolasMahe:
```suggestion
        string key = 2;     // Key of a specific parameter of the referenced node's output data.
```

scheduler/scheduler.go
```diff
@@ -127,8 +128,47 @@ func (s *Scheduler) processExecution(exec *execution.Execution) {
...
+			return nil, err
+		}
+		inputs[input.Key] = value
+	}
+	return inputs, nil
+}
+
+func (s *Scheduler) resolveInput(wfHash hash.Hash, exec *execution.Execution, nodeKey string, outputKey string) (interface{}, error) {
+	if !wfHash.Equal(exec.WorkflowHash) {
+		return nil, nil
```
NicolasMahe: shouldn't it return an error in this case?

scheduler/scheduler.go
```diff
@@ -127,8 +128,47 @@ func (s *Scheduler) processExecution(exec *execution.Execution) {
 			s.ErrC <- err
 			continue
 		}
-		if _, err := s.execution.Execute(wf.Hash, nextStep.InstanceHash, nil, exec.Hash, nextStepID, nextStep.TaskKey, exec.Outputs, []string{}); err != nil {
+		edge, err := wf.FindEdge(exec.StepID, nextStepID)
```
NicolasMahe: The functions `ChildrenIDs` and `FindEdge` should be replaced by `EdgesFrom(src nodeKey)`. Both function are doing almost the same thing.

NicolasMahe: Same apply to `ParentIDs`. Returning the edges instead of the just the node key is more reusable.

workflow/workflow.go
```diff
@@ -33,6 +30,24 @@ func (w *Workflow) Validate() error {
 	return nil
 }
 
+func (e *Edge) validate(w *Workflow) error {
```
NicolasMahe: that's a weird function. An edge that accept the workflow? why moving it to a new function anyway?

antho1404: just wanted to move it because this validation will be more and more complex with validation of the references, inputs/outputs etc... but I agree maybe `func (w Workflow) validateEdge(e Edge)` would be better

PR #1243, opened 2019-08-22
Add store package

NicolasMahe:
Add a new store package in database package to create a common implementation of both cosmos kvstore and goleveldb.

---

TODO:
- [ ] unit test


krhubert:
no panic handle, for example has method might look like:

```go
func (s *CosmosStore) Has(key []byte) (has bool, err error) {
  defer func() {
    if r := recover(); r != nil {
      var ok bool
      if err, ok = r.(error); !ok {
        err = fmt.Errorf("store: %s", r)
      }
    }
  }()
  has = s.store.Has(key)
  return
}
```

I know it's ugly, but cosmos store uses panics and there is no other option to handle this ...

krhubert:
Also do we need store dir under `database`? I think we can just have `database/` or `store/`, what do you think?

NicolasMahe:
> no panic handle, for example has method might look like:
> 
> ```go
> func (s *CosmosStore) Has(key []byte) (has bool, err error) {
>   defer func() {
>     if r := recover(); r != nil {
>       var ok bool
>       if err, ok = r.(error); !ok {
>         err = fmt.Errorf("store: %s", r)
>       }
>     }
>   }()
>   has = s.store.Has(key)
>   return
> }
> ```
> 
> I know it's ugly, but cosmos store uses panics and there is no other option to handle this ...

Could you create a PR that add the error management?

> Also do we need store dir under database? I think we can just have database/ or store/, what do you think?
The files were originally in `/store` but i moved them in database at it is database related and only use by database. I'm open to move it back. @antho1404 what do you think?

krhubert:
> The files were originally in /store but i moved them in database at it is database related and only use by database. I'm open to move it back. @antho1404 what do you think?

Ok, could be `database/store` then :)

antho1404:
Please fix the comments issue on codacy before merge

database/store/leveldb.go
```diff
@@ -0,0 +1,82 @@
...
+	return s.db.Put(key, value, nil)
+}
+
+func (s *LevelDBStore) Close() error {
+	return s.db.Close()
+}
+
+// LevelDBIterator
+type LevelDBIterator struct {
+	iter iterator.Iterator
```
krhubert: Why you want to create a new type if level db iterator already has all needed method. You can return it directly.

NicolasMahe: good catch

PR #1244, opened 2019-08-22
Update service db to use store

NicolasMahe:
Dependant on https://github.com/mesg-foundation/engine/pull/1243

- Remove useless interface ServiceDB
- Rename LevelDBServiceDB to ServiceDB
- Update ServiceDB to accept a store instead
- Update rest of codebase and tests


antho1404:
Please fix the comments issue on codacy before merge

database/service_db.go
```diff
@@ -83,55 +63,49 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
 }
 
 // Delete deletes service from database.
-func (d *LevelDBServiceDB) Delete(hash hash.Hash) error {
-	tx, err := d.db.OpenTransaction()
+func (d *ServiceDB) Delete(hash hash.Hash) error {
+	has, err := d.s.Has(hash)
```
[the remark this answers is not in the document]
krhubert: + ServiceDB should only be responsible for marshal/unmarshal inputs/outpus or handle special errors

database/service_db_test.go
```diff
@@ -82,7 +84,10 @@ func TestServiceDBDelete(t *testing.T) {
 	require.IsType(t, &ErrNotFound{}, err)
 }
 
+// TOFIX: the database is not thread safe anymore...
```
krhubert: the database is still thread safe, but the question is why the transaction has gone

database/service_db.go
```diff
@@ -83,55 +63,49 @@ func (d *LevelDBServiceDB) All() ([]*service.Service, error) {
...
 		return err
 	}
-	if _, err := tx.Get(hash, nil); err != nil {
-		tx.Discard()
-		if err == leveldb.ErrNotFound {
-			return &ErrNotFound{resource: "service", hash: hash}
-		}
-		return err
+	if !has {
+		return &ErrNotFound{resource: "service", hash: hash}
```
krhubert: Also `ErrNotFound` should not be used anymore. A method like `Has` should be exposed.

The database.IsErrNotFound is used exactly **once** in instance sdk package:

```go
  // check if instance already exists
  _, err = i.instanceDB.Get(instanceHash)
  if err == nil {
    return nil, &AlreadyExistsError{Hash: instanceHash}
  } 
  if !database.IsErrNotFound(err) {
    return nil, err
  }
```

`Has` should be used here and only implemented in InstanceDB because there is needed.


So to summurize, here fmt.Error should be used + with refactoring instance db the ErrNotFound type should completly gone.

krhubert: Btw the same story is with `AlreadyExistsError` in service sdk.

database/service_db_test.go
NicolasMahe: kvstore does not implement transaction as it should modify and read the store one transaction at a time. so they don't need concurrent access..

PR #1248, opened 2019-08-22
Add Cosmos helpers

NicolasMahe:
This PR adds a bunch of helper to interact with cosmos easier:

- App builder
This struct help to create a cosmos app and setup the base app correctly.
This struct will be sent to SDK to register every module necessary for the Cosmos App.

- Module builder
This stuct helps to create a cosmos app module builder from a name, a querier and an handler function.
SDK sub package will use it to register cosmos app module without hundreds of repetitive lines of code.

- Keybase
A nicer and easier API on top of cosmos keybase

- Transaction builder
A nicer and easier API on top of cosmos tx builder

- Node
A struct to init a tendermint correctly using the engine's config structs

---

EDIT
- Add file to init the modules of cosmos in order to start the cosmos app properly

---

TODO in other PR:
- [ ] unit tests: https://github.com/mesg-foundation/engine/issues/1256


antho1404:
I also got an error running just `./dev`.
```
FATA[0000] Cannot encode unregistered concrete type types.MsgCreateValidator.  module=main
```

cosmos/module.go
```diff
@@ -0,0 +1,118 @@
...
+func (m AppModule) QuerierRoute() string {
+	return m.name
+}
+
+func (m AppModule) NewQuerierHandler() sdk.Querier {
+	return func(request sdk.Request, path []string, req abci.RequestQuery) ([]byte, sdk.Error) {
+		data, err := m.querier(request, path, req)
+		if err != nil {
+			return nil, sdk.ErrInternal(err.Error())
+		}
```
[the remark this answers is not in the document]
krhubert: > I don't really like to force the error as an internal error. 

The think is cosmos expects that interface.

> We should just return the error that we have and m.querier should return a sdk.Error (not sure this is the best solution).

Querier is just helper struct that dosen't have to implement any interface thus it should stick with `error` type.

> The stuff I don't like here is that we will not have the possibility to have specific error like NotFound, AuthFailure or stuff like that that will provide good context.

I think we can. Internal is one of basic [codes](https://github.com/cosmos/cosmos-sdk/blob/master/types/errors.go#L30) and I guess the app can retrun any code type we want.

cosmos/node.go
```diff
@@ -1,123 +1,84 @@
...
+		return nil, err
+	}
+
+	// init node
+	node, err := node.NewNode(
+		cfg,
+		privval.LoadOrGenFilePV(cfg.PrivValidatorKeyFile(), cfg.PrivValidatorStateFile()),
 		nodeKey,
-		proxy.NewLocalClientCreator(app),
+		proxy.NewLocalClientCreator(app.BaseApp()),
```
krhubert: We should pass an instance of the app here, not the base app.

cosmos/node.go
```diff
@@ -1,123 +1,84 @@
...
 
-	return node.NewNode(cfg,
-		me,
+	nodeKey, err := p2p.LoadOrGenNodeKey(cfg.NodeKeyFile())
+	if err != nil {
+		return nil, err
+	}
+
+	// init node
+	node, err := node.NewNode(
```
krhubert:
```suggestion
	return node.NewNode(
```

Instead of `if err`

cosmos/node.go
```diff
@@ -1,123 +1,84 @@
...
+	}
 
-	return txBldr.SignStdTx(accountName, accountPassword, stdTx, false)
+	return node, nil
 }
 
-func createAppState(cdc *codec.Codec, address sdktypes.AccAddress, signedStdTx authtypes.StdTx) (map[string]json.RawMessage, error) {
-	appState := app.ModuleBasics.DefaultGenesis()
-
+func createAppState(defaultGenesis map[string]json.RawMessage, cdc *codec.Codec, address sdktypes.AccAddress, signedStdTx authtypes.StdTx) (map[string]json.RawMessage, error) {
```
krhubert:
```suggestion
func createAppState(defaultGenesisŚtate map[string]json.RawMessage, cdc *codec.Codec, address sdktypes.AccAddress, signedStdTx authtypes.StdTx) (map[string]json.RawMessage, error) {
```

Here we create app state and the funciton recives `defaultGenesisState` (aka `appState` `defaultAppState`). `defaultGenesis` would mean the whole genesis struct and it wouldn't be true in that case.

Also `func SetGenTxsInAppGenesisState(cdc *codec.Codec, appGenesisState map[string]json.RawMessage` expects appGenesisState, not appGenesis

cosmos/txbuilder.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// NewTxBuilder returns a new initialized TxBuilder.
+func NewTxBuilder(cdc *codec.Codec, accNumber, accSeq uint64, kb *Keybase, chainID string) TxBuilder {
+	return TxBuilder{
+		authtypes.NewTxBuilder(
+			authutils.GetTxEncoder(cdc),
+			accNumber,
+			accSeq,
+			flags.DefaultGasLimit,
```
krhubert: This won't be true for every message, but we can figure it out later in another PR.

cosmos/txbuilder.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/tendermint/tendermint/types"
+)
+
+// TxBuilder implements a transaction context created in SDK modules.
+type TxBuilder struct {
+	authtypes.TxBuilder
+}
+
+// NewTxBuilder returns a new initialized TxBuilder.
+func NewTxBuilder(cdc *codec.Codec, accNumber, accSeq uint64, kb *Keybase, chainID string) TxBuilder {
```
krhubert:
```suggestion
func NewTxBuilder(cdc *codec.Codec, accNumber, accSeq uint64, kb *keys.Keybase, chainID string) TxBuilder {
```

We don't need to pass our struct, as TxBuilder needs the interface.

cosmos/txbuilder.go
```diff
@@ -0,0 +1,48 @@
...
+			chainID,
+			"",
+			sdktypes.NewCoins(),
+			sdktypes.DecCoins{},
+		).WithKeybase(kb),
+	}
+}
+
+// Create a signed transaction from a message.
+func (b TxBuilder) Create(msg sdktypes.Msg, accountName, accountPassword string) (authtypes.StdTx, error) {
```
krhubert:
```suggestion
func (b TxBuilder) BuildAndSignStdTx(msg sdktypes.Msg, accountName, accountPassword string) (authtypes.StdTx, error) {
```

Create is too general in this context.

cosmos/module.go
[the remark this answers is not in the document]
NicolasMahe: > The stuff I don't like here is that we will not have the possibility to have specific error like `NotFound`, `AuthFailure` or stuff like that that will provide good context.

This can be improved over time. The goal here is to expose a simple API. We can imagine a system that cast error to cosmos error and return it if it is, otherwise create a generic error.

cosmos/initModules.go
```diff
@@ -0,0 +1,117 @@
+package cosmos
```
NicolasMahe: this file will move in another place in the next PR

NicolasMahe:
> I also got an error running just `./dev`.
> 
> ```
> FATA[0000] Cannot encode unregistered concrete type types.MsgCreateValidator.  module=main
> ```

fixed. had to bring back the cosmos modules initialization

PR #1252, opened 2019-08-22
Add service sdk cosmos version step 2

NicolasMahe:
Dependant on https://github.com/mesg-foundation/engine/pull/1250 and https://github.com/mesg-foundation/engine/pull/1249

- Add Cosmos compatible service sdk


core/main.go
```diff
@@ -173,11 +173,11 @@ func main() {
 		}
 	} else {
 		sdk = enginesdk.NewDeprecated(c, serviceDB, instanceDB, executionDB, workflowDB, cfg.Name, strconv.Itoa(port))
-	}
 
-	// init system services.
-	if err := deployCoreServices(cfg, sdk); err != nil {
-		logrus.WithField("module", "main").Fatalln(err)
+		// init system services.
+		if err := deployCoreServices(cfg, sdk); err != nil {
```
antho1404: just a warning here, core services are just started when we run the deprecated sdk but are stopped all the time. We might have some errors when we stop them.

PR #1254, opened 2019-08-23
Remove database ErrNotFound. Introduce Database Exist function.

krhubert:
- add exist method to instance db
- use db.exist in instance sdk

### DIFF
--- a/database/instance_db.go
+++ b/database/instance_db.go
@@ -16,6 +16,9 @@ var (
 
 // InstanceDB describes the API of Instance database.
 type InstanceDB interface {
+	// Exist check if instance with given hash exist.
+	Exist(hash hash.Hash) (bool, error)
+
 	// Get retrives instance by instance hash.
 	Get(hash hash.Hash) (*instance.Instance, error)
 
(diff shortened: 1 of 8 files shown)


sdk/service/service.go
```diff
@@ -67,9 +66,9 @@ func (s *Service) Create(srv *service.Service) (*service.Service, error) {
 	}
 	srv.Hash = hash.Hash(h)
 
-	// check if service already exists.
+	// check if service already exists and retrun it.
 	if _, err := s.serviceDB.Get(srv.Hash); err == nil {
-		return nil, &AlreadyExistsError{Hash: srv.Hash}
+		return srv, nil
```
NicolasMahe: i think an error should be returned anyway. the modification you want to do is changing the behavior of this function.
You should do something like instance, with instance.Exist and returns an error:
https://github.com/mesg-foundation/engine/pull/1254/files#diff-8e12c6dba187b74b4392823599528018R90

The same can be applied to workflow,

sdk/service/service.go
```diff
@@ -67,9 +66,9 @@ func (s *Service) Create(srv *service.Service) (*service.Service, error) {
 	}
 	srv.Hash = hash.Hash(h)
 
-	// check if service already exists.
+	// check if service already exists and retrun it.
```
NicolasMahe: typo

sdk/workflow/workflow.go
```diff
@@ -27,6 +25,11 @@ func New(instance *instancesdk.Instance, workflowDB database.WorkflowDB) *Workfl
 func (w *Workflow) Create(wf *workflow.Workflow) (*workflow.Workflow, error) {
 	wf.Hash = hash.Dump(wf)
 
+	// check if workflow already exists.
+	if _, err := w.workflowDB.Get(wf.Hash); err == nil {
+		return nil, err
```
NicolasMahe: err is nil here
```suggestion
		return nil, fmt.Errorf("workflow %q already exists", wf.Hash.String())
```

PR #1255, opened 2019-08-23
Create tendermint client

krhubert:
Create a simple helper that could be used as tendermint client


cosmos/client/client.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	if txres.Code != abci.CodeTypeOK {
+		return fmt.Errorf("transaction returned with invalid code %d", txres.Code)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+	defer cancel()
+
+	out, err := c.Subscribe(ctx, "", "tx.hash='"+txres.Hash.String()+"'")
```
antho1404: Just to make sure, this subscribe is only to get the result of the txhash but we already have the `txres` so why do we need to subscribe?
If this subscribe is to wait for the end of the transaction, a timeout of 20 seconds might be too short.
If this subscribe is for some reason get the transaction isn't it a problem that we subscribe after broadcasting? what happens if it's too fast and we didn't subscribe yet? especially that there is a `BroadcastTxSync`

NicolasMahe: This is the first draft of this system. We need to use it to see the best way to implement this ;)

PR #1257, opened 2019-08-23
Recover from panic in cosmos store

krhubert:


database/store/cosmos.go
```diff
@@ -24,32 +25,67 @@ func (s *CosmosStore) NewIterator() Iterator {
...
+	defer func() {
+		if r := recover(); r != nil {
+			var ok bool
+			if err, ok = r.(error); !ok {
+				err = fmt.Errorf("store: %s", r)
+			}
+		}
+	}()
+	has = s.store.Has(key)
+	return
```
antho1404: can we return explicitly the error here and avoid to have named outputs to be consistent with the rest of the codebase

NicolasMahe: Unfortunately, that's not possible with the error recover system. The output have to be named in order to be "replaced" by the defer function.

antho1404: declaring an error variable at the beginning of the function wouldn't work? 
```go
func (s *CosmosStore) Has(key []byte) (bool, error) {
	var err error
	defer func() {
		if r := recover(); r != nil {
			var ok bool
			if err, ok = r.(error); !ok {
				err = fmt.Errorf("store: %s", r)
			}
		}
	}()
	return s.store.Has(key), err
}
```

krhubert: Nope, because defer is invoked 
"after" retrun statment, not before :/

you can read about this in details here https://medium.com/@blanchon.vincent/go-how-does-defer-statement-work-1a9492689b6e

PR #1258, opened 2019-08-26
Pure graph implementation for the orchestration system

antho1404:
This PR changes the way we designed the workflow.
Instead of having a graph only for the tasks and trigger this graph of tasks when a trigger is matched. We now include the trigger as a node of this graph and we process all the dependencies for this node (this even allows a multi-tree of tasks based on the same trigger or event multiple triggers).

The mapping has also been considered as a node in order to be able to chain them and reduce the complexity of a task node and providing more flexibility in the future.

Benefits:
With this structure where everything is a node, we can easily switch to a graph database where different objects type are different nodes in the database and so having better performances on the lookup of the nodes to resolve. It also provides the flexibility to add new types of processing.

Here is the graph structure:

### Nodes:
- [x] **Task**: A definition that will create a new execution when processed.
- [x] **Event**: A definition of an event to listen. Previously called trigger. Only one trigger is allowed in the graph.
- [x] **Result**: A definition of a result to listen. Previously called trigger. Only one trigger is allowed in the graph. A result can be from a different workflow where connections between tasks are in the same workflow.
- [x] **Map**: A definition to map data. If a `nodeKey` is execution, it will find the matching outputs in the execution's graph, otherwise, it will map based on the previous current data (from a trigger or another mapping)
- [ ] **Filter**: A definition of a filter based on the data. This filter will stop the discovery in the graph if not valid. (will be implemented later)

### Edges:
Classic edges composed of `src` and `dst` that can connect any kind of node

## To Test:

3 services:
```json
// serviceA
{"name":"serviceA","tasks":[{"inputs":[{"type":"String","key":"message"}],"outputs":[{"type":"String","key":"value"}],"key":"taskX"},{"key":"emits"}],"events":[{"data":[{"type":"String","key":"message"}],"key":"eventX"}],"configuration":{},"hash":"F8v5pwfTcFK8BDUXAbRUGyjJZ88Xd2d8J7p4RgKZdczg","sid":"serviceA","source":"QmT4DzLGD8GHsvJkfQuzmabpbDCVZAA7NgeRYz3y485kW9"}
// serviceB
{"name":"serviceB","tasks":[{"inputs":[{"type":"String","key":"inputA"}],"outputs":[{"type":"String","key":"res"}],"key":"taskY"}],"configuration":{},"hash":"7GJiYp6XRtZGgQKbQrnvTvDWeTPhAmjSMs9T5S83aPiY","sid":"serviceB","source":"QmRzco9eDNjyyjagd5qd1c1nDJ6eZupg71fYXdyU9JXVpL"}
// serviceC
{"name":"serviceC","tasks":[{"inputs":[{"type":"String","key":"message"}],"outputs":[{"type":"String","key":"message"}],"key":"taskZ"},{"inputs":[{"type":"String","key":"message"}],"outputs":[{"type":"String","key":"message"}],"key":"taskA"}],"configuration":{},"hash":"F3qDSqUwn2mxBpRv3LUzL6vTNqfvTFvoQnv1KhGnuTsn","sid":"serviceC","source":"QmPuHX3LumnDkrxY3wUR8J2J9xmXGw8yTVy5jf7P24MprN"}
```

2 types of workflow:

`serviceA#eventX` => `serviceA#taskX` => `map(inputA: serviceA#taskX.value)` => `serviceB#taskY`
```json
{
  "key": "Event",
  "nodes": [
    {
      "event": {
        "key": "trigger:event",
        "instanceHash": "6hDyE3NurqLtxgV4AtXMpLemNMEzvEiTgDDwtxjZzdPR",
        "eventKey": "eventX"
      }
    },
    {
      "task": {
        "key": "node1",
        "taskKey": "taskX",
        "instanceHash": "6hDyE3NurqLtxgV4AtXMpLemNMEzvEiTgDDwtxjZzdPR"
      }
    },
    {
      "map": {
        "key": "map1",
        "outputs": [{
          "key": "inputA",
          "ref": {
            "nodeKey": "node1",
            "key": "value"
          }
        }]
      }
    },
    {
      "task": {
        "key": "node2",
        "taskKey": "taskY",
        "instanceHash": "2iLAvYeJu78wi4Y4sHBKmVYnM3PVVqP8gpCocUnnLJBH"
      }
    }
  ],
  "edges": [
    {
      "src": "trigger:event",
      "dst": "node1"
    },
    {
      "src": "node1",
      "dst": "map1"
    },
    {
      "src": "map1",
      "dst": "node2"
    }
  ]
}
```

`serviceB#taskY` => `map(message: serviceB#taskY.res)` => `serviceC#taskZ`

```json
{
  "key": "result-from-other-workflow",
  "nodes": [
    {
      "result": {
        "key": "trigger:result",
        "instanceHash": "2iLAvYeJu78wi4Y4sHBKmVYnM3PVVqP8gpCocUnnLJBH",
        "taskKey": "taskY"
      }
    },
    {
      "task": {
        "key": "task1",
        "instanceHash": "FXGpKzcEHMnyr2BQrH44Dq2SoAYtjw1C7qNRVGbxTNpa",
        "taskKey": "taskZ"
      }
    },
    {
      "map": {
        "key": "map",
        "outputs": [
          {
            "key": "message",
            "ref": {
              "nodeKey": "trigger:result",
              "key": "res"
            }
          }
        ]
      }
    }
  ],
  "edges": [
    {
      "src": "trigger:result",
      "dst": "map"
    },
    {
      "src": "map",
      "dst": "task1"
    }
  ]
}
```

Happy hacking :)


workflow/filter.go
```diff
@@ -0,0 +1,41 @@
+package workflow
+
+// // Predicate is the type of conditions that can be applied in a filter of a workflow trigger
```
antho1404: keep this in comment to reuse it to manage the filters on another PR

workflow/graph.go
```diff
@@ -0,0 +1,178 @@
+package workflow
```
antho1404: This is basically a file renaming from algorithm.go to graph.go but with a new graph struct

workflow/workflow.go
```diff
@@ -6,32 +6,44 @@ import (
 	validator "gopkg.in/go-playground/validator.v9"
 )
 
+// ID is the ID of the Result's node
+func (r Result) ID() string {
+	return "trigger:result"
```
antho1404: we could let the user set its own key on the triggers but as it is unique for now it's not necessary. But happy to change if you have any problem with this hardcoded value

scheduler/scheduler.go
```diff
@@ -10,6 +10,7 @@ import (
 	executionsdk "github.com/mesg-foundation/engine/sdk/execution"
 	workflowsdk "github.com/mesg-foundation/engine/sdk/workflow"
 	"github.com/mesg-foundation/engine/workflow"
+	"github.com/sirupsen/logrus"
 )
 
 // Scheduler manages the executions based on the definition of the workflows
```
krhubert: This package got bigger with PR and still, there is not a single test :/

scheduler/scheduler.go
```diff
@@ -47,111 +48,141 @@ func (s *Scheduler) Start() error {
...
 }
 
-func (s *Scheduler) mapInputs(wfHash hash.Hash, prevExec *execution.Execution, edge workflow.Edge) (map[string]interface{}, error) {
-	if len(edge.Inputs) == 0 {
-		return prevExec.Outputs, nil
-	}
-	inputs := make(map[string]interface{})
-	for _, input := range edge.Inputs {
-		value, err := s.resolveInput(wfHash, prevExec, input.Ref.NodeKey, input.Ref.Key)
+func (s *Scheduler) canProcessChildren(node workflow.Node) bool {
```
krhubert:
```go
func (s *Scheduler) canProcessChildren(node workflow.Node) bool {
  switch node.(type) {
  case *workflow.Event, *workflow.Result, *workflow.Mapping:
    return true
  default:
  return false
  }
}
```

scheduler/scheduler.go
```diff
@@ -47,111 +48,141 @@ func (s *Scheduler) Start() error {
...
-	if err != nil {
-		s.ErrC <- err
-		return
-	}
-	for _, wf := range workflows {
-		nextStep, err := wf.FindNode(wf.Trigger.NodeKey)
-		if err != nil {
-			s.ErrC <- err
-			continue
+func (s *Scheduler) eventFilter(event *event.Event) func(wf *workflow.Workflow, node workflow.Node) (bool, error) {
```
krhubert: There are no comments for methods

scheduler/scheduler.go
```diff
@@ -167,3 +198,15 @@ func (s *Scheduler) resolveInput(wfHash hash.Hash, exec *execution.Execution, no
...
+
+func (s *Scheduler) processTask(task *workflow.Task, wf *workflow.Workflow, exec *execution.Execution, event *event.Event, data map[string]interface{}) error {
+	var eventHash, execHash hash.Hash
+	if event != nil {
+		eventHash = event.Hash
+	}
+	if exec != nil {
+		execHash = exec.Hash
+	}
+	_, err := s.execution.Execute(wf.Hash, task.InstanceHash, eventHash, execHash, task.ID(), task.TaskKey, data, []string{})
```
krhubert:
```suggestion
	_, err := s.execution.Execute(wf.Hash, task.InstanceHash, eventHash, execHash, task.ID(), task.TaskKey, data, nil)
```

sdk/workflow/workflow.go
```diff
@@ -27,12 +27,20 @@ func New(instance *instancesdk.Instance, workflowDB database.WorkflowDB) *Workfl
 func (w *Workflow) Create(wf *workflow.Workflow) (*workflow.Workflow, error) {
 	wf.Hash = hash.Dump(wf)
 
-	if _, err := w.instance.Get(wf.Trigger.InstanceHash); err != nil {
-		return nil, err
-	}
 	for _, node := range wf.Nodes {
-		if _, err := w.instance.Get(node.InstanceHash); err != nil {
-			return nil, err
+		switch n := node.(type) {
```
krhubert: If the `node` is an interface the method `GetInstanceHash` could be added and used here instead of switch. Otherwise:

```
    switch n := node.(type) {
    case workflow.Result:
      instanceHash = n.InstanceHash
    case workflow.Event:
      instanceHash = n.InstanceHash
    case workflow.Task:
      instanceHash = n.InstanceHash
    }
    if _, err := w.instance.Get(instanceHash); err != nil { 
      return nil, err
    }
```

workflow/marshal.go
```diff
@@ -0,0 +1,145 @@
+package workflow
```
krhubert: I don't see where the MarshalJSON is used, can you point me

workflow/graph.go
```diff
@@ -0,0 +1,178 @@
...
+
+// walk through all the children of a node and populate a map of visited children.
+func (g Graph) dfs(node string, fn func(node string)) {
+	fn(node)
+	for _, n := range g.ChildrenIDs(node) {
+		g.dfs(n, fn)
+	}
+}
+
+func (g Graph) getRoot(node string) string {
```
krhubert: No docs

scheduler/scheduler.go
antho1404: this actually makes the scheduler smaller and yes I agree it needs testing, this is based on the sdk and don't have any mock for it so it's hard to reproduce behaviors. But yes I will add some tests on another PR

sdk/workflow/workflow.go
antho1404: `instanceHash` is specific to some types of node, not all of them, mapping and filter doesn't have this.

Maybe I can use this `case workflow.Result, workflow.Event...`

workflow/marshal.go
antho1404: The marshal is just on the nodes, not on the graph, but the unmarshal is on the graph and create the specific nodes.

The marshal is used in the database because we still encode everything in json in the db

scheduler/scheduler.go
krhubert: > and don't have any mock for it so it's hard to reproduce behaviors.

Please don't use mocks. Rather then mocks try to organize the code so it could be unit tested.

>  But yes I will add some tests on another PR

This functionality is in this PR so I think the tests should also be here. It's not the first pr about graph without testing and PR about filter has already been started. I'm ok to add them in PR about filters, but not later.

sdk/workflow/workflow.go
krhubert: > instanceHash is specific to some types of node, not all of them, mapping and filter doesn't have this.

Ok. So in order to have this method, they need to return nil on `GetInstanceHash` if we want. such behavior. 

> Maybe I can use this case workflow.Result, workflow.Event...

Nope, it won't work here.

workflow/marshal.go
krhubert: > The marshal is used in the database because we still encode everything in json in the db

👍 . So why do we need custom marhsal/unmurshal in that case?

antho1404: json is not typed so when we marshal the data we have a json with the data of each node but when we unmarshal we don't know what type of node it is.
The marshal/unmarshal here adds the type in the serialization.
https://github.com/mesg-foundation/engine/pull/1258/files#diff-f46b0dfcf69190d0795e825675bfdf5aR13 and https://github.com/mesg-foundation/engine/pull/1258/files#diff-f46b0dfcf69190d0795e825675bfdf5aR97

sdk/workflow/workflow.go
antho1404: But instance hash is not in all the types of node so id doesn't make sense to have this for each node. We could try to have another interface that is `nodeWithInstanceHash` or something like that but that's definitely overkilled in my opinion

krhubert: So switch stays :)

workflow/workflow.go
NicolasMahe: I definitely suggest to force the node to define the key. Otherwise the client (cli) have also to know that this specific type of node has this specific key during the compilation of the workflow.

protobuf/types/workflow.proto
```diff
@@ -3,62 +3,73 @@ syntax = "proto3";
...
+  message Task {
     string key = 1;           // Key that identifies the node.
     string instanceHash = 2;  // Hash of the instance to execute.
     string taskKey = 3;       // Task of the instance to execute.
   }
-
-  message Edge {
-    message Input {
+  
+  message Mapping {
```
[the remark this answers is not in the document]
antho1404: done

protobuf/types/workflow.proto
```diff
@@ -3,62 +3,73 @@ syntax = "proto3";
...
-    oneof key {
-      string taskKey = 2;         // Key of the task that triggers the workflow.
-      string eventKey = 3;        // Key of the event that triggers the workflow.
-    }
-    repeated Filter filters = 4;  // List of filters to apply on the data of the event/result.
-    string nodeKey = 5;           // First node to trigger when the workflow starts.
+
+// A workflow is a configuration to trigger a specific task when certains conditions of a trigger are valid.
+message Workflow {
+  message Result {
```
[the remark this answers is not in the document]
antho1404: done

workflow/workflow.go
antho1404: done

scheduler/scheduler.go
```diff
@@ -47,111 +48,136 @@ func (s *Scheduler) Start() error {
...
-			s.ErrC <- err
-			continue
+	case *workflow.Task:
+		if err := s.processTask(node, wf, exec, event, data); err != nil {
+			return err
 		}
-		if _, err := s.execution.Execute(wf.Hash, nextStep.InstanceHash, nil, exec.Hash, edge.Dst, nextStep.TaskKey, inputs, []string{}); err != nil {
-			s.ErrC <- err
+	}
+	if s.canProcessChildren(n) {
```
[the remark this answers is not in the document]
antho1404: I simplified the logic for the processNode function

NicolasMahe:
@antho1404 can you update the workflow in description so they can work with the latest modif?

antho1404:
description updated

PR #1260, opened 2019-08-26
Keep only one Service struct (the one generated from protobuf definition)

krhubert:
This PR remove custom service struct (and mapping) leaving only one protobuf Service

Changes:

- all service tags move to protobuf definition
- replace service struct with protobuf generated
- (protobuf definition) Dependency has Configuration as a field
- (protobuf definition) change package name
- add marshal/unmarshal/size/marhslajson/unmarhsaljson methods to hash in order to use it as protobuf custom type
- download and install gogo protobuf in Dockerfile.tools
- update build-proto.sh to use gogo for service

CONS:

- no grpcurl support anymore

TODO:

- protobuf id and hash tag name are different and it literally burns my eyes.
```
    string key = 8 [
      (gogoproto.moretags) = 'hash:"name:1" validate:"printascii"'
    ];

    // Image's name of the container.
    string image = 1 [
      (gogoproto.moretags) = 'hash:"name:2" validate:"printascii"'
    ];

    // Configuration of the container.
    Configuration configuration = 3 [
      (gogoproto.nullable) = false
    ];
```
I think we should update it with this PR and keep them consistent from now on.

- testing with grpccurl
- testing customtype (Hash) with javascript

I put this PR as Ready To Review, as this part could be done even without testing.


protobuf/types/service.proto
```diff
@@ -45,21 +49,31 @@ message Service {
...
   message Dependency {
-    string key = 8;                   // Dependency's key.
-    string image = 1;                 // Image's name of the container.
-    repeated string volumes = 2;      // List of volumes.
-    repeated string volumesFrom = 3;  // List of volumes mounted from other dependencies.
-    repeated string ports = 4;        // List of ports the container exposes.
-    repeated string args = 6;         // Args to pass to the container.
-    string command = 5;               // Command to run the container.
-    repeated string env = 9;          // Default env vars to apply to dependency on runtime.
+    string key = 1;                   // Dependency's key.
```
NicolasMahe:
```suggestion
    string key = 8;                   // Dependency's key.
```
keep previous number so the CLI and mesg-js are still compatible!

protobuf/types/service.proto
```diff
@@ -1,68 +1,206 @@
...
+      (gogoproto.moretags) = 'hash:"name:1" validate:"printascii"'
+    ];
+
+    // Image's name of the container.
+    string image = 1 [
+      (gogoproto.moretags) = 'hash:"name:2" validate:"printascii"'
+    ];
+
+    // Configuration of the container.
+    Configuration configuration = 3 [
```
NicolasMahe: Let's find another way to do this without breaking the API ;)
One solution is to keep `Dependency` and `Configuration` as two different messages (no composition) like "before".
And then, because proto already generate the getter on the properties, a interface can be created for the functions that use both structs :)
What do you think guys?

krhubert:
gogo protobuf has an option to remove those fields. They were already removed.

protobuf/types/service.proto
krhubert: I think we should fix our API asap to have it clean and consistent. 

Instead of refactoring more code and creating more interfaces we should just change it to make it cleaner.

NicolasMahe: I don't think it really cleaner.. it forces `configuration` and `dependency` to share the exact same properties and behavior. This is more an internal simplification that an external (API) system.
I'm going to do a PR from this PR to do it. I think it will be pretty simple and small.

protobuf/types/service.proto
```diff
@@ -1,68 +1,206 @@
...
+    ];
+
+    // List of ports the container exposes.
+    repeated string ports = 3 [
+      (gogoproto.moretags) = 'hash:"name:3" validate:"unique,dive,portmap"'
+    ];
+
+    // Args to pass to the container.
+    repeated string args = 4 [
+      (gogoproto.moretags) = 'hash:"name:5" validate:"printascii"'
```
NicolasMahe: missing validation
```suggestion
      (gogoproto.moretags) = 'hash:"name:5" validate:"dive,printascii"'
```

protobuf/types/service.proto
```diff
@@ -1,68 +1,229 @@
 syntax = "proto3";
 
-package types;
-option go_package = "github.com/mesg-foundation/engine/protobuf/types";
+import "gogo/protobuf/gogoproto/gogo.proto";
+
+package service;
```
NicolasMahe: the proto package doesn't need to change. only the go_package need to change 👍 
```suggestion
package types;
```
like this the file `protobuf/api/service.proto` doesn't need any modification

krhubert: I don't liek the idea of different package names in different context, this is why I think it's better to have the same package name.

NicolasMahe: this have some impact on mesg-js. and again we are speaking about proto package, not go.
the generated types are different now
```diff
-            tasks?: (types.Service.ITask[]|null);
+            tasks?: (service.Service.ITask[]|null);
```

krhubert: yes I know and it's only one line of code and I still think they should stay the same (both types or service)

PR #1266, opened 2019-08-28
Workflow/Scheduler renaming

antho1404:
Dependency https://github.com/mesg-foundation/engine/pull/1259

Rename workflow to process and scheduler to orchestrator


workflow/workflow.go
```diff
@@ -38,8 +38,8 @@ func (w *Workflow) Validate() error {
 		return err
 	}
 	for _, node := range w.Graph.Nodes {
-		n, isMapping := node.(Mapping)
-		if isMapping {
+		n, isMap := node.(Map)
+		if isMap {
```
krhubert: if n, ok  := node.(Map); ok {

filter/filter.go
```diff
@@ -0,0 +1,42 @@
...
+		if !condition.Match(data) {
+			return false
+		}
+	}
+
+	return true
+}
+
+// Match returns true the current filter matches the given data
+func (f Condition) Match(inputs map[string]interface{}) bool {
```
krhubert: Keep method definition to it's corespondent struct

protobuf/types/workflow.proto
```diff
@@ -56,11 +40,29 @@ message Workflow {
       string key = 1;               // Key of the mapping.
       repeated Output outputs = 2;  // Outputs of the mapping.
     }
+
+    message Filter {
```
krhubert: If `Filter` with contintion is here, then go `filter` package should be moved under workflow|scheducler directory.

Otherwise move message Filter out of workflow space.

config/config.go
```diff
@@ -49,7 +49,7 @@ type Config struct {
 		ServiceRelativePath   string
 		InstanceRelativePath  string
 		ExecutionRelativePath string
-		WorkflowRelativePath  string
+		ProcessRelativePath   string
```
krhubert: Remove RelativePath as engine dosen't need it them and they are not used anymore.

orchestrator/orchestrator.go
```diff
@@ -0,0 +1,201 @@
+package orchestrator
```
krhubert: No Test :(

krhubert:
So, In general, this is a really big pr, but so few tests. PR needs a lot more tests for me.

config/config.go
antho1404: This is totally needed in order to create and access the database

antho1404:
closing in favor of https://github.com/mesg-foundation/engine/pull/1278 that has the same modifications without the bad history from https://github.com/mesg-foundation/engine/pull/1259

PR #1267, opened 2019-08-28
Service SDK step 3

NicolasMahe:


core/main.go
```diff
@@ -35,7 +36,7 @@ func initDatabases(cfg *config.Config) (*database.ServiceDB, *database.LevelDBIn
 	if err != nil {
 		return nil, nil, nil, nil, err
 	}
-	serviceDB := database.NewServiceDB(serviceStore)
+	serviceDB := database.NewServiceDB(serviceStore, json.Marshal, json.Unmarshal)
```
krhubert: we shouldn't pass the marshaler and unmarshaler here. It can be coded directly in the database. We will use only one encoding/decoding so it shouldn't be in the public api.

NicolasMahe: it's because currently there is 2 encoding function used: one for the goleveldb, one with cosmos..
I cannot find of a nicer way..
Maybe two `NewServiceDB` functions? one for deprecated and one for cosmos?

NicolasMahe: otherwise, what about simply removing the database package? the sdk could simply write directly into the underlying store without duplication of code.

krhubert:
Please `git amend` wip message, or maybe just `squash` those 3 commits into one with a more descriptive message.

cosmos/client.go
```diff
@@ -83,7 +81,10 @@ func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg, accName, accPassword str
 		if !ok {
 			return nil, errors.New("result data is not the right type")
 		}
-		return &data.TxResult, nil
+		if data.TxResult.Result.IsErr() {
```
krhubert: What do you think about replacing waiting for event to that [halper func](https://github.com/tendermint/tendermint/blob/b73cfe878682e0b73d9a21ec1d8dc456ddd90215/rpc/client/helpers.go#L58). I just found it and it does exactly the same steps as BuildAndBroadcastMsg needs

sdk/service/sdk.go
```diff
@@ -1,38 +1,64 @@
...
 	return sdk
 }
 
 // Create creates a new service from definition.
-func (s *sdk) Create(req *api.CreateServiceRequest) (*service.Service, error) {
-	return nil, fmt.Errorf("create not implemented")
+func (s *SDK) Create(req *api.CreateServiceRequest) (*service.Service, error) {
+	// TODO: pass account by parameters
+	accountName, accountPassword := "bob", "pass"
+	accNumber, accSeq := uint64(0), uint64(0)
```
krhubert: There is no seq increment -  it won't work with multiple service, will it? the same is for delete method

NicolasMahe: try to create a service and delete it. it's working. maybe the sequence needs to be set when doing multiple transaction in the same block?

cosmos/client.go
NicolasMahe: The helper function was almost perfect but it's for event and not transaction.
But I switch to `EventQueryTxFor(encodedTx).String()` to construct the query instead of doing it manually.

sdk/service/sdk.go
antho1404: I think this works because create then delete will be 2 different type of transaction so the final hash will anyway be different but if you do 2 similar transactions (like re-create the service after) then it will fail because the sequence will be 0 and it will generate exactly the same tx.

This `accSeq` or `accNumber` (not sure what's the difference) should be incremented. This is a guess but it would make sense and that could be the reason why you cannot recreate the same service.

When I manually increment the seq and recreate the service (after deleting it) everything works fine. Same if I don't delete the service the transaction goes through and I have an error that the service already exists.

PR #1270, opened 2019-08-28
Change proto hash type from string to bytes

krhubert:
close https://github.com/mesg-foundation/engine/issues/1265


hash/hash.go
```diff
@@ -104,18 +90,14 @@ func (h Hash) Equal(h1 Hash) bool {
 	return bytes.Equal(h, h1)
 }
 
-// Marshal marshals hash into slice of base58 bytes. It's used by protobuf.
+// Marshal marshals hash into slice of bytes. It's used by protobuf.
```
NicolasMahe: Are those function necessary? the type is still `[]byte`. maybe there is no need for it.

protobuf/types/service.proto
```diff
@@ -177,7 +177,7 @@ message Service {
   }
 
   // Service's hash.
-  string hash = 10 [
+  bytes hash = 10 [
```
NicolasMahe: why not using gogoproto custom type to have directly the hash in the generated struct?
This question apply to every hash.
```
```

sdk/service/utils.go
```diff
@@ -43,7 +43,7 @@ func create(container container.Container, db *database.ServiceDB, srv *service.
 		return nil, err
 	}
 	hsh := hash.Hash(h)
-	srv.Hash = hsh.String()
+	srv.Hash = hsh
```
NicolasMahe:
```suggestion
	srv.Hash = hash.Hash(h)
```

NicolasMahe:
I did all the hash modification from `string` to `[]byte` in this commit: 281dd927da462be1308b8fe90aa44944d116e92e
maybe you could just revert it ;)

protobuf/types/service.proto
krhubert: because this pr is only about changing the hash type. Adding custom type will be in next pr

hash/hash.go
krhubert: Maybe, but I prefer to left them until I will be sure that they are not required by `customtype` extensions. If the are not then I will remove them :)

NicolasMahe:
mesg-js and cli updated in PRs: https://github.com/mesg-foundation/mesg-js/pull/124 and https://github.com/mesg-foundation/cli/pull/138.
@mesg-foundation/mesg-core-team please review and test so we can merge all at once!

hash/hash.go
```diff
@@ -124,29 +106,15 @@ func (h Hash) Size() int {
...
 	}
-	return base58size
+	return size
 }
 
-// MarshalJSON mashals hash into base58 encoded json string.
+// MarshalJSON mashals hash into encoded json string.
 func (h Hash) MarshalJSON() ([]byte, error) {
-	return json.Marshal(base58.Encode(h))
+	return json.Marshal([]byte(h))
```
NicolasMahe: I suggest that the JSON version to be a base58 encoded string.
I the current implementation will put the hash in a binary form inside the json...

NicolasMahe: It will display it as base64: https://play.golang.org/p/q59QB-dZutT
```go
s := S{Bin: []byte("hello world")}
d, _ := json.Marshal(s)
fmt.Println(string(d))
```
produces:
```
{"Bin":"aGVsbG8gd29ybGQ="}
```

PR #1286, opened 2019-09-02
Fix marshal issue with process filter

antho1404:
When creating a new process with filters, the filter was not correctly marshaled/unmarshaled.


antho1404:
@NicolasMahe I added some tests on the PR to check the behavior, that was added in another PR but probably a wrong merge...

PR #1291, opened 2019-09-02
Customize protobuf messages 

krhubert:
Dep: https://github.com/mesg-foundation/engine/pull/1290

- set customtype for hash
- set moretags for hash calculation
- regenerate protobuf apis and types

fix https://github.com/mesg-foundation/engine/issues/1277


protobuf/types/process.proto
```diff
@@ -1,27 +1,43 @@
...
 // A process is a configuration to trigger a specific task when certains conditions of a trigger are valid.
 message Process {
   // Node of the process
   message Node {
     message Result {
       string key = 1;           // Key that identifies the node.
-      bytes instanceHash = 2;  // Hash of the instance that triggers the process.
+
+      // Hash of the instance that triggers the process.
+      bytes instanceHash = 2 [
```
antho1404: missing the `moretags` for all the data. You can find the whole definition here https://github.com/mesg-foundation/engine/blob/2b9afb302adc152576d83a0f8e49811ad1bb5dbb/protobuf/types/process.proto

krhubert: This one will be set (with all validation tags etc) when with PR about removing custom type for event/execution/process.

PR #1296, opened 2019-09-03
Remove and refactor rest protobuf types

krhubert:
- remove filter and condition types
- move filter tests to process
- replace process type
- replace event type
- replace execution type
- add missing hash and validation tags
- fix hash tags

close https://github.com/mesg-foundation/engine/issues/1283 
close https://github.com/mesg-foundation/engine/issues/1282 
close https://github.com/mesg-foundation/engine/issues/1281


orchestrator/orchestrator.go
```diff
@@ -119,21 +128,24 @@ func (s *Orchestrator) execute(filter func(wf *process.Process, node process.Nod
...
 
-func (s *Orchestrator) executeNode(wf *process.Process, n process.Node, exec *execution.Execution, event *event.Event, data map[string]interface{}) error {
-	logrus.WithField("module", "orchestrator").WithField("nodeID", n.ID()).WithField("type", fmt.Sprintf("%T", n)).Debug("process process")
-	if node, ok := n.(*process.Task); ok {
+func (s *Orchestrator) executeNode(wf *process.Process, n *process.Process_Node, exec *execution.Execution, event *event.Event, data map[string]interface{}) error {
+	logrus.WithField("module", "orchestrator").
+		WithField("nodeID", n.ID()).
+		WithField("type", fmt.Sprintf("%T", n)).Debug("process process")
+
+	if node, ok := n.Type.(*process.Process_Node_Task_); ok {
```
antho1404: We could use something like that to be more consistent and I find it easier to read but definitely not something blocking
```suggestion
	if task := n.GetTask(); task != nil {
```

krhubert: Done

krhubert:
I've tested all service commands and found that service:logs dosent' work, @antho1404 can you confrim it?

service/service.go
```diff
@@ -71,8 +74,14 @@ func (s *Service) ValidateEventData(eventKey string, eventData map[string]interf
...
-func (s *Service) RequireTaskInputs(taskKey string, taskInputs map[string]interface{}) error {
-	warnings, err := s.ValidateTaskInputs(taskKey, taskInputs)
+func (s *Service) RequireTaskInputs(taskKey string, taskInputs *types.Struct) error {
+	in := make(map[string]interface{})
+	if taskInputs != nil {
+		if err := convert.Marshal(taskInputs, &in); err != nil {
+			return err
+		}
+	}
+	warnings, err := s.ValidateTaskInputs(taskKey, in)
```
NicolasMahe: will it be possible to update the function `ValidateTaskInputs` and co to accept directly `*types.Struct`?

NicolasMahe: I guess it will be possible and maybe even better for the type casting as `struct.Value` seems to already manage type.

NicolasMahe:
> I've tested all service commands and found that service:logs dosent' work, @antho1404 can you confrim it?

yeah logs is broken.. but it is not because of this PR. it doesn't work on dev branch as well. i will investigate.

Otherwise this PR is working well. I'm approving 👍

PR #1297, opened 2019-09-03
Add explicit marshal and unmarshal function to service db

NicolasMahe:
Add explicit marshal and unmarshal function to service db to make it compatible both with amino and json encoding.


database/service_db.go
```diff
@@ -17,25 +18,33 @@ var (
...
 type ServiceDB struct {
-	s store.Store
+	s         store.Store
+	marshal   func(interface{}) ([]byte, error)
+	unmarshal func([]byte, interface{}) error
 }
 
 // NewServiceDB returns the database which is located under given path.
-func NewServiceDB(s store.Store) *ServiceDB {
+func NewServiceDB(s store.Store, cdc *codec.Codec) *ServiceDB {
```
NicolasMahe: @antho1404 @krhubert do you have a better idea than having 2 New functions?
My first implementation accept directly the `marshal` and `unmarshal` but it was maybe to much.

krhubert: Yes, I have an idea, remove the deprecated version!

otherwise, as temp solution create cdc (same as in sdk-tutorial-app) and use it for marhsal/unmarshal (instead of passing it)

NicolasMahe: Actually, creating new cdc could be a good idea :)
I need to check and test more but from my current understanding, only the cosmos's msg needs to be registered on the cosmos app's cdc. It means that any other encoding/decoding using the cdc could be done on different/new instance of cdc, thus removing a lot of cdc dependency injection.

NicolasMahe: I removed the `NewDeprecated` function and create a new codec for legacy codebase and test.

PR #1309, opened 2019-09-05
Do not convert proto.Struct to map in order to Validate it

krhubert:
It's part of https://github.com/mesg-foundation/engine/issues/1302 issue, where the service validation is done.

I need to do manual testing, but the source code is ready to review.


service/service.go
```diff
@@ -40,103 +36,41 @@ func (s *Service) GetEvent(eventKey string) (*Service_Event, error) {
...
-}
-
-// ValidateEventData produces warnings for event datas that doesn't satisfy their parameter schemas.
-func (s *Service) ValidateEventData(eventKey string, eventData map[string]interface{}) ([]*ParameterWarning, error) {
-	e, err := s.GetEvent(eventKey)
-	if err != nil {
-		return nil, err
-	}
-	return validateParametersSchema(e.Data, eventData), nil
+	return nil, fmt.Errorf("servcie %q - event %q not found", s.Name, eventKey)
```
```suggestion
	return nil, fmt.Errorf("service %q - event %q not found", s.Name, eventKey)
```

service/service_parameter.go
```diff
@@ -0,0 +1,80 @@
+package service
+
+import (
+	"fmt"
+
+	"github.com/gogo/protobuf/types"
+	"github.com/mesg-foundation/engine/x/xerrors"
+)
+
+// ValidateParametersSchema validates data to see if it matches with parameters schema.
```
NicolasMahe:
```suggestion
// validateServiceParameters validates data to see if it matches with parameters schema.
```

PR #1310, opened 2019-09-05
Update gogo protobuf with fixed json and custom tags generation

krhubert:
This PR contains updated gogo protobuf with fixed tag generation.

- Regenerated all protobuf files
- go mod tidy
- fix marshaler with new tags.

### DIFF
--- a/process/marshal.go
+++ b/process/marshal.go
@@ -58,25 +58,25 @@ func (w *Process) UnmarshalJSON(b []byte) error {
 
 func (w *Process) unmarshalNode(nodeType string, marshalData []byte) (*Process_Node, error) {
 	switch nodeType {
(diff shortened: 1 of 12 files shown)


process/marshal.go
```diff
@@ -58,25 +58,25 @@ func (w *Process) UnmarshalJSON(b []byte) error {
 
 func (w *Process) unmarshalNode(nodeType string, marshalData []byte) (*Process_Node, error) {
```
NicolasMahe: I suggest to get rid of those custom marshal function and use jsonpb or proto package to encode and decode instead of default json encoder.
The default json encode (encoding/json) is not working well with interface and oneof fields.

PR #1311, opened 2019-09-06
Use protobuf marshal/unmarshal in process database

krhubert:
- Generate equal method for process
- Change process database marshal to proto
- Remove custom json marshaler from proto
- add tests

dep #1310


database/process_db_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+func TestProcessDB(t *testing.T) {
+	dir, _ := ioutil.TempDir("", "process.db.test")
+	defer os.Remove(dir)
+
+	db, err := NewProcessDB(dir)
+	require.NoError(t, err)
+	defer db.Close()
+
+	p := &process.Process{
```
NicolasMahe: Could you add more process with different data?
like one with only hash and another one with ALL:
```
var testProcess = &process.Process{
	Hash: hash.Int(1),
	Key:  "key",
	Edges: []*process.Process_Edge{
		&process.Process_Edge{
			Src: "dw",
			Dst: "dw",
		},
	},
	Nodes: []*process.Process_Node{
		&process.Process_Node{
			Type: &process.Process_Node_Result_{
				Result: &process.Process_Node_Result{
					Key:          "foo",
					InstanceHash: hash.Int(1),
					TaskKey:      "taskKey",
				},
			},
		},
	},
}
```

PR #1316, opened 2019-09-09
Add constant to process map

antho1404:
Add support of constants for mapping.

Constants are helpful to get the user inputs, specific value, specific function for conversion using service-js, add value that is not part of any result...

It is only possible to set either a ref or a constant for a mapping.

To test:

Services
```json
# serviceA
{"name":"serviceA","tasks":[{"inputs":[{"type":"String","key":"message"}],"outputs":[{"type":"String","key":"value"}],"key":"taskX"},{"key":"emits"}],"events":[{"data":[{"type":"String","key":"message"}],"key":"eventX"}],"configuration":{},"hash":"6QkdJaTHbx3qYT4H169twd7A2pzxDJt8MTy882h9nevn","sid":"serviceA","source":"QmTQW4Dnq44oU2DLq5dxXCHRNEQWaDaGc9ZKDmG1ocJGuv"}

# serviceB
{"name":"serviceB","tasks":[{"inputs":[{"type":"String","key":"inputA"}],"outputs":[{"type":"String","key":"res"}],"key":"taskY"}],"configuration":{},"hash":"j64r3wPevCd8f1vcZVNePBvnwRw6kZKdunAtRtd9hCY","sid":"serviceB","source":"QmQu8kCwPJ1Qydrawmz4STkY4FAwxrZHgkui83umkp94iq"}
```

Process
```yaml
key: Event
steps:
  - type: trigger
    instanceHash: H74Qqq8nT5JZ9GSJmuSWLN5benWZPkUb5pYcvQLsoZX 
    eventKey: eventX
  - type: task
    instanceHash: DTjy36ApV5YVbyXm67iWeAsD9uCbZETcPL1TbUPwHeYJ
    taskKey: taskY
    inputs:
      inputA: hello world
```

You can use the CLI from this Pull Request https://github.com/mesg-foundation/cli/pull/153 in order to test with the API updated


protobuf/types/process.proto
```diff
@@ -91,7 +92,10 @@ message Process {
         oneof value {
           // Input defined as reference.
           Reference ref = 2 [
-            (gogoproto.moretags) = 'hash:"name:2" validate:"required"'
+            (gogoproto.moretags) = 'hash:"name:2" validate:"required_without=Constant"'
```
krhubert: What is the case to put `require_without=Constant` here?

orchestrator/orchestrator.go
```diff
@@ -156,24 +156,27 @@ func (s *Orchestrator) processMap(mapping *process.Process_Node_Map, wf *process
...
-		if ref == nil {
-			continue
-		}
-
-		node, err := wf.FindNode(ref.NodeKey)
-		if err != nil {
-			return nil, err
-		}
+		if ref := output.GetRef(); ref != nil {
+			if ref == nil {
```
krhubert: ref will never be nil here (look above if check)

protobuf/types/process.proto
antho1404: to keep the check that ref is required. I cannot put just required because if constant is given then ref will be nil so the required will fail. In that case, just say that it's required only when there is no constant

orchestrator/orchestrator.go
antho1404: good catch, got some issue with a conflict while rebasing maybe comming from that.

PR #1318, opened 2019-09-09
Fix hash calculation based on proto.Structs

krhubert:
close https://github.com/mesg-foundation/engine/issues/1315


hash/structhash/structhash.go
```diff
@@ -91,15 +91,22 @@ func write(buf *bytes.Buffer, v reflect.Value) *bytes.Buffer {
...
 		vt := v.Type()
 		items := make([]string, 0)
 		for i := 0; i < v.NumField(); i++ {
 			sf := vt.Field(i)
 			to := parseTag(sf)
+
+			// NOTE: structhash will allow to process all interface types.
+			// gogo/protobuf is not able to set tags for directly oneof interface.
+			// see: https://github.com/gogo/protobuf/issues/623
+			if to.name == "" && reflect.Zero(sf.Type).Kind() == reflect.Interface {
```
antho1404: Do I understand this correctly? This means that every single attribute that doesn't declare `hash:"name:X"` will be written in the hash if it's an interface. 

In that case, we really need to make sure that we never forgot any tag `hash` and also that all the generated attributes from proto `XXX_` are not impacting this and added in the hash when they shouldn't or we should add the ignore tag for these attributes.

protobuf/types/struct.proto
```diff
@@ -0,0 +1,115 @@
...
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+syntax = "proto3";
+
+import "gogo/protobuf/gogoproto/gogo.proto";
+
+package mesg.protobuf;
```
antho1404: why not just `type` like the rest of our messages? This would be easier to generate for the engine but also mesg-js lib

krhubert: > why not just type like the rest of our messages? 

I wanted to put emphasis that is still google.protobuf.Struct and not some of our types like event, execution etc...

> This would be easier to generate for the engine but also mesg-js lib

How is this gonna make it easier? Before was google.protobuf and did you find trouble with that?

I can put any type here, but the mesg.protobuf package name was only dictated to not include this "sepcial" case in all our types.

antho1404: Ok i'm fine to keep it like that, I agree it's nice to have something "special".

I just have a hard time to try to generate the types for mesg-js from this script 
https://github.com/mesg-foundation/mesg-js/blob/master/copy-proto and we also need to update the doc here https://github.com/mesg-foundation/docs/blob/master/generate-docs-api.sh

Maybe if you can have a look that would be great, I might be struggling on something for no reasons

krhubert:
```bash
#!/bin/bash -e

DEF_REPO="mesg-foundation/core"
DEF_BRANCH="${1:-dev}"
PROTO_PATH="https://raw.githubusercontent.com/$DEF_REPO/$DEF_BRANCH/protobuf"
DIR="./"

rm -rf src/protobuf
rm -rf src/api/typedef
mkdir -p src/protobuf/api
mkdir -p src/protobuf/types
mkdir -p src/api/typedef

curl -so "./src/protobuf/types/struct.proto" "$PROTO_PATH/types/struct.proto"

ressources="event execution instance service process"
for ressource in $ressources
do
  curl -so "./src/protobuf/api/${ressource}.proto" "$PROTO_PATH/api/${ressource}.proto"
  curl -so "./src/protobuf/types/${ressource}.proto" "$PROTO_PATH/types/${ressource}.proto"
  npx pbjs \
    -t static-module \
    --no-create \
    --no-encode \
    --no-decode \
    --no-verify \
    --no-convert \
    --no-delimited \
    --path ./src \
    -o "./src/api/typedef/$ressource.js" \
    "./src/protobuf/types/${ressource}.proto" \
    "./src/protobuf/api/${ressource}.proto"

  npx pbts --path ./src -o "./src/api/typedef/$ressource.d.ts" --name mesg "./src/api/typedef/$ressource.js"
  rm "./src/api/typedef/$ressource.js"
done

curl -so "./src/protobuf/api/core.proto" "$PROTO_PATH/coreapi/api.proto"
```

changes:
- (nice) sh to bash + added -e
- (nice) add -s option to curl to make output cleaner
- **(must)** get the struct.proto first
- **(must)** add --path - prevoius version worked because struct.proto was in /usr/local/include (or similar on mac) which is listed by protoc. Here the location of `struct.proto` has changed so I've added `--path` option to tell pbjs to search that directory.

It's already late so I'm leaving commit to you. The script works:

```
➜  mesg-js git:(master) ✗ gst
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   copy-proto
	modified:   src/api/typedef/event.d.ts
	modified:   src/api/typedef/execution.d.ts
	modified:   src/api/typedef/instance.d.ts
	modified:   src/api/typedef/process.d.ts
	modified:   src/api/typedef/service.d.ts
	modified:   src/protobuf/api/core.proto
	modified:   src/protobuf/api/event.proto
	modified:   src/protobuf/api/execution.proto
	modified:   src/protobuf/api/instance.proto
	modified:   src/protobuf/api/service.proto
	modified:   src/protobuf/types/event.proto
	modified:   src/protobuf/types/execution.proto

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	src/protobuf/types/struct.proto
```

NOTE:

- you can add `.sh` extension to script name :)
- DIR variable is not used
- DEF_REPO is used once and can be removed
- rm & mkdir can be squashed
- Script download coreapi/api.proto but dosen't generate typedef, the question is - is it really needed to download it then?

script after all above modifications:

```bash
#!/bin/bash -e

DEF_BRANCH="${1:-dev}"
PROTO_PATH="https://raw.githubusercontent.com/mesg-foundation/core/$DEF_BRANCH/protobuf"

rm -rf src/protobuf src/api/typedef
mkdir -p src/protobuf/api src/protobuf/types src/api/typedef

curl -so "./src/protobuf/types/struct.proto" "$PROTO_PATH/types/struct.proto"

for ressource in event execution instance service process
do
  curl -so "./src/protobuf/api/${ressource}.proto" "$PROTO_PATH/api/${ressource}.proto"
  curl -so "./src/protobuf/types/${ressource}.proto" "$PROTO_PATH/types/${ressource}.proto"

  npx pbjs \
    -t static-module \
    --no-create \
    --no-encode \
    --no-decode \
    --no-verify \
    --no-convert \
    --no-delimited \
    --path ./src \
    -o "./src/api/typedef/$ressource.js" \
    "./src/protobuf/types/${ressource}.proto" \
    "./src/protobuf/api/${ressource}.proto"

  npx pbts --path ./src -o "./src/api/typedef/$ressource.d.ts" --name mesg "./src/api/typedef/$ressource.js"
  rm "./src/api/typedef/$ressource.js"
done
```

protobuf/types/struct.proto
```diff
@@ -0,0 +1,115 @@
+// Protocol Buffers - Google's data interchange format
```
NicolasMahe: A link to original file should be included: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/struct.proto
```suggestion
// Source: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/struct.proto
// Protocol Buffers - Google's data interchange format
```

protobuf/types/struct.proto
antho1404: Pull request created on mesg-js, thanks @krhubert for that https://github.com/mesg-foundation/mesg-js/pull/130
Also did some testing on the CLI and this seems to work correctly

hash/structhash/structhash_test.go
```diff
@@ -69,6 +69,37 @@ func TestDump(t *testing.T) {
 			}{nil},
 			"{a:nil}",
 		},
+		{
```
NicolasMahe: The following struct returns `{a:{}}` instead of `{}`.
If the content of the interface doesn't return an hash, the interface should not be in the hash at all. What do you think?
```suggestion
		{
			struct{ a interface{} }{
				a: struct{ b int }{
					b: 1,
				},
			},
			"{}",
		},
		{
```

krhubert: I think no, as it would introduce complexity. Also, it would require to keep a recursive stack to make a decision.

NicolasMahe: Done in https://github.com/mesg-foundation/engine/pull/1319 with a simple solution.

hash/structhash/structhash_test.go
```diff
@@ -69,6 +69,37 @@ func TestDump(t *testing.T) {
...
+						"c": 1,
+					},
+				},
+			},
+			"{a:{b:(\"c\":1)}}",
+		},
+		// NOTE: structhash will allow to process all interface types.
+		// gogo/protobuf is not able to set tags for directly oneof interface.
+		// see: https://github.com/gogo/protobuf/issues/623
+		{
```
NicolasMahe: Another, and way worst problem. The following is returning `{a:{b:1}}`:
```
		},
		{
			struct {
				a interface{}
			}{
				struct {
					b interface{}
				}{
					b: 1,
				},
			},
			"{}",
		},
```

NicolasMahe: Solution, replace in [structhash.go line 105](https://github.com/mesg-foundation/engine/pull/1318/files#diff-dfaac64e3ba689ae3bd308fca77f695cR105):
```diff
- if to.name == "" && reflect.Zero(sf.Type).Kind() == reflect.Interface {
+ if to.name == "" && v.CanInterface() {
```

NicolasMahe: Solution, replace in [structhash.go line 105](https://github.com/mesg-foundation/engine/pull/1318/files#diff-dfaac64e3ba689ae3bd308fca77f695cR105) by:
```diff
- if to.name == "" && reflect.Zero(sf.Type).Kind() == reflect.Interface {
+ if to.name == "" && v.Field(i).Kind() == reflect.Interface && v.Field(i).Elem().Kind() == reflect.Struct {
```

krhubert:
> Test line 60 is wrong:

It's not. Field `a` has no `name` in the tag, only omitempty and `a` is not an interface so it should return an empty struct.

hash/structhash/structhash_test.go
krhubert: This is how it interface works. It's not a bug, it done design.
So let's break it down:

a `struct` is passed with `a interface{]` fileds which has no tag.

A struct is serialized and `a` field is serialized because it's an interface.

Then `interface{}` has struct inside. That struct is serialized because it's not a field of any struct, but rather the object (interface value). The struct contains `b` fields which is an interface so it's serialized and  interface value contains an integer. So the final output should be `{a:{b:1}}`

PR #1320, opened 2019-09-10
Use Structs in the orchestrator

antho1404:
dependency: https://github.com/mesg-foundation/engine/pull/1318

Fix issue with data comparison and replacement in the orchestrator due to the change from `map[string]interface{}` to `types.Struct`. The struct was unmarshalled as `map[string]interface{}` and so containing the struct of the `types.Struct` where our comparison was based on a simple map of value.


krhubert:
@antho1404 can you just squash commits to have a clean git history, otherwise LGTM

PR #1321, opened 2019-09-10
Only force hash calculation on interface containing struct

NicolasMahe:
Small improvement over https://github.com/mesg-foundation/engine/pull/1318.
Only force hash calculation on interface containing struct.

Will merge in https://github.com/mesg-foundation/engine/pull/1318.


antho1404:
@NicolasMahe don't forget to update labels/project/milestone before merge

PR #1322, opened 2019-09-10
Update golang to 1.13

### LINKED ISSUE #1307: Update to go1.13
We should update engine to go1.13. It's still in go1.11.

krhubert:
close #1307

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,7 @@ jobs:
 
   "lint":
     docker:
-      - image: golangci/golangci-lint:v1.12
+      - image: golangci/golangci-lint:v1.18
     environment:
       - GOCACHE: *go_cache_dir
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ run_on_machine: &run_on_machine
 
 run_on_docker: &run_on_docker
   docker:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -69,7 +69,7 @@ jobs:
 
   "lint":
     docker:
-      - image: golangci/golangci-lint:v1.12
+      - image: golangci/golangci-lint:latest
```
krhubert: Nice catch, but don't use `latest` label, as we will have to track it and make sure we have everything (like go version) up to date.

NicolasMahe: I think the linter should be on latest to quickly tell us new problem that have been discovered.
But I agree that it could be boring if suddenly the lint break for some code that is not related to the PR in progress.

NicolasMahe:
```suggestion
      - image: golangci/golangci:v1.18
```

.circleci/config.yml
```diff
@@ -69,7 +69,7 @@ jobs:
 
   "lint":
     docker:
-      - image: golangci/golangci-lint:v1.12
+      - image: golangci/golangci:v1.18
```
NicolasMahe:
```suggestion
      - image: golangci/golangci-lint:v1.18
```

PR #1324, opened 2019-09-10
Update EthWallet service to latest gRPC API version

### LINKED ISSUE #1302: Use Struct from proto instead of map[string]interface
The proto are using `Struct` and `map[string]*Value` from [proto](https://github.com/gogo/protobuf/blob/master/types/struct.pb.go#L64).
The engine should also use it instead of the custom `map[string]interface`.

- [ ] service validation
- [ ] process
- [ ] orchestrator
- [ ] ethwallet client
- [ ] remove convert functions!

krhubert:
EDITED BY NICO:

The ethwallet was broken since the update of the proto namespace and i couldn't find a way to make it work with the `map[string]interface` conversion system.
So I simplify the task `import` to directly return a json string.
The cli has also been updated to be compatible with this updated task: https://github.com/mesg-foundation/cli/pull/166

Related to https://github.com/mesg-foundation/engine/issues/1302


systemservices/ethwallet/ethwallet/export.go
```diff
@@ -33,5 +30,34 @@ func (s *Ethwallet) export(input map[string]interface{}) (map[string]interface{}
...
+	return &types.Struct{
+		Fields: map[string]*types.Value{
+			"address": {
+				Kind: &types.Value_StringValue{
+					StringValue: accountJSON.Address,
+				},
+			},
+			"crypto": {
+				Kind: &types.Value_StringValue{
+					StringValue: string(crypto),
```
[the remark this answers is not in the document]
NicolasMahe: `exportInputs` should be removed

systemservices/ethwallet/ethwallet/create.go
```diff
@@ -1,30 +1,22 @@
...
+func (s *Ethwallet) create(inputs *types.Struct) (*types.Struct, error) {
+	account, err := s.keystore.NewAccount(inputs.Fields["passphrase"].GetStringValue())
 	if err != nil {
 		return nil, err
 	}
 
-	return client.Marshal(createOutputSuccess{
-		Address: account.Address,
-	})
+	return &types.Struct{
```
NicolasMahe: Isn't possible to keep the easy to use `createInputs` and `createOutput` struct and use convert marshal/unmarshal function to convert from those struct to `*types.Struct`?

krhubert: The problem here is api uses `*types.Struct` as inputs, outputs and data. So there are two option 

1) keeps `map[string]interface{}` (or any other type) and converts it to `*types.Struct`
2) Use `*types.Struct` directly

We wanted to get rid of 1 solution, so we go with 3. I can create helper functions for getting fields values or createing values, but that's all.

3) any proposition???

NicolasMahe: just for reference: https://godoc.org/github.com/robteix/protoconv

systemservices/ethwallet/ethwallet/export.go
```diff
@@ -33,5 +25,74 @@ func (s *Ethwallet) export(input map[string]interface{}) (map[string]interface{}
...
-	return client.Marshal(accountJSON)
+
+	// type CryptoJSON struct {
+	// 	Cipher       string                 `json:"cipher"`
+	// 	CipherText   string                 `json:"ciphertext"`
+	// 	CipherParams cipherparamsJSON       `json:"cipherparams"`
+	// 	KDF          string                 `json:"kdf"`
+	// 	KDFParams    map[string]interface{} `json:"kdfparams"`
+	// 	MAC          string                 `json:"mac"`
+	// }
```
NicolasMahe: Don't forget KDFParams.
A example in JSON of cryptoJSON:
```json
{
  "version": 3,
  "address": "c0698bf7fb9c86df805dbd22eefc2728913b5b14",
  "crypto": {
    "mac": "38da0d64786560d7ec3cedd85cf9357ffd5795b50c7d38888a962fafb73ab1be",
    "cipher": "aes-128-ctr",
    "cipherparams": {
      "iv": "febe5f59dc0d21fed9fa1351f12d2dab"
    },
    "ciphertext": "4a08d6c26060e1472c7015dcd2a51a78966755d94ea7d58326f76bb2708da5ee",
    "kdf": "scrypt",
    "kdfparams": {
      "n": 262144,
      "p": 1,
      "r": 8,
      "salt": "c2473a637ddcb5a42e4f86c4cd8f93db69dffb498f9530be0da32953b87fc261",
      "dklen": 32
    }
  },
  "id": "a26fae36-b637-48d1-b13b-0808492db505"
}
```

systemservices/ethwallet/ethwallet/sign.go
```diff
@@ -32,39 +33,61 @@ type signOutputSuccess struct {
 	SignedTransaction string `json:"signedTransaction"`
 }
 
-func (s *Ethwallet) sign(input map[string]interface{}) (map[string]interface{}, error) {
-	var inputs signInputs
-	if err := client.Unmarshal(input, &inputs); err != nil {
-		return nil, err
-	}
+func (s *Ethwallet) sign(inputs *types.Struct) (*types.Struct, error) {
```
NicolasMahe: don't forget to delete `signInputs`, `transaction` and `signOutputSuccess`

NicolasMahe:
This PR needs to be finished and merge in order to release.
The update of the gRPC API and maybe some other stuff really broke the ethwallet.

systemservices/ethwallet/ethwallet/export.go
```diff
@@ -33,5 +25,24 @@ func (s *Ethwallet) export(input map[string]interface{}) (map[string]interface{}
 	if err = json.Unmarshal(keyJSON, &accountJSON); err != nil {
 		return nil, err
 	}
-	return client.Marshal(accountJSON)
+
+	return types.NewStruct(map[string]*types.Value{
```
NicolasMahe: Let's export and import the account as JSON string instead of object! this will simplify the service and give more time to fine a proper solution.

NicolasMahe:
@mesg-foundation/mesg-core-team service updated and fully working. please check and approve 👍

systemservices/ethwallet/ethwallet/create.go
```diff
@@ -1,30 +1,16 @@
...
-
-func (s *Ethwallet) create(input map[string]interface{}) (map[string]interface{}, error) {
-	var inputs createInputs
-	if err := client.Unmarshal(input, &inputs); err != nil {
-		return nil, err
-	}
-
-	account, err := s.keystore.NewAccount(inputs.Passphrase)
+func (s *Ethwallet) create(inputs *types.Struct) (*types.Struct, error) {
+	account, err := s.keystore.NewAccount(inputs.GetStringValue("passphrase"))
```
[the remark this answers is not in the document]
krhubert: What was the reason to remove it? It was just to make the code shorter and in that case you don't check for a nil value so panic is possible.

PR #1326, opened 2019-09-11
Initialize fields for map processing

antho1404:
The map for fields was not initialized and was leading to a crash when trying to assign any value on it


antho1404:
closed in favor of https://github.com/mesg-foundation/engine/pull/1327

PR #1331, opened 2019-09-12
Add --exp to dev command

NicolasMahe:
It's REALLY annoying to deactivate the network by removing the flag in the dev script.
So now we can pass the flag `--experimental-network` to enable the network feature.
This is only a temp solution until the network is the default solution and fully implemented!

This PR is a simplified version of https://github.com/mesg-foundation/engine/pull/1262.

```bash
./dev # no network
./dev --exp # with network
./dev --exp --validator # with network and node is validator
```

### DIFF
--- a/dev
+++ b/dev
@@ -4,8 +4,10 @@ set -e
 
 DIR=$(pwd)
 
-# network to connect multiple engine instances
-TENDERMINT_NETWORK="mesg-tendermint"
+# Flags
+if [[ $* == *--exp* ]]; then
+  EXPERIMENTAL_FLAGS+=" --experimental-network"
+fi
 
 # mesg config variables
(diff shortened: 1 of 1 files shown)


krhubert:
I like it because it was annoying to me as well. 

If it's temporary, can we make a short flag for this? It will be really painful to write `--experimental-network` if one start working on the network. `-n` for network? what do you think?

dev
```diff
@@ -4,8 +4,11 @@ set -e
 
 DIR=$(pwd)
 
-# network to connect multiple engine instances
-TENDERMINT_NETWORK="mesg-tendermint"
+# Flags
+NETWORK=false
```
antho1404: Not used

```suggestion
```

PR #1335, opened 2019-09-12
Remove plural form from instance list proto service

krhubert:
Make api types consistent.


NicolasMahe:
Let's wait for a PR on mesg-js and CLI to implement those modification before merging

PR #1361, opened 2019-09-25
Refactor sdk deprecated and system services

### LINKED ISSUE #1351: Remove deprecated service sdk
- Remove the deprecated service sdk
- Simplify stuff that can be simplified!

---

- Keep the exp flag for future use!

krhubert:
close #1351 

- remove deprecated service sdk
- remove `experimental-network` flag
- keep `experimental` flag in dev script (in go code was removed because linter complains that variable is unused - to add when needed)
- remove support for system services (dev script, deploy, config, core info api)
- refactor config package (remove singleton)

TODO:
- one test from execution package was commented. We need to figure out how we want to approach testing of service sdk (and features sdk services) which will be done on tm network.


protobuf/coreapi/api.proto
```diff
@@ -22,12 +22,5 @@ message InfoRequest {}
 
 // Information concerning the Core from the `info` API.
 message InfoReply {
```
NicolasMahe: we can simply totally remove `Core` api :)

NicolasMahe: Move to https://github.com/mesg-foundation/engine/issues/1362

PR #1368, opened 2019-09-26
Add ownership SDK to support service ownership.

NicolasMahe:
This PR add the ownership support to service.

Instead of simply adding a new parameter `owner` to `service`, a new `ownership` sdk has been created to manage the ownership of any resource.
The main argument to separate ownership from service is to be able to transfer the ownership without modification the service resource itself. The same approach should be used on ALL resource: no edition of data: only creating new resource that reference the previous to change its behavior.

Using this PR and the latest of the CLI, run:
```
./bin/run service:create '{"name":"serviceB","tasks":[{"inputs":[{"type":"String","key":"message"}],"outputs":[{"type":"String","key":"value"}],"key":"taskX"},{"key":"emits"}],"events":[{"data":[{"type":"String","key":"message"}],"key":"eventX"}],"configuration":{},"sid":"serviceD","source":"QmT4DzLGD8GHsvJkfQuzmabpbDCVZAA7NgeRYz3y485kW9"}'  --account=bob --passphrase=pass
```
Wait a bit. Then check the creation of the ownership by running:
```
grpcurl -proto protobuf/api/ownership.proto -plaintext localhost:50052 mesg.api.Ownership/List | jq .
```

Dependant on https://github.com/mesg-foundation/engine/pull/1361
Closes https://github.com/mesg-foundation/engine/issues/1358


antho1404:
Just a note, there are still 2 issues:
- There is a timeout when creating a service (certainly because docker build the image), it works if the image is already built
- We cannot republish the same transaction but that might be a good thing and maybe we shouldn't be able to have two times the exact same transaction.

PR #1375, opened 2019-09-27
Change service hash calculation

### LINKED ISSUE #1372: Simplify service hash calculation and creation.
- The service create backend function need to do less stuff in order to take less time to run. This function will be run by the validators of the network and they have less than 10 sec to create a block that could contains thousands of service creation tx.
- The service hash is currently calculated with the actual source files of the service. This requires to download from IPFS all the source, untar it, iterate of the files and finally create a hash for the source files.
As the IPFS hash that reference the source files is already in the Service struct, and it is deterministic, we could simply use it to replace the manual source files hash calculation.
- The service create backend function is also building the docker image for pre-optimization (as the image is already available locally, it is fast to start the associated instance). This need to be removed as well.

krhubert:
close #1372 

- use only service struct to calculate the hash
- remove dirhash package
- remove container ref from backend struct


NicolasMahe:
I removed the dependence to the other PR as it not necessary

PR #1380, opened 2019-09-30
Feature/config validation

### LINKED ISSUE #1376: Improve config validation
Add go-validator to the config struct to validate and make sure all required info are there!
And also like this we can easily know what is missing when the engine start rather than a `invalid memory` error :)

krhubert:
close #1376 

Add validation for required fields + require minium one signers

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -17,6 +17,7 @@ import (
 	"github.com/sirupsen/logrus"
 	tmconfig "github.com/tendermint/tendermint/config"
 	"github.com/tendermint/tendermint/crypto/ed25519"
+	"gopkg.in/go-playground/validator.v9"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


NicolasMahe:
I add the required flag on all data even if it means that the default config (without envs) is not valid anymore.
Also use function `ValidateBasic` from `authtypes.StdTx`

config/config.go
```diff
@@ -30,46 +31,44 @@ const (
...
+	Path string `validate:"required"`
 
 	Server struct {
-		Address string
+		Address string `validate:"required"`
 	}
 
 	Log struct {
-		Format      string
+		Format      string `validate:"required"`
```
antho1404: We could add `oneof=text json` > https://godoc.org/gopkg.in/go-playground/validator.v9#hdr-One_Of

With that, we can remove l149 and we could do the same for the logLevel

PR #1381, opened 2019-09-30
Remove service deletion

### LINKED ISSUE #1378: Remove delete service function
Return an error for now.

krhubert:
close #1378


antho1404:
Also just as a note, we can always delete a service with the following steps:
- Export all the services with `service:detail` command
- Delete `./mesg/database`
- Reimport all the exported services except the one we want to delete

It's kind of hardcore but there's a way.

krhubert:
> Also just as a note, we can always delete a service with the following steps:

I don't get you. Yes we can do that, but it will only delete the service from your local database. And when node connects to the network, the database will be sync with that service again.

antho1404:
I was just talking in the case of a local dev network ;)

PR #1384, opened 2019-09-30
Add new service exists api

antho1404:
fix https://github.com/mesg-foundation/engine/issues/1382

Add 2 new APIs for the services
- Exists(hash) that returns true if a service hash exists
- Hash(createServiceRequest) that returns the calculated hash of the service

Also, add support of queries with data. Every data sent should be encoded in proto.

**To Test**

You can call the api with grpcurl
```bash
grpcurl -proto protobuf/api/service.proto -plaintext -d "$(cat service.json)" localhost:50052 mesg.api.Service/Hash
```

This will return the hash for the service

```bash
grpcurl -proto protobuf/api/service.proto -plaintext -d "{\"hash\": \"xxxx\"}" localhost:50052 mesg.api.Service/Exists
```

This should return false and if you create the service and execute it again you should have true


sdk/service/backend.go
```diff
@@ -73,6 +74,19 @@ func (s *Backend) querier(request cosmostypes.Request, path []string, req abci.R
 		return s.Get(request, hash)
 	case "list":
 		return s.List(request)
+	case "hash":
```
NicolasMahe: does it need to be a backend function? the sdk could do it no?

protobuf/api/service.proto
```diff
@@ -82,3 +88,27 @@ message ListServiceResponse {
...
+// The response's data for the `Exists` API.
+message ExistsServiceResponse {
+  // True if a service already exists, false otherwise.
+  bool exists = 1;
+}
+
+// The request's data for the `Hash` API.
+message HashServiceResponse {
+  // Hash of the service calculated.
+  bytes hash = 2 [
```
krhubert: Why hash starts from 2?

protobuf/api/service.proto
```diff
@@ -82,3 +88,27 @@ message ListServiceResponse {
   // List of services that match the request's filters.
   repeated types.Service services = 1;
 }
+
+// The request's data for the `List` API.
+message ExistsServiceRequest {
+  // The service's hash of the existing service. This hash is nil if exists is fals.
+  bytes hash = 2 [
```
krhubert: Why hash starts from 2?

sdk/service/backend.go
krhubert: Don't we want to have everthing on the backend at some point?

antho1404: it could but then we have some logic on two different places which is hard to manage

antho1404: @NicolasMahe suggested to have the hash calculation on the sdk and not backend file to avoid the call to the abci, let's think about that, I like to have all the business logic on the backend at least for now until we really see ways to improve it but at least it centralizes the "processing" part.
Anyway let's merge and think about that later

PR #1394, opened 2019-10-07
Standalone script gen gensis

NicolasMahe:
Closes https://github.com/mesg-foundation/engine/issues/1392

- [x] generate the validators identity. store them on the disk
- [x] generate the cosmos account. store them locally
- [x] create the genesis transaction. store it locally
- [x] generate peer list. store it locally
- [x] accept list of validator names instead of the number of validator to generate
- [x] build the script and put it in the engine docker image
---
- [x] extract dev config to a new file
---
To run:
```bash
./scripts/gen-genesis.sh
# to generate more validators:
MESG_VALIDATORS=alice,bob,charles ./scripts/gen-genesis.sh
```


Dockerfile
```diff
@@ -24,4 +25,5 @@ RUN apt-get update && \
       rm -rf /var/lib/apt/lists/*
 WORKDIR /app
 COPY --from=build /project/engine .
+COPY --from=build /project/gen-genesis .
```
krhubert: Why gen-genesis is copied to the image?

Dockerfile
```diff
@@ -16,6 +16,7 @@ COPY . .
 ARG version
 ENV version=${version}
 RUN ./scripts/build-engine.sh
+RUN ./scripts/build-tools.sh
```
krhubert: We shouldn't use our internal dev tools in the official docker image.

Dockerfile
NicolasMahe: Like this the cli can directly execute it from the docker container.
The cli will be able to generate new genesis by using the script.

krhubert: cli calling internal scripts inside docker container grr....

Again this is a workaround for the poorly configuration we have right now. Probably, you don't want to pass and manage all those env in cli, so you have decided to copy the internal script as a "hack", instead of fix the pbm which is hard config management.

If that is the case, sorry I won't approve this PR :P

Dockerfile
NicolasMahe: This is not an internal dev tools. It’s a generic tool also use in dev.

Dockerfile
NicolasMahe: The issue of the cli is not the envs or config but the code itself to generate the genesis transaction, keybase, validator identities. I don’t know any way to do it in JS for now. So the solution of this PR is to be able to call this script from the engine image. We could also compile the script and include it in the CLI but then it brings other issues

NicolasMahe:
I'm closing this PR in favor of https://github.com/mesg-foundation/engine/pull/1412

PR #1400, opened 2019-10-07
Container package refactor

krhubert:
close https://github.com/mesg-foundation/engine/issues/1395

Refactor package and tests:

- cache shared network id
- add clean up method (will also stop the services)
- remove dockertest and mocks packages
- refactor for Build and Shared network
- update docker client
- remove usage on context timeout due to new client option
- update golangci-lint skip files and dirs
- update circleci config


core/main.go
```diff
@@ -165,5 +165,9 @@ func main() {
 	if err := stopRunningServices(sdk); err != nil {
 		logrus.WithField("module", "main").Fatalln(err)
 	}
+
+	if err := c.Cleanup(); err != nil {
```
antho1404: This function is only accessible for `DockerContainer`, it could be good to have it for the `Container` interface. Not blocking at all

container/container.go
```diff
@@ -1,105 +1,243 @@
...
+}
+
+// SharedNetworkID returns the ID of the shared network.
+func (c *DockerContainer) SharedNetworkID() string {
+	return c.sharedNetworkID
+}
+
+// StartService starts a docker service.
+func (c *DockerContainer) StartService(options ServiceOptions) (string, error) {
+	if status, _ := c.serviceStatus(options.Namespace); status == RUNNING {
```
NicolasMahe: No error management of `serviceStatus` here?

container/container.go
```diff
@@ -1,105 +1,243 @@
...
 }
 
-// TimeoutOption receives d which will be set as a timeout value for Docker API calls.
-func TimeoutOption(d time.Duration) Option {
-	return func(c *DockerContainer) {
-		c.callTimeout = d
+// Cleanup cleans all configuration like shared network of running services.
+func (c *DockerContainer) Cleanup() error {
+	// remove shared network
+	if err := c.client.NetworkRemove(context.Background(), c.nsprefix); err != nil {
```
NicolasMahe: I don't think the engine can delete a network its still using...

container/container.go
```diff
@@ -138,52 +271,186 @@ func (c *DockerContainer) FindContainer(namespace string) (types.ContainerJSON,
...
+	defer f.Close()
+	return dockerignore.ReadAll(f)
+}
+
+func (c *DockerContainer) deletePendingContainer(namespace string, maxGraceTime time.Time) error {
+	var (
+		id  string
+		err error
+	)
+	for start := time.Now(); start.Before(maxGraceTime); time.Sleep(100 * time.Millisecond) {
```
NicolasMahe: Does `start` values change over time? I don't see anywhere where `start` is re-assigned.

container/container.go
```diff
@@ -1,105 +1,243 @@
...
+		return nil, err
 	}
-	for _, option := range options {
-		option(c)
+	return c, nil
+}
+
+// Build builds a docker image.
+func (c *DockerContainer) Build(path string) (tag string, err error) {
+	excludeFiles, err := readDockerIgnoreFile(path)
```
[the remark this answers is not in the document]
NicolasMahe: This suggestion could be done in another PR.

container/service_options_test.go
```diff
@@ -1,125 +0,0 @@
-package container
```
NicolasMahe: Any reason to delete the tests of this files?

container/container.go
krhubert: No, because only check needed here if it running.

container/service_options_test.go
krhubert: Yes. It is tested with the docker build test. No need to test it twice. In fact I wanted to get rid of those options and just pass SwarmSpec, but I thought you want to keep it simplified.

core/main.go
krhubert: Why `Container` should have that one? Container is used by sdk and sdk is not and it shouldn't be responsible for cleanups. That interface was created to manage containers, not the container "space" like networks, swarm etc..

container/container.go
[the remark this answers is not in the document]
antho1404: The fix (when we run directly the binary on osx and not in docker) is the following
```go
archive.Untar(resp.Body, path, &archive.TarOptions{
	ChownOpts: &idtools.IDPair{
		UID: os.Geteuid(),
		GID: os.Getegid(),
	},
})
```

antho1404:
I didn't see the modification because of the force push, good that the linter broke.
@krhubert could you fix the issue please :)

container/container.go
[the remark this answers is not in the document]
NicolasMahe: ok so it's the untar that cause problem, not the tar ;)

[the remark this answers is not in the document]
NicolasMahe: @krhubert why not following what docker cli is doing? you didn't answer anything. https://github.com/mesg-foundation/engine/pull/1400#discussion_r334261545

container/container.go
```diff
@@ -1,105 +1,241 @@
...
-// Option is a configuration func for Container.
-type Option func(*DockerContainer)
+	// sharedNetworkID is cache for an id network.
+	sharedNetworkID string
+}
 
 // New creates a new Container with given options.
-func New(nsprefix string, options ...Option) (*DockerContainer, error) {
+func New(nsprefix string) (*DockerContainer, error) {
+	client, err := client.NewClientWithOpts(client.FromEnv, client.WithTimeout(clientHTTPTimeout))
```
NicolasMahe: Setting up the timeout on the whole docker client doesn't allow any build that take more than `clientHTTPTimeout` (10sec in current implementation).
Do we really need a timeout anyway?
```suggestion
	client, err := client.NewClientWithOpts(client.FromEnv)
```

container/container.go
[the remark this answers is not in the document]
krhubert: So, I have already updated the code to use cli.

About that part. This should be done in cli (like we had in previous cli). Also if you run build in cli that part should be done, otherwise server modifies service somehow and I'm not sure if that is desirable behavior.

container/container.go
krhubert: Done

NicolasMahe:
@antho1404 please do a last review and merge if everything is ok

PR #1401, opened 2019-10-09
Fix command and args fields in service protobuf

krhubert:
Fix https://github.com/mesg-foundation/engine/issues/1399
dep https://github.com/mesg-foundation/engine/pulls

- change command type to string slice
- remove `printascii` validation as the command/args is not required to have only ascii chars - the simple example is to echo Unicode chars.


antho1404:
I put a comment on the issue https://github.com/mesg-foundation/engine/issues/1399. I would like to know what this is really fixing and if it really fixes it.

PR #1402, opened 2019-10-09
Run scripts from Makefile

krhubert:
- create Makefile
- remove mesg-tools and require-mesg-tools scripts
- move build-engine to Dockerfile
- remove go-bindata from Dockefile.tools
- require changelog arg
- update .gitignore
- remove Dockerfile.dev
- move dev to scripts dir

close https://github.com/mesg-foundation/engine/issues/1393

TODO: There is no option to pass an argument to dev script right now. I need to figure it out, but it will be done with YAML config update.

I think everything now is in Makefile and it's a lot cleaner


Dockerfile
```diff
@@ -15,7 +15,8 @@ RUN go mod download
 COPY . .
 ARG version
 ENV version=${version}
-RUN ./scripts/build-engine.sh
+RUN go build -o engine -ldflags="-X 'github.com/mesg-foundation/engine/version.Version=$version'" core/main.go
```
antho1404: can't we do a `make install` or something similar

krhubert: I intentionally haven't used any make commands inside a Dockerfile or cricleci. I wanted to see how it will envolve and make that transition later. The first step is to move everything there so we can use it easily.

Makefile
```diff
@@ -0,0 +1,52 @@
+.PHONY: all docker docker-dev docker-tools dev dep build test mock protobuf changelog clean 
+
+all: clean test build docker
+
+MAJOR_VERSION := $(shell git describe --abbrev=0 | cut -d . -f 1)
+MINOR_VERSION := $(shell git describe --abbrev=0 | cut -d . -f 1-2)
+PATCH_VERSION := $(shell git describe --abbrev=0)
+DEV_VERSION := $(shell git describe)
```
antho1404: I have a version 0.14.2 with git describe where I actually have a v0.15 in my tags.
Is that something normal? I'm afraid that the CI will generate the wrong version

krhubert: Following https://git-scm.com/docs/git-describe

> The command finds the most recent tag that is reachable from a commit. 

So probably, your tag v0.15 is on a different branch and it's not reachable from the building commit. Otherwise, `git describe` shows the most recent tag.

antho1404: Even from `dev` and `master` my `git describe` shows `v0.14.2-152-gd3f71c74` but we did release v0.15. Can you try and tell me what you've got

krhubert: Yes, on `dev` branch the latest tag is `v0.14.2` because tag v0.15.0 is only on master branch.

The commit merge `9062446d7cbc049c8482f0606703e3154331a017` has that tag, and that one is only on master branch. So it is working as expected and the master release will work. 

We should use tag only in one way to have it clear:

1) use on dev branch (will be merged into master with all tags)
2) use only on master branch

For now we tag sometimes on dev, sometimes on master.

antho1404: I think this is source of problems for now and we don't use the `docker` command anyway. What do you think about just removing/commenting this for now. I'm afraid we will want to use it for the CI and this will mess up everything.

krhubert: > I think this is source of problems 

What pbms do you mean? The makefile is just to use locally (you can also build a docker image to check if everything will work on ci). The real source of the problem is the inconsistent tagging policy.

> don't use the docker command anyway.

Yep, it's there only for us to testing 

> What do you think about just removing/commenting this for now.

If you want so, I can remove, comment it, but as I said previously. I intentionally didn't use that makefile anywhere - it's for local dev only.

antho1404: We might be tempted to change the CI to use the makefile instead, to have all our scripts in one and if we have the wrong versioning that will be bad

Dockerfile
```diff
@@ -14,14 +9,14 @@ RUN go mod download
 
 COPY . .
 ARG version
-ENV version=${version}
-RUN ./scripts/build-engine.sh
+ENV PATCH_VERSION=${version}
```
NicolasMahe: Why put this env `PATCH_VERSION`? it is not used (except in makefile, is it why?)

Makefile
```diff
@@ -0,0 +1,34 @@
...
+	docker build -t mesg/tools:local -f Dockerfile.tools .
+
+dev: docker-dev
+	- ./scripts/dev.sh
+
+dep:
+	go mod download
+
+build: dep
+	go build -mod=readonly -o ./bin/engine -ldflags="-X 'github.com/mesg-foundation/engine/version.Version=$(PATCH_VERSION)'" core/main.go
```
NicolasMahe: There is a `PATCH_VERSION ` env used here but not defined..

Makefile
```diff
@@ -0,0 +1,34 @@
...
+	go test ./...
+
+mock: docker-tools
+	docker run --rm -v $(PWD):/project mesg/tools:local	./scripts/build-mocks.sh
+
+protobuf: docker-tools
+	docker run --rm -v $(PWD):/project mesg/tools:local	./scripts/build-proto.sh
+
+changelog:
+	./scripts/changelog.sh
```
NicolasMahe:
```suggestion
	./scripts/changelog.sh $(milestone)
```

.dockerignore
```diff
@@ -1,16 +1,33 @@
...
+**/*_test.go
 **/*.test
 **/debug
 **/node_modules
-bin/cli
-docs/
-docker-images/
-.tendermint-validator/
\ No newline at end of file
+/docker-images
```
NicolasMahe: ignore bin folder to avoid sending to docker previously locally build bin
```suggestion
/bin
/docker-images
```

Makefile
```diff
@@ -0,0 +1,61 @@
...
+	docker build \
+		--build-arg version=$(PATCH_VERSION) \
+		-t mesg/engine:$(MAJOR_VERSION) \
+		-t mesg/engine:$(MINOR_VERSION) \
+		-t mesg/engine:$(PATCH_VERSION) \
+		-t mesg/engine:latest \
+		.
+
+docker-dev:
+	docker build -t mesg/engine:dev --build-arg version=$(version) .
```
NicolasMahe: Why using `$(version)` without the `check-env`?
I would suggest to hardcode to `dev` the version (like before last commit):
```suggestion
	docker build -t mesg/engine:dev --build-arg version=dev .
```

Makefile
```diff
@@ -0,0 +1,61 @@
...
+	docker build -t mesg/tools:local -f Dockerfile.tools .
+
+dev: docker-dev
+	- ./scripts/dev.sh
+
+dep:
+	go mod download
+
+build: dep
+	go build -mod=readonly -o ./bin/engine -ldflags="-X 'github.com/mesg-foundation/engine/version.Version=$(version)'" core/main.go
```
NicolasMahe: Again, Why using $(version) without the check-env?
Here i would suggest to force version:
```suggestion
	check-env
	go build -mod=readonly -o ./bin/engine -ldflags="-X 'github.com/mesg-foundation/engine/version.Version=$(version)'" core/main.go
```

Makefile
```diff
@@ -0,0 +1,61 @@
...
+endif
+
+docker: check-env
+	docker build \
+		--build-arg version=$(PATCH_VERSION) \
+		-t mesg/engine:$(MAJOR_VERSION) \
+		-t mesg/engine:$(MINOR_VERSION) \
+		-t mesg/engine:$(PATCH_VERSION) \
+		-t mesg/engine:latest \
+		.
```
NicolasMahe: we could use `docker tag` command to tag an image multiple times instead of doing it during the build.
Like this, we could have a simple build command for both dev and release version ;)
https://docs.docker.com/engine/reference/commandline/tag/

Dockerfile
```diff
@@ -14,14 +9,13 @@ RUN go mod download
...
 FROM ubuntu:18.04
 RUN apt-get update && \
       apt-get install -y --no-install-recommends ca-certificates=20180409 && \
       apt-get clean && \
       rm -rf /var/lib/apt/lists/*
 WORKDIR /app
-COPY --from=build /project/engine .
-CMD ["./engine"]
+COPY --from=build /project/bin/engine .
+CMD ["./bin/engine"]
```
NicolasMahe: The binary is not in bin folder:
```suggestion
CMD [".engine"]
```

scripts/dev.sh
```diff
@@ -78,42 +75,6 @@ function docker_network_remove {
...
-DOCKER_SUM="$(openssl md5 ./Dockerfile.dev)"
-DOCKER_SUM_PREV="$(cat $DOCKER_SUM_PATH)"
-if [[ "$DOCKER_SUM" == "$DOCKER_SUM_PREV" ]]; then
-  DOCKERCACHED=true
-else
-  echo "$DOCKER_SUM" > "$DOCKER_SUM_PATH"
-fi
-
-# create mesg data folder on host machine
-mkdir -p $MESG_PATH
```
NicolasMahe: this is still important!

PR #1406, opened 2019-10-16
Wip/reading from yaml + e2e test

krhubert:
Please follow the changes and left a note how do you see it:

- configs are stored in yaml file (and rather pass as directories) then envs
- in that case we don't need docket to run it (we can use `host` network)
- from that point docker network can be easily removed
- gen-genesis script is rather use for test data (it won't be fully auto generated, but we can easily create as many users as we want and start just with config)
- there is first full real e2e test in e2e directory.
- you can run engine with cmd ` go run main.go  -config ./e2e/testdata/config/testuser/test-config.yml`
- in that case the dev script is useless (assuming we use docker host network)


e2e/account_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	t.Run("list", func(t *testing.T) {
+		resp, err := c.List(context.Background(), &pb.ListAccountRequest{})
+		require.NoError(t, err)
+
+		accs, err := kb.List()
+		require.NoError(t, err)
+
+		require.Equal(t, len(resp.Accounts), len(accs))
+	loop:
```
NicolasMahe: please no

PR #1412, opened 2019-10-22
Load genesis from file and generate it if it doesn't exist

NicolasMahe:
This PR changes the way the engine is dealing with the Cosmos genesis.

Previously, the engine reconstructed the genesis from the genesis validator transaction and a few other mandatory values that were passed to the engine as env variables.
This was making the engine hard to start (mandatory envs) and make the dev scripts doing way too much stuff.

This PR revert the reconstruction of the genesis to simply loading it from disk (already implemented by tendermint). If it doesn't exist, the engine will generate one with defined values from the config. This allows the engine to start without any mandatory envs or existing files on dev network.

The standalone script `gen-genesis` has also been updated to reflect those modification and is directly accessible from the makefile:
```
make genesis chain-id=mesg path=/genesis validators=alice,bob
```

closes https://github.com/mesg-foundation/engine/issues/1410 and closes https://github.com/mesg-foundation/engine/issues/1392


cosmos/genesis.go
```diff
@@ -0,0 +1,165 @@
...
+// Map exports the info of GenesisValidator in a nice way for logrus.
+func (validator GenesisValidator) Map() map[string]interface{} {
+	return map[string]interface{}{
+		"name":     validator.Name,
+		"address":  validator.Address,
+		"password": validator.Password,
+		"mnemonic": validator.Mnemonic,
+		"nodeID":   validator.NodeID,
+		"peer":     fmt.Sprintf("%s@%s:26656", validator.NodeID, validator.Name),
+	}
```
krhubert: Why there are all those fields in map, while only `peer` is used?

cosmos/genesis.go
```diff
@@ -0,0 +1,165 @@
+package cosmos
```
krhubert: There are no tests for this file.

config/config.go
```diff
@@ -93,6 +86,10 @@ func Default() (*Config, error) {
 	c.Tendermint.Instrumentation.Prometheus = true
 	c.Tendermint.Instrumentation.PrometheusListenAddr = "0.0.0.0:26660"
 
+	c.DevGenesis.AccountName = "engine"
+	c.DevGenesis.AccountPassword = "password"
```
krhubert:
```suggestion
	c.DevGenesis.AccountPassword = "pass"
```

config/config.go
```diff
@@ -132,6 +129,7 @@ func (c *Config) Prepare() error {
 	if err := os.MkdirAll(c.Path, os.FileMode(0755)); err != nil {
 		return err
 	}
+	// TODO: find a way to not hardcode "config" and "data" folder from tendermint
```
krhubert: Why? What is the value of changing the tendermint dir layout?

config/config.go
```diff
@@ -54,15 +48,15 @@ type Config struct {
 		Path             string `validate:"required"`
 	}
 
-	Cosmos CosmosConfig
-}
+	Cosmos struct {
+		Path string `validate:"required"`
```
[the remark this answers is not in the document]
NicolasMahe: Yes there is one in load function but it cannot be override by envs:
https://github.com/mesg-foundation/engine/pull/1412/files#diff-3baf47c64847a8fb8aaa8cc2e088513bR121

config/config.go
```diff
@@ -54,15 +48,15 @@ type Config struct {
...
+	}
 
-// CosmosConfig is the struct to hold cosmos related configs.
-type CosmosConfig struct {
-	Path               string    `validate:"required"`
-	ChainID            string    `validate:"required"`
-	GenesisTime        time.Time `validate:"required"`
-	GenesisValidatorTx StdTx     `validate:"required"`
+	DevGenesis struct {
```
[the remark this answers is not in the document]
NicolasMahe: The goal of this `DevGenesis` is a fallback to create a deterministic environment for developer with one account with the same name and password (maybe it also should have the same address this hardocode its mnemonic), same chain id.

Any other genesis should be injected in the engine as a JSON file. If you see the current logic in the `main.go`, this `DevGenesis` is only used when no genesis is present on the disk. The path to the genesis is the default tendermint and is `.mesg/tendermint/config/genesis.json`.

> how to load genesis for multiple instances
i don't understand this question.

---

Maybe the name `DevGenesis` is not good. Maybe something like `DefaultGenesisGenerator` will be more explicit?

config/config.go
NicolasMahe: I just would like to get the value from tendermint config instead of hardcoded it in our codebase.
I don't want to change it or whatever, just don't want to define it and use the default tendermint one.

cosmos/genesis.go
NicolasMahe: It's for logging when using `logrus.WithFields`. How could I do it nicely?

core/main.go
```diff
@@ -119,14 +122,37 @@ func main() {
...
+		}
+		genesis, err = cosmos.GenGenesis(app, kb, cfg.DevGenesis.ChainID, cfg.Tendermint.GenesisFile(), []cosmos.GenesisValidator{validator})
+		if err != nil {
+			logrus.WithField("module", "main").Fatalln(err)
+		}
+		logrus.WithFields(validator.Map()).Infoln("Validator")
+	} else if err != nil {
+		logrus.WithField("module", "main").Fatalln(err)
+	}
+
```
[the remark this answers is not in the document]
NicolasMahe: The problem if config is handling the loading or generation of the genesis is that config will have a huge dependency to cosmos and tendermint. Also, to generate the generate the genesis, cosmos app, keybase are required.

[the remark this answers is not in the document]
NicolasMahe: I moved this logic to a specific function in `main.go` to ease a bit the main function

config/config.go
NicolasMahe: I found a way to do it :)

cosmos/genesis.go
NicolasMahe: I would like to implement a similar function to `String()` but for logrus that can return this `map[string]interface{}`

config/config.go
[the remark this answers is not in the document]
NicolasMahe: I simplified the path of cosmos and tendermint

cosmos/genesis.go
NicolasMahe: test added 👍

cosmos/genesis.go
krhubert: Either log a struct as json or use `WithFields`  func

```
  log.WithFields(log.Fields{
	"name":     validator.Name,
		"address":  validator.Address,
		"password": validator.Password,
		"mnemonic": validator.Mnemonic,
		"nodeID":   validator.NodeID,
		"peer":     fmt.Sprintf("%s@%s:26656", validator.NodeID, validator.Name),
  }).Info("A walrus appears")
```

Instead of creating a dedicated method for return `map`

core/main.go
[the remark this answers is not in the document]
krhubert: I see. I thought that it will need only the tendermint and cosmos package (not the one created by us), but for generate genesis validator, it won't need any extra dependencies. Only to generate the app state. 

So I'm ok to keep validator generation in cosmos (although I would put that in config :P),
but what do you think about moving that function cosmos package itself. 

- I want to keep main.go as small as possible. 
- This logic doesn't need any data from main except config.
- in cosmos package that function could be tested easily.

config/config.go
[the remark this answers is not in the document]
krhubert: > how to load genesis for multiple instances

What I mean I want to start >2 engine instances on a test environment. How to do that with the current config.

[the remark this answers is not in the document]
NicolasMahe: > What I mean I want to start >2 engine instances on a test environment. How to do that with the current config.

there is 2 ways to do:

1. Start a first engine. Copy its genesis and put it in the config path of the second engine. start second engine.
2. Generate new genesis. Copy it in both the first and second engine config path.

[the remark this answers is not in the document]
krhubert: So in both cases users need to be aware of soem internal configuration. I want to avoid such behaviour, do you think it's possible?

antho1404:
I have an issue running the engine now.
```
FATA[0000] config.Cosmos.GenesisValidatorTx error: no signers 
```

Shouldn't we be able to run the engine with no pre-configuration?

Even if I create a genesis with `make genesis chain-id=mesg path=~/.mesg validators=alice,bob` I still get the same error

PR #1430, opened 2019-10-25
Set block creation time to 1s instead of 10s

NicolasMahe:
Set block creation time to 1s instead of 10s


config/config.go
```diff
@@ -88,7 +88,7 @@ func defaultConfig() (*Config, error) {
 	c.Tendermint.Config.RPC.ListenAddress = "tcp://0.0.0.0:26657"
 	c.Tendermint.Config.P2P.AddrBookStrict = false
 	c.Tendermint.Config.P2P.AllowDuplicateIP = true
-	c.Tendermint.Config.Consensus.TimeoutCommit = 10 * time.Second
+	c.Tendermint.Config.Consensus.TimeoutCommit = 1 * time.Second
```
antho1404: can simply be removed as the default is `TimeoutCommit:               1000 * time.Millisecond,`

PR #1438, opened 2019-10-26
Update golangci-lint v1.21

NicolasMahe:
Changelog: https://github.com/golangci/golangci-lint/releases

I think we should deactivate `wsl`, setup `gocognit` properly, understand what `stylecheck` is saying, remove `TODO` from the codebase or maybe (if possible) display them as warning instead of error.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ jobs:
 
   "lint":
     docker:
-      - image: golangci/golangci-lint:v1.18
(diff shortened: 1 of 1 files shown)


krhubert:
Watch out, as I've already tried update golang lint but it gets killed on circleci machine. I've tried to debug it but without success;

krhubert:
clsoing in favor of https://github.com/mesg-foundation/engine/pull/1474

PR #1440, opened 2019-10-28
e2e service test

### LINKED ISSUE #1421: Create e2e tests for service api
Dependant on https://github.com/mesg-foundation/engine/issues/1419

Create e2e tests for the service api with a pre-compiled service and the dev account:
- [ ] create
- [ ] get
- [ ] list
- [ ] exists
- [ ] hash

krhubert:
close #1421 

dep https://github.com/mesg-foundation/engine/pull/1439


e2e/service_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/mesg-foundation/engine/hash"
+	pb "github.com/mesg-foundation/engine/protobuf/api"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/metadata"
+)
+
+var (
+	testServiceHash  hash.Hash
+	testInstanceHash hash.Hash
```
NicolasMahe:
```suggestion
//	testInstanceHash hash.Hash
```

this break the test thus no e2e test are run on CI

Makefile
```diff
@@ -53,6 +53,10 @@ build: check-version dep
 
 e2e: export MESG_PATH = $(PWD)/e2e.test/mesg
 e2e: clean-e2e
+	# setup - copy config to MESG_PATH
```
NicolasMahe: what about a new command to compile the test service using the mesg-cli? It's not a really nice solution but it prevents to test against an old test

e2e/service_test.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+func testService(t *testing.T) {
+	req := readCreateServiceRequest("testdata/test-service/test-service.json")
```
NicolasMahe: path is wrong
```suggestion
	req := readCreateServiceRequest("testdata/test-service.json")
```

Makefile
krhubert: Yes I can add the new target, but let's not do that in this PR. 

Btw, It won't prevent testing old service if someone forgets to run it. We can do this automatically (by depended targets) and get rid of `test-service.json` from repo, but in that case we have to prepare docker with  mesg-cli to be used in cirlceci.

e2e/testdata/test-service.json
```diff
@@ -0,0 +1,85 @@
+{
+  "sid": "test-service",
+  "name": "test-service",
+  "configuration": {},
```
antho1404: It could be good to add a service with envs, volumes, dependencies, etc...

I'm using this service for my tests, I have volumes, volumesFrom, envs, dependencies, env override and test of all that when the service starts
```json
{
  "sid": "testservice",
  "name": "test service",
  "configuration": {
    "volumes": [
      "/volume/test/"
    ],
    "volumesFrom": [
      "nginx"
    ],
    "env": [
      "ENVA=do_not_override",
      "ENVB=override"
    ]
  },
  "dependencies": [
    {
      "image": "nginx",
      "volumes": [
        "/etc/nginx"
      ],
      "key": "nginx"
    }
  ],
  "tasks": [
    {
      "key": "taskx",
      "inputs": [
        {
          "key": "foo",
          "type": "String",
          "object": []
        },
        {
          "key": "bar",
          "type": "String",
          "object": []
        }
      ],
      "outputs": [
        {
          "key": "result",
          "type": "String",
          "object": []
        }
      ]
    }
  ],
  "events": [
    {
      "key": "event",
      "data": [
        {
          "key": "x",
          "type": "Any",
          "object": []
        }
      ]
    }
  ],
  "source": "QmbNWV7UV4hHj62SVjFEWyAroSBepgDMWg45x6CfkqbLfb"
}
```

e2e/service_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+		testServiceHash = resp.Hash
+	})
+
+	t.Run("get", func(t *testing.T) {
+		ctx := metadata.NewOutgoingContext(context.Background(), passmd)
+
+		service, err := client.ServiceClient.Get(ctx, &pb.GetServiceRequest{Hash: testServiceHash})
+		require.NoError(t, err)
+		require.Equal(t, testServiceHash, service.Hash)
```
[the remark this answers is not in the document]
krhubert: Maybe with another PR. The problem is the [equal](https://godoc.org/github.com/gogo/protobuf/plugin/equal) option was removed and there is no quick way to do that. 

I must prepare another PR with `equal` set on + exported function to convert `CreateServiceRequest` to `Service`

e2e/testdata/test-service.json
krhubert: This could be the next step. For now on I just wanted to test real simple service.

Makefile
NicolasMahe: we could use `npx mesg-cli ....` that download mesg-cli if it not installed ;)

e2e/testdata/test-service.json
NicolasMahe: Created related issue https://github.com/mesg-foundation/engine/issues/1446

e2e/service_test.go
```diff
@@ -0,0 +1,75 @@
...
+	})
+
+	t.Run("list", func(t *testing.T) {
+		resp, err := client.ServiceClient.List(context.Background(), &pb.ListServiceRequest{})
+		require.NoError(t, err)
+		require.Len(t, resp.Services, 1)
```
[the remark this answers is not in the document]
NicolasMahe: Created related issue https://github.com/mesg-foundation/engine/issues/1446

e2e/service_test.go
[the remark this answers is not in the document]
NicolasMahe: Created related issue https://github.com/mesg-foundation/engine/issues/1446

PR #1443, opened 2019-10-28
e2e event test

### LINKED ISSUE #1426: Create e2e tests for event api
Dependant on https://github.com/mesg-foundation/engine/issues/1423

Create e2e tests for the event api with a running instance:
- [ ] create (done by instance)
- [ ] stream

For this tests, i suggest the following:
1. starts a stream to listen for event
2. creates an execution for the instance that will actually emit an event
3. waits for an event to pop on the stream
4. checks the data

krhubert:
dep https://github.com/mesg-foundation/engine/pull/1442
close #1426

### DIFF
--- a/e2e/event_test.go
+++ b/e2e/event_test.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/mesg-foundation/engine/protobuf/acknowledgement"
 	pb "github.com/mesg-foundation/engine/protobuf/api"
+	"github.com/mesg-foundation/engine/protobuf/types"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


e2e/event_test.go
```diff
@@ -0,0 +1,31 @@
...
+		Key:          "ping_ok",
+	})
+	require.NoError(t, err)
+
+	event, err := stream.Recv()
+	require.NoError(t, err)
+
+	require.Equal(t, resp.Hash, event.Hash)
+	require.Equal(t, testInstanceHash, event.InstanceHash)
+	require.Equal(t, "ping_ok", event.Key)
```
NicolasMahe: Can you add a check on the event's data?

krhubert: Can you remind me why have you removed `compare` opinion for gogo.protobuf types?

NicolasMahe: i did? i think we should have those helpers (compare, equal, etc..)

NicolasMahe: actually, only process, ownership and execution have `option (gogoproto.equal_all) = true;`
we can definitely set this on all proto

PR #1444, opened 2019-10-28
e2e execution test

### LINKED ISSUE #1423: Create e2e tests for execution api
Dependant on https://github.com/mesg-foundation/engine/issues/1422

Create e2e tests for the execution api with a running instance:
- [ ] create
- [ ] get
- [ ] stream
- [ ] update (done by the instance)

For this tests, i suggest the following:
1. starts a stream to listen for execution in all status
2. creates an execution for the instance
3. waits for an execution to pop on the stream
4. checks the data (eg: status = InProgress)
5. the service receive the execution and call the update execution api with data
6. waits for an execution to pop on the stream
7. checks the data (eg: status = Completed)
8. fetch the execution using the get api
7. checks the data are the same

another test for failed execution:
1. creates an execution for the instance to trigger a task that will return an error
2. the service receive the execution and call the update execution api with an error
6. waits for an execution to pop on the stream
7. checks the data (eg: status = Failed)
8. fetch the execution using the get api
7. checks the data are the same

krhubert:
dep https://github.com/mesg-foundation/engine/pull/1443
close #1423

### DIFF
--- a/e2e/execution_test.go
+++ b/e2e/execution_test.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"context"
+	"testing"
+
+	"github.com/mesg-foundation/engine/execution"
+	"github.com/mesg-foundation/engine/protobuf/acknowledgement"
+	pb "github.com/mesg-foundation/engine/protobuf/api"
+	"github.com/mesg-foundation/engine/protobuf/types"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/metadata"
+)
+
+func testExecution(t *testing.T) {
+	stream, err := client.ExecutionClient.Stream(context.Background(), &pb.StreamExecutionRequest{
+		Filter: &pb.StreamExecutionRequest_Filter{
+			InstanceHash: testInstanceHash,
+		},
+	})
+	require.NoError(t, err)
+	acknowledgement.WaitForStreamToBeReady(stream)
+
+	ctx := metadata.NewOutgoingContext(context.Background(), passmd)
+	resp, err := client.ExecutionClient.Create(ctx, &pb.CreateExecutionRequest{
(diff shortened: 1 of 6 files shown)


e2e/execution_test.go
```diff
@@ -0,0 +1,62 @@
...
+					Kind: &types.Value_StringValue{
+						StringValue: "test",
+					},
+				},
+			},
+		},
+	})
+	require.NoError(t, err)
+
+	fmt.Println("recives execution")
```
NicolasMahe: typo `recive` many times

NicolasMahe:
When running the e2e tests, the service logs are only (even when the first execution listener test passed
```
2019/10/29 05:30:55 main.go:75: connect to engine:50052
2019/10/29 05:30:55 main.go:86: create a stream
```
```
=== RUN   TestAPI/execution
recives execution
recives second execution
```

when running in `service:dev`, i get:
```
test-service/service: 2019/10/29 05:23:59 main.go:75: connect to engine:50052
test-service/service: 2019/10/29 05:23:59 main.go:86: create a stream
test-service/service: 2019/10/29 05:24:21 main.go:93: recive exeuction ping 9Fxuu5s9pGisLLvhZ49NiQkVc4xWcBnAi6Cq1iZjAqG
```

So maybe the execution is never received by the service?!? the service does not restart and there is no error log. weird problem

e2e/execution_test.go
krhubert: Yes, one is suppose to catch `in_progres` status, the second one for `complete`, but becaous service never recived execution, there is no chance to update it.

krhubert:
> So maybe the execution is never received by the service?!?

Yes, that's what I wrote yesterday. I've added the logs and the execution is published to pubsub but never sent to service :/

PR #1449, opened 2019-10-30
Add config.log.format validation

krhubert:
close https://github.com/mesg-foundation/engine/issues/1387

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -35,7 +35,7 @@ type Config struct {
 	}
 
 	Log struct {
-		Format      string `validate:"required"`
+		Format      string `validate:"required,oneof=json text"`
 		ForceColors bool
 		Level       string `validate:"required"`
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -7,7 +7,6 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/mesg-foundation/engine/x/xstrings"
 	homedir "github.com/mitchellh/go-homedir"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -35,7 +35,7 @@ type Config struct {
 	}
 
 	Log struct {
-		Format      string `validate:"required"`
+		Format      string `validate:"required,oneof=json text"`
 		ForceColors bool
 		Level       string `validate:"required"`
```
antho1404: Should we do that for the log level too?

NicolasMahe: No because logrus already have a function that check it and it is used in the `config.validate` function.

NicolasMahe:
@krhubert Don't forget to remove the now useless test lines L163-R165:
https://github.com/mesg-foundation/engine/blob/43514a9e1c01b21c77c98deb74c12c5cfb3ef2ab/config/config.go#L163-L165

config/config.go
antho1404: So we could remove the `logrus.ParseLevel` test and just keep the test on the config like that we have consistent errors.

krhubert: So I will change the error format then to keep it consistent, ok for you? :)

krhubert: Or maybe not, the validation errors are more just not so good. 

`Key: 'Config.Log.Format' Error:Field validation for 'Format' failed on the 'oneof' tag`

I would rather change the validation error format to make it more human friendly.

antho1404: hum ok, just keep it like it is it's fine

PR #1450, opened 2019-10-30
Add more explicit error on service create

krhubert:
close https://github.com/mesg-foundation/engine/issues/1390


sdk/service/sdk.go
```diff
@@ -44,6 +47,9 @@ func (s *SDK) Create(req *api.CreateServiceRequest, accountName, accountPassword
 	msg := newMsgCreateService(s.cdc, req, owner)
 	tx, err := s.client.BuildAndBroadcastMsg(msg, accountName, accountPassword, accNumber, accSeq)
 	if err != nil {
+		if err == mempool.ErrTxInCache {
```
NicolasMahe: I don't like to check this error. This error could never happen if the `accSeq` changes.
What about a pre-check on that calculate the hash and check if it already exists?

krhubert: >This error could never happen if the accSeq changes.

Ąre you sure?

I've made a test and error is still the same if `accSeq` changes.

 I also think that we should leave that check to network, as the network will validate that for us better (for example it handles the race condition when one service is created at the same time more than once).

So that check won't give us anserw if the service exists or not. Engine have to still handle that error case.

sdk/service/sdk.go
```diff
@@ -44,6 +47,9 @@ func (s *SDK) Create(req *api.CreateServiceRequest, accountName, accountPassword
...
+			return nil, errors.New("service already exists")
```
NicolasMahe: what about wrapping the original error? I would like keep it to not forget it in the future. I see this as a small hack to help user understand what is happening. But as a developer, i want to know the tx already exists.
```suggestion
			return nil, fmt.Error("service already exists: %w", err)
```

PR #1451, opened 2019-10-30
Validate service nested object params

krhubert:
close https://github.com/mesg-foundation/engine/issues/1088


NicolasMahe:
Can you add unit test for this function?
Will be perfect if you can add unit test on all the functions of `service/validator/validator.go`. There is none for now..

krhubert:
No, I don't want to add that for two reasons:

- I just want to keep this PR as small as possible

and what's more important:

- validation needs a bit of refactoring first so I can add the good unit tests.

Based on that we can open PR about refactor + add unit test if we want to.

PR #1452, opened 2019-10-30
Add EnvHash to instance and calculate the hash based on it

krhubert:
close https://github.com/mesg-foundation/engine/issues/1279

It was so easy to test that with e2e!


NicolasMahe:
> It was so easy to test that with e2e!

Speaking of which, could you add a check in e2e test about the env hash calculation?

sdk/instance/instance.go
```diff
@@ -102,27 +102,25 @@ func (i *Instance) Create(serviceHash hash.Hash, env []string) (*instance.Instan
...
 	instanceEnv := xos.EnvMergeMaps(xos.EnvSliceToMap(srv.Configuration.Env), xos.EnvSliceToMap(env))
 
 	// calculate instance's hash.
-	h := hash.New()
-	h.Write(srv.Hash)
-	h.Write([]byte(xos.EnvMapToString(instanceEnv)))
-	instanceHash := h.Sum(nil)
+	inst := &instance.Instance{
+		ServiceHash: srv.Hash,
+		EnvHash:     hash.Dump(instanceEnv),
```
NicolasMahe: Does the hash is the same when the order of the elements of `instanceEnv` is not the same?
The function `xos.EnvMapToSlice` or `xos. EnvMapToString` could be used as it sorts the elements.

NicolasMahe: And of course with test please!

krhubert:
I've added a transaction error log

`+ return nil, fmt.Errorf("an error occurred in transaction: %s", data.TxResult.Result.Log)`

Now the cli outputs:

```
Create service... !
 ›   Error: 2 UNKNOWN: an error occurred in transaction: 
 ›   [{"msg_index":0,"success":false,"log":"{\"codespace\":\"sdk\",\"code\":1,\"message\":\"Service.Configuration.Env[0] must be a valid 
 ›   env variable name\\nService.Configuration.Env[1] must be a valid env variable 
 ›   name\"}","events":[{"type":"message","attributes":[{"key":"action","value":"create_service"}]}]}]
```

I know it's not prefect error, but gives you context. Previously was:

```
Create service... !
 ›   Error: 2 UNKNOWN: an error occurred in transaction: 
```

And there was clue what's might going on.


========

Now envs are stable sort.

========

I've updated the validation error for env (thanks e2e tests ❤️ )

========

Add e2e test for env hash.

sdk/instance/instance.go
NicolasMahe: The env should be sorted in order to create a deterministic hash
```suggestion
		EnvHash:     hash.Dump(sort.Stable(sort.StringSlice(instanceEnv))),
```

e2e/instance_test.go
```diff
@@ -38,6 +41,7 @@ func testInstance(t *testing.T) {
 		require.NoError(t, err)
 		require.Equal(t, testInstanceHash, resp.Hash)
 		require.Equal(t, testServiceHash, resp.ServiceHash)
+		require.Equal(t, hash.Dump([]string{"BAR=2", "FOO=1", "REQUIRED", "BAR=3", "REQUIRED=4"}), resp.EnvHash)
```
NicolasMahe: this test should be replaced by:
```suggestion
		require.Equal(t, hash.Dump([]string{"BAR=3", "FOO=1", "REQUIRED=4"}), resp.EnvHash)
```
we don't want multiple env with the same key.

sdk/instance/instance.go
krhubert: It's already done in xos Package

sdk/instance/instance.go
```diff
@@ -102,32 +102,30 @@ func (i *Instance) Create(serviceHash hash.Hash, env []string) (*instance.Instan
 	if err != nil {
 		return nil, err
 	}
+
 	// calculate the final env vars by overwriting user defined one's with defaults.
-	instanceEnv := xos.EnvMergeMaps(xos.EnvSliceToMap(srv.Configuration.Env), xos.EnvSliceToMap(env))
+	instanceEnv := xos.EnvMergeSlices(srv.Configuration.Env, env)
```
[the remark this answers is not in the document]
krhubert: All functions were removed and only one remains in `xos` package. I would still keep it there because it's prefectly suits there.

PR #1459, opened 2019-11-04
Single cosmos account

NicolasMahe:
Simplify the account management to use only one account across the Engine that can be configured in Config. Rename the generated account to `engine`

closes https://github.com/mesg-foundation/engine/issues/1458


core/main.go
```diff
@@ -76,15 +76,31 @@ func stopRunningServices(sdk *enginesdk.SDK) error {
 	return errs.ErrorOrNil()
 }
 
+func loadOrGenConfigAccount(kb *cosmos.Keybase, cfg *config.Config) (keys.Info, error) {
```
krhubert: this could be in engine keybase package to extend cosmos keybase.

cosmos/keybase.go
```diff
@@ -23,11 +24,23 @@ func NewKeybase(dir string) (*Keybase, error) {
...
+func (kb *Keybase) Exist(name string) (bool, error) {
+	_, err := kb.Get(name)
+	if keyerror.IsErrKeyNotFound(err) {
+		return false, nil
+	}
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
```
krhubert: return !keyerror.IsErrKeyNotFound(err), err

or just remove the wrapper

server/grpc/api/service.go
```diff
@@ -11,19 +12,21 @@ import (
...
 func (s *ServiceServer) Create(ctx context.Context, req *protobuf_api.CreateServiceRequest) (*protobuf_api.CreateServiceResponse, error) {
-	credUsername, credPassphrase, err := GetCredentialFromContext(ctx)
-	if err != nil {
-		return nil, err
-	}
+	// credUsername, credPassphrase, err := GetCredentialFromContext(ctx)
+	// if err != nil {
+	// 	return nil, err
+	// }
+	credUsername, credPassphrase := s.cfg.Account.Name, s.cfg.Account.Password
```
krhubert: Hum, what? Do we have only one user here? I don't know what purpose is this for, but it looks like step back.

core/main.go
NicolasMahe: All the function you mean?
I already put the `Exist` function in `keybase`.
I don't think we will have this logic of `loadOrGen` in any other place.

server/grpc/api/service.go
NicolasMahe: Yes, only one account will be used until we implement a full account management with role or user authorization.
See comment in issue <https://github.com/mesg-foundation/engine/issues/1458>.

cosmos/keybase.go
NicolasMahe: You suggestion doesn't work as `err` still contain the `IsErrKeyNotFound`.
The best I can do is:
```go
	exist := !keyerror.IsErrKeyNotFound(err)
	if err != nil && exist {
		return exist, err
	}
	return exist, nil
```
and it's not so much better.

core/main.go
krhubert: > All the function you mean?

Yes

>  I don't think we will have this logic of loadOrGen in any other place.

And that's fine. The Keybase wrapper was created to handle such cases. so you don't put them in other places (like main.go). 

```go
func (kb *Keybase) LoadOrGenAccount(name, password string) (keys.Info, error) { 
}
```

core/main.go
```diff
@@ -76,15 +76,31 @@ func stopRunningServices(sdk *enginesdk.SDK) error {
...
+	exist, err := kb.Exist(cfg.Account.Name)
+	if err != nil {
+		return nil, err
+	}
+	if exist {
+		return nil, nil
+	}
+	logrus.WithField("module", "main").Warn("Config account not found. Generating one for development...")
+	mnemonic, err := kb.NewMnemonic()
```
antho1404: We should have the mnemonic in the config (empty by default) and if empty generate a new one, otherwise use the one in the config.

This will let us start the engine with a pre-determined address and we don't need to rely on a generated address.

internal/tools/gen-genesis/main.go
```diff
@@ -98,9 +106,9 @@ func main() {
 		peer := fmt.Sprintf("%s@%s:26656", genVal.NodeID, genVal.Name)
 		logrus.WithFields(map[string]interface{}{
 			"name":     genVal.Name,
-			"address":  genVal.Address,
+			"address":  acc.GetAddress().String,
 			"password": genVal.Password,
-			"mnemonic": genVal.Mnemonic,
+			"mnemonic": mnemonic,
```
antho1404: you are generating the same account (identical mnemonic) just with different names. Each validator should have a different mnemonic, the same as the password

server/grpc/api/service.go
```diff
@@ -11,19 +12,21 @@ import (
...
+	return &ServiceServer{sdk: sdk, cfg: cfg}
 }
 
 // Create creates a new service from definition.
 func (s *ServiceServer) Create(ctx context.Context, req *protobuf_api.CreateServiceRequest) (*protobuf_api.CreateServiceResponse, error) {
-	credUsername, credPassphrase, err := GetCredentialFromContext(ctx)
-	if err != nil {
-		return nil, err
-	}
+	// credUsername, credPassphrase, err := GetCredentialFromContext(ctx)
```
antho1404: Is it too complex to have it in the context and we just override the context with an interceptor?
This way we still use that credentials from context and when we migrate to multi-account we just need to remove the interceptor.

This way the serviceserver will not have any knowledge about the config too.

internal/tools/gen-genesis/main.go
NicolasMahe: No, it's not. There is a `for` line 78.

antho1404: my bad, I was confused by the passwords that are managed in a different loop

core/main.go
NicolasMahe: issue created https://github.com/mesg-foundation/engine/issues/1470

core/main.go
NicolasMahe: This is not compatible with Anthony's suggestion https://github.com/mesg-foundation/engine/pull/1459/files#r342436168

PR #1462, opened 2019-11-04
Use sequence number from cosmos auth module

krhubert:
close https://github.com/mesg-foundation/engine/issues/1334


antho1404:
Put a commit on the issue https://github.com/mesg-foundation/engine/issues/1334#issuecomment-549664134, there is a better way to do that fix

cosmos/client.go
```diff
@@ -47,8 +52,19 @@ func (c *Client) Query(path string, data cmn.HexBytes, ptr interface{}) error {
 }
 
 // BuildAndBroadcastMsg builds and signs message and broadcast it to node.
-func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg, accName, accPassword string, accNumber, accSeq uint64) (*abci.ResponseDeliverTx, error) {
-	txBuilder := NewTxBuilder(c.cdc, accNumber, accSeq, c.kb, c.chainID)
+func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg, accName, accPassword string, accNumber uint64) (*abci.ResponseDeliverTx, error) {
```
NicolasMahe: `accNumber ` can be removed :)
```suggestion
func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg, accName, accPassword string) (*abci.ResponseDeliverTx, error) {
```

sdk/service/sdk.go
```diff
@@ -39,13 +39,13 @@ func (s *SDK) Create(req *api.CreateServiceRequest, accountName, accountPassword
 		return nil, err
 	}
 	// TODO: pass account by parameters
-	accNumber, accSeq := uint64(0), uint64(0)
+	accNumber := uint64(0)
```
NicolasMahe: can be removed

sdk/service/sdk.go
```diff
@@ -39,13 +39,13 @@ func (s *SDK) Create(req *api.CreateServiceRequest, accountName, accountPassword
...
 	owner, err := cosmostypes.AccAddressFromBech32(account.Address)
 	if err != nil {
 		return nil, err
 	}
 	msg := newMsgCreateService(s.cdc, req, owner)
-	tx, err := s.client.BuildAndBroadcastMsg(msg, accountName, accountPassword, accNumber, accSeq)
+	tx, err := s.client.BuildAndBroadcastMsg(msg, accountName, accountPassword, accNumber)
```
NicolasMahe: `accNumber ` can be removed :)
```suggestion
	tx, err := s.client.BuildAndBroadcastMsg(msg, accountName, accountPassword)
```

cosmos/client.go
```diff
@@ -24,6 +25,10 @@ type Client struct {
 	chainID string
 }
 
+// AccSeq is a sequence number that could be used to build the transaction.
+// It should be increased after every usage.
+var AccSeq uint64
```
antho1404: Now useless or am I missing something?

cosmos/client.go
```diff
@@ -94,3 +110,29 @@ func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg, accName, accPassword str
...
+func (c *Client) QueryWithData(path string, data []byte) ([]byte, int64, error) {
+	return c.query(path, data)
+}
+
+// query performs a query to a Tendermint node with the provided store name
+// and path. It returns the result and height of the query upon success
+// or an error if the query fails. In addition, it will verify the returned
+// proof if TrustNode is disabled. If proof verification fails or the query
+// height is invalid, an error will be returned.
+func (c *Client) query(path string, key cmn.HexBytes) (res []byte, height int64, err error) {
```
antho1404: Kind of duplicated with `Query`, maybe we should have `Query` that uses `query`

cosmos/client.go
```diff
@@ -47,8 +52,19 @@ func (c *Client) Query(path string, data cmn.HexBytes, ptr interface{}) error {
...
 // BuildAndBroadcastMsg builds and signs message and broadcast it to node.
-func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg, accName, accPassword string, accNumber, accSeq uint64) (*abci.ResponseDeliverTx, error) {
-	txBuilder := NewTxBuilder(c.cdc, accNumber, accSeq, c.kb, c.chainID)
+func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg, accName, accPassword string, accNumber uint64) (*abci.ResponseDeliverTx, error) {
+	info, err := c.kb.Get(accName)
+	if err != nil {
+		return nil, err
+	}
+
+	acc, err := auth.NewAccountRetriever(c).GetAccount(info.GetAddress())
```
antho1404: could use `GetAccountNumberSequence` instead

antho1404:
The sequence is always 0, this is because the auth `antehandler` is never called. There is an issue with the initialization of the cosmos app. Also because we generate the genesis in dev mode and send one transaction, the initial sequence should be incremented and be 1.

I'm already working a way to fix that, I can put the fix on this PR and make sure it actually works

cosmos/client.go
krhubert: The `Query` is needed here because `NewAccountRetriver` expects [NodeQuerier](https://github.com/cosmos/cosmos-sdk/blob/master/x/auth/types/account_retriever.go#L11)

Check out https://github.com/cosmos/cosmos-sdk/blob/master/client/context/query.go#L41 for Query implementation. I left `query` because probably at some point we will need `QueryWitData` as well.

cosmos/client.go
krhubert: I didn't spot that method. Of course we can use.

PR #1463, opened 2019-11-04
Execution SDK on Cosmos

krhubert:
This PR moves the Execution SDK on Cosmos 🎊 

Dependent on https://github.com/mesg-foundation/engine/pull/1509 and https://github.com/mesg-foundation/engine/pull/1512

You can check the commits after 499b03b84cf150f0dfe6a72947da81446a6a2f27:
https://github.com/mesg-foundation/engine/pull/1463/files/499b03b84cf150f0dfe6a72947da81446a6a2f27..02e26b848ab0388a8957b359c1bcd8c24b11f88c


database/execution_db_test.go
```diff
@@ -1,67 +1,26 @@
...
-		}
-		require.NoError(t, err)
-		require.NotNil(t, execution)
-		e, err := db.Find(execution.Hash)
-		require.NoError(t, err)
-		require.NotNil(t, e)
-	}
-}
+func TestExecutionDB(t *testing.T) {
+	db := NewExecutionDB(store.NewCommitMultiStore(dbm.NewMemDB()).(rootmulti.Store).GetStore(), codec.New())
```
NicolasMahe: Why not using the levelDB store for now?
Like service:
```go
store, err := store.NewLevelDBStore(testdbname)
require.NoError(t, err)
db := NewServiceDB(store, codec.New())```

Comment (krhubert) [reply database/execution_db_test.go]: I wanted to create cosmos store to get rid of leveldb in testing. In that case we can remove it completly once we switch to cosmos db.

Comment (NicolasMahe) [reply database/execution_db_test.go]: Look at cosmos_test in store. It already create a simple store:
```go
func newCosmosStore(t *testing.T) (*CosmosStore, func()) {
	store := NewCosmosStore(transient.NewStore())
	return store, func() {
		require.NoError(t, store.Close())
	}
}
```

NicolasMahe:
@krhubert please check the cosmos stream function, it's completely related with https://github.com/mesg-foundation/engine/pull/1509.
I'm ok if you think it easier to combine https://github.com/mesg-foundation/engine/pull/1509 in this PR to make proper implementation of stream.

cosmos/client.go
```diff
@@ -129,3 +131,34 @@ func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg, accName, accPassword str
...
+	if err != nil {
+		return nil, err
+	}
+
+	hashC := make(chan hash.Hash)
+	go func() {
+		for event := range eventStream {
+			ressHashes := event.Events[EventHashType]
+			if len(ressHashes) != 1 {
+				// or panic(err) - grpc api do not support
```
NicolasMahe: could we return those error on the chan?
either with a new chan or a struct that contain both the hash and the error.

orchestrator/orchestrator.go
```diff
@@ -28,18 +33,26 @@ func (s *Orchestrator) Start() error {
...
 		select {
 		case event := <-s.eventStream.C:
 			go s.execute(s.eventFilter(event), nil, event, event.Data)
-		case execution := <-s.executionStream.C:
+		case execution := <-s.executionStream:
 			go s.execute(s.resultFilter(execution), execution, nil, execution.Outputs)
 			go s.execute(s.dependencyFilter(execution), execution, nil, execution.Outputs)
 		}
 	}
+	close(executionStream)
```
NicolasMahe: the read should never close the chan

sdk/execution/sdk.go
```diff
@@ -0,0 +1,120 @@
...
+	var executions []*execution.Execution
+	if err := s.client.Query("custom/"+backendName+"/list", nil, &executions); err != nil {
+		return nil, err
+	}
+	return executions, nil
+}
+
+// Stream returns execution that matches given hash.
+func (s *SDK) Stream(req *api.StreamExecutionRequest) (chan *execution.Execution, error) {
+	if err := req.Filter.Validate(); err != nil {
```
NicolasMahe: This new validate function doesn't contains any validation of the presence of the data in the databases.
Here is the `validateFilter` that you replaced:
```go
func (s *SDK) validateFilter(f *api.StreamExecutionRequest_Filter) error {	
	if f == nil {	
		return nil	
	}	
	var err error	
	var run *runner.Runner	
	if !f.ExecutorHash.IsZero() {	
		if run, err = s.runnerSDK.Get(f.ExecutorHash); err != nil {	
			return err	
		}	
	}	
	var inst *instance.Instance	
	if !f.InstanceHash.IsZero() {	
		if inst, err = s.instanceSDK.Get(f.InstanceHash); err != nil {	
			return err	
		}	
	}	
	if (f.TaskKey == "" || f.TaskKey == "*") || (inst == nil && run == nil) {	
		return nil	
	}	
	// check task key if at least instance or runner is set	
	if inst == nil && run != nil {	
		inst, err = s.instanceSDK.Get(run.InstanceHash)	
		if err != nil {	
			return err	
		}	
	}	
	srv, err := s.serviceSDK.Get(inst.ServiceHash)	
	if err != nil {	
		return err	
	}	
	if _, err := srv.GetTask(f.TaskKey); err != nil {	
		return err	
	}	
	return nil	
}
```

cosmos/module.go
```diff
@@ -27,12 +28,15 @@ type AppModuleBasic struct {
 // AppModule is a main element of an cosmos app.
 type AppModule struct {
 	AppModuleBasic
-	handler sdk.Handler
+	handler Handler
 	querier Querier
 }
 
+// Handler defines the core of the state transition function of an application.
+type Handler func(request cosmostypes.Request, msg cosmostypes.Msg) (hash.Hash, cosmostypes.Error)
```
NicolasMahe: Why forcing the type of the error?
There is no need to force the type. The handler function can just wrap the error to an cosmos error if needed. See https://github.com/mesg-foundation/engine/pull/1463/files/a54e6902fede8ef8ec23c70df1791243581d9c4f#diff-472d4db73a8fc27ef1b87e0e6c294afcR109-R114

```suggestion
type Handler func(request cosmostypes.Request, msg cosmostypes.Msg) (hash.Hash, error)
```

cosmos/type.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"fmt"
+
+	cosmostypes "github.com/cosmos/cosmos-sdk/types"
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
+)
+
+// common attribute keys.
+const (
+	AttributeKeyHash = "hash"
```
NicolasMahe: Why removing the prefix event?
Everything in this file is prefix with `Event`
```suggestion
	EventAttributeKeyHash = "hash"
```

protobuf/api/execution.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if !f.ExecutorHash.Valid() {
+		return fmt.Errorf("stream filter: executor hash is invalid")
+	}
+
+	if !f.InstanceHash.Valid() {
+		return fmt.Errorf("stream filter: instance hash is invalid")
+	}
+
+	// TODO: add validation (after adding in protobuf with print ascii)
```
NicolasMahe: See comment https://github.com/mesg-foundation/engine/pull/1463#discussion_r348413155

sdk/execution/sdk.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	execC := make(chan *execution.Execution)
+	go func() {
+		for hash := range stream {
+			exec, err := s.Get(hash)
+			if err != nil {
+				// or panic(err) - grpc api do not support
```
NicolasMahe: same comment as https://github.com/mesg-foundation/engine/pull/1463#discussion_r348411298

server/grpc/api/execution.go
```diff
@@ -48,48 +42,45 @@ func (s *ExecutionServer) Get(ctx context.Context, req *api.GetExecutionRequest)
...
-			InstanceHash: req.Filter.InstanceHash,
-			Statuses:     req.Filter.Statuses,
-			Tags:         req.Filter.Tags,
-			TaskKey:      req.Filter.TaskKey,
-		}
+	stream, err := s.sdk.Execution.Stream(req)
+	if err != nil {
+		return err
 	}
+	defer close(stream)
```
NicolasMahe: the caller should not close the stream

cosmos/client.go
```diff
@@ -129,3 +131,34 @@ func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg, accName, accPassword str
...
+				continue
+			}
+
+			hash, err := hash.Decode(ressHashes[0])
+			if err != nil {
+				continue
+			}
+			hashC <- hash
+		}
+		c.Unsubscribe(context.Background(), subscriber, query)
```
NicolasMahe: this function will never be called as the eventStream will never be closed...
Except if the eventStream chan is closed..
I think it much more common to return a close or cancel function that the caller have to call.

cosmos/client.go
krhubert: Yes, but then what to do with them? Retrun back to client?

sdk/execution/sdk.go
krhubert: I know and that's for purpose. You should be able to create a filter before the creation of service/instance. We already need such functionality and the filter shouldn't block it.

cosmos/module.go
krhubert: That's true, to change.

cosmos/type.go
krhubert: Check out cosmos-sdk package. Every attribiute in x/module has such convetion name.

protobuf/api/execution.go
krhubert: This is a validation about TaskKey (with go validator)

cosmos/client.go
NicolasMahe: > Retrun back to client?

yes

cosmos/client.go
```diff
@@ -133,32 +133,39 @@ func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg, accName, accPassword str
...
+					break
+				}
+
+				hash, err := hash.Decode(tags[0])
+				if err != nil {
+					errC <- err
+				} else {
+					hashC <- hash
+				}
+			case <-ctx.Done():
```
NicolasMahe: Nice ;)

e2e/execution_test.go
```diff
@@ -15,15 +15,16 @@ import (
...
 	ctx := metadata.NewOutgoingContext(context.Background(), passmd)
 	t.Run("stream", func(t *testing.T) {
-		t.Run("with nil filter", func(t *testing.T) {
+		t.Run("nil filter", func(t *testing.T) {
 			_, err := client.ExecutionClient.Stream(context.Background(), &pb.StreamExecutionRequest{})
 			require.NoError(t, err)
 		})
 		// TODO: no error are returned but it supposed to...
-		// t.Run("with not valid filter", func(t *testing.T) {
+		// t.Run("not valid filter", func(t *testing.T) {
```
NicolasMahe: let's remove those test if the filter is not tested...
related to https://github.com/mesg-foundation/engine/pull/1463#discussion_r348532764

NicolasMahe:
Update of the js libs: https://github.com/mesg-foundation/js-sdk/pull/135

core/main.go
```diff
@@ -123,6 +115,8 @@ func loadOrGenDevGenesis(app *cosmos.App, kb *cosmos.Keybase, cfg *config.Config
 }
 
 func main() {
+	xrand.SeedInit()
```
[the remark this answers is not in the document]
NicolasMahe: yes. but i think the goal of the package xrand is to centralized on random related function.
So we don't have the send the random everywhere in the code. it's done once and that's it.

sdk/execution/sdk.go
```diff
@@ -0,0 +1,127 @@
...
+
+	stream, serrC, err := s.client.Stream(ctx, cosmos.EventModuleQuery(backendName))
+	if err != nil {
+		return nil, nil, err
+	}
+
+	execC := make(chan *execution.Execution)
+	errC := make(chan error)
+	go func() {
+	loop:
```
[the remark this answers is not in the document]
NicolasMahe: @krhubert i let you make the decision ;)

cosmos/client.go
```diff
@@ -129,3 +131,41 @@ func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg, accName, accPassword str
...
+				break loop
+			}
+		}
+		close(errC)
+		close(hashC)
+		c.Unsubscribe(context.Background(), subscriber, query)
+	}()
```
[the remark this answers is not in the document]
NicolasMahe: @krhubert i let you make the decision ;)

PR #1469, opened 2019-11-05
Remove servicesdk interface

NicolasMahe:
Other small PR to remove the useless `servicesdk` interface

Dependent on https://github.com/mesg-foundation/engine/pull/1468


.golangci.yml
```diff
@@ -31,3 +31,4 @@ linters:
     - lll
     - prealloc
     - funlen
+    - interfacer
```
krhubert: The linter is deprecated (https://github.com/mvdan/interfacer) - let's not keep it. Also is not the issue in this PR.

PR #1473, opened 2019-11-05
Use filter in list instace api

krhubert:
close https://github.com/mesg-foundation/engine/issues/1466


server/grpc/api/instance.go
```diff
@@ -21,7 +21,7 @@ func NewInstanceServer(sdk *sdk.SDK) *InstanceServer {
 
 // List instances.
 func (s *InstanceServer) List(ctx context.Context, request *protobuf_api.ListInstanceRequest) (*protobuf_api.ListInstanceResponse, error) {
-	instances, err := s.sdk.Instance.List(&instancesdk.Filter{ServiceHash: request.ServiceHash})
+	instances, err := s.sdk.Instance.List(&instancesdk.Filter{ServiceHash: request.Filter.ServiceHash})
```
NicolasMahe: `request.Filter` should be nilable.
Should add a check if `request.Filter` is nil

PR #1474, opened 2019-11-05
Update golangci-lint to 1.21.0

krhubert:
- disable new linters


NicolasMahe:
> any reason why we disable more things? the goal with the linter is overtime to enable more and more not disable more and more.
> 
> gocognit seems really useful

because of https://circleci.com/gh/mesg-foundation/engine/14190?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

krhubert:
> any reason why we disable more things? the goal with the linter is overtime to enable more and more not disable more and more.

Yes, I just wanted to update the linter, so it will be compatible with go 1.13. About enabling linters we can do it for each linter in sperate PRs, because enabling any of this linter mean also code modifications.

PR #1480, opened 2019-11-06
Use cosmos module to manage accounts and use single codec

antho1404:
dependency: https://github.com/mesg-foundation/engine/pull/1462

By using the `auth.NewAnteHandler` we let cosmos manage all the sequence numbers (so https://github.com/mesg-foundation/engine/pull/1462 will work), but also tokens and fees for transactions (it seems but I didn't look at it a lot yet).

By doing I realized that the **ABCI messages should be pure data** (as they are unmarshalled by the post validation step) so they cannot contain the codec as it was initially designed.

There is now a specific codec for each SDK that registers all the types needed for this SDK.

I also removed all the `mempool.ErrTxInCache` as this shouldn't happen anymore and if it happens that's an issue we need to fix and not hide with an "xxx already exists".


krhubert:
> I also removed all the mempool.ErrTxInCache as this shouldn't happen anymore and if it happens that's an issue we need to fix and not hide with an "xxx already exists".

Just wonder if two clients ask for the account sequence number at the same time, they might get the same seq number. The sequence number is only increased in one place (https://github.com/cosmos/cosmos-sdk/blob/master/x/auth/ante/sigverify.go#L252) - on sig validation. So my question is are you sure the two clients get new sequence when they retrive account at the same time.

sdk/instance/codec.go
```diff
@@ -0,0 +1,16 @@
+package instancesdk
+
+import (
+	"github.com/cosmos/cosmos-sdk/codec"
+)
+
+// ModuleCdc is the codec for the module
+var ModuleCdc = codec.New()
```
krhubert: I don't like global variables, but ...

just wonder if we can create one global codec instead of creating many and then juggle them around the source code.

But if this cdc will be used only in for decode msgs then it's fine.

Btw. This one is empty, maybe we don't need init func in that case.

sdk/ownership/codec.go
```diff
@@ -0,0 +1,16 @@
+package ownershipsdk
+
+import (
+	"github.com/cosmos/cosmos-sdk/codec"
+)
+
+// ModuleCdc is the codec for the module
+var ModuleCdc = codec.New()
+
+func init() {
```
krhubert: Can we replace all init function with explicit call?

antho1404: That's following the guidelines of cosmos. Also, we need to make sure that the codec is initialized even if the backend is not yet created. We could have some sync of the messages before we initialize the backend.

We could find a way to remove the init and have a singleton pattern but that will be complex for no real advantage.

sdk/instance/codec.go
antho1404: Agree, multiple codecs are a mess especially that some are as global variables and some in dependencies. 
I will create a single codec global variable that every package will register to.

sdk/ownership/codec.go
krhubert: > That's following the guidelines of cosmos

Yes, i think cosmos is not the right place to get best pracitices. 

>  Also, we need to make sure that the codec is initialized even if the backend

That's easy, we can call InitCodes inside main function. 

Liek for me `init` should be avoided at all cost, because it changes the workflow of execution. There are only few places where it should be used (like goto, or brake/continue outer loops)

cosmos/appfactory.go
```diff
@@ -28,14 +28,12 @@ type AppFactory struct {
 
 // NewAppFactory returns a new AppFactory.
 func NewAppFactory(logger log.Logger, db dbm.DB) *AppFactory {
-	cdc := codec.New()
-	sdk.RegisterCodec(cdc)
-	codec.RegisterCrypto(cdc)
+	sdk.RegisterCodec(codec.Codec)
+	cosmoscodec.RegisterCrypto(codec.Codec)
```
NicolasMahe: Do you think it should be done directly in codec init function?

sdk/ownership/codec.go
antho1404: > Yes, i think cosmos is not the right place to get best pracitices

Maybe not but protobuf or gogoproto might be ;)
Anyway we could have a function that the is called on every package and make sure that this is called at the beginning of main but at the end that's similar to init.

cosmos/appfactory.go
antho1404: done

codec/codec.go
```diff
@@ -0,0 +1,41 @@
...
+
+// UnmarshalBinaryBare https://godoc.org/github.com/tendermint/go-amino#Codec.UnmarshalBinaryBare
+func UnmarshalBinaryBare(bz []byte, ptr interface{}) error {
+	return Codec.UnmarshalBinaryBare(bz, ptr)
+}
+
+// MarshalBinaryBare https://godoc.org/github.com/tendermint/go-amino#Codec.MarshalBinaryBare
+func MarshalBinaryBare(o interface{}) ([]byte, error) {
+	return Codec.MarshalBinaryBare(o)
+}
```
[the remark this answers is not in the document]
krhubert: What would be good to see for example is shortcut for

```
cosmostypes.MustSortJSON(codec.MustMarshalJSON(msg))
```

But it's an option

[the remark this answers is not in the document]
antho1404: we can use `codec.Codec.XXX` I just find it more convenient to have the functions directly like an alternative codec (that in fact uses another codec).

Like that, if we want we can put this `Codec` private or directly implement the un/marshal directly in the functions etc...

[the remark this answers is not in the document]
krhubert: Still, I'm not convinced to that. I don't see pros of "hiding" that codec method. 

> Like that, if we want we can put this Codec private

So let's put the codec as private in the first place ;) But then you'll probably find that is not convenient to use :P

PR #1490, opened 2019-11-07
More verbose error in service backend

krhubert:
Closes https://github.com/mesg-foundation/engine/issues/1471


sdk/service/backend.go
```diff
@@ -53,6 +53,9 @@ func (s *Backend) handler(request cosmostypes.Request, msg cosmostypes.Msg) cosm
 	case msgCreateService:
 		srv, err := s.Create(request, &msg)
 		if err != nil {
+			if cerr, ok := err.(cosmostypes.Error); ok {
+				return cerr.Result()
+			}
```
NicolasMahe: that's a good idea but can you move it to a new handler wrapper in `cosmos/module.go` (like the querier wrapper)?

cosmos/errors.go
```diff
@@ -0,0 +1,14 @@
+package cosmos
+
+import "github.com/cosmos/cosmos-sdk/types"
+
+const (
+	CodeForbidden types.CodeType = 403
+	CodeNotFound                 = 404
+	CodeExist                    = 409
```
NicolasMahe: nice. but why 4xx? is there any reason?

cosmos/errors.go
```diff
@@ -0,0 +1,14 @@
...
+)
+
+const (
+	CodespaceDatabase types.CodespaceType = "database"
+	CodespaceAccount                      = "account"
```
NicolasMahe: what about only one codespace? cosmos seems to use only one:
https://github.com/cosmos/cosmos-sdk/blob/d81d46192a0c26a4ed8bec911813b3239c14a55e/types/errors.go#L56-L57

krhubert: yes cosmos is using one, but left it configurable, so we can adjust the errors.

cosmos/errors.go
krhubert: is just an example. I took it from http rest api , but we can came up with our own codes

krhubert:
The point here is to set up the codespaces and codes - what we want to use.

cosmos/errors.go
NicolasMahe: good idea to reuse the http ones ;)

NicolasMahe:
@krhubert could you finish this PR?

sdk/service/backend.go
krhubert: I'm not sure if we need to implement that into handler wrapper.

Querier expects response (bytes) and error. In querier wrapper a response is marsla from interface into bytes.

For Handler there is no response expected, only `result` interface. So we don't need any wrapper there.

cosmos/errors.go
```diff
@@ -0,0 +1,26 @@
...
+
+const (
+	// CodespaceMesg is a cosmos codespace for all mesg errors.
+	CodespaceMesg types.CodespaceType = "mesg"
+)
+
+const (
+	// Base mesg codes.
+	CodeMesgInternal   types.CodeType = 1000
+	CodeMesgValidation                = 2000
```
NicolasMahe: Why 1000 and then 2000? Why not 1001?
Also the variable name can simply be `CodeValidation`
```suggestion
	CodeValidation                = 1001
```

cosmos/errors.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+const (
+	// CodespaceMesg is a cosmos codespace for all mesg errors.
+	CodespaceMesg types.CodespaceType = "mesg"
+)
+
+const (
+	// Base mesg codes.
+	CodeMesgInternal   types.CodeType = 1000
```
NicolasMahe:
```suggestion
	CodeInternal   types.CodeType = 1000
```

cosmos/errors.go
krhubert: As we discussed, they are different types of error.

1) 1st 1000 is reserved for errors
2) 2nd 1000 is reserverd for validation error - in the future we might want have different validation errors like: HashValidation, DataValidation, AccountValidation, etc

cosmos/errors.go
krhubert: WIll change that.

cosmos/errors.go
NicolasMahe: oh ok. do we want to have so specific error code (HashValidation, DataValidation, etc..)? I would go with just one CodeValidation for now and a nice error message

krhubert: Maybe we will want. I started to numerate such errors from 2000 exactly for that purpose. It the future we migth have password, hash, data, user, accout, and more validations. And by setting it to 1001 we don't have such option, and what are the pros?

NicolasMahe:
@krhubert can you fix linter:
```
cosmos/errors.go:14:2: SA9004: only the first constant in this group has an explicit type (staticcheck)
```

cosmos/errors.go
```diff
@@ -0,0 +1,26 @@
...
+
+const (
+	// CodespaceMesg is a cosmos codespace for all mesg errors.
+	CodespaceMesg types.CodespaceType = "mesg"
+)
+
+const (
+	// Base mesg codes.
+	CodeInternal   types.CodeType = 1000
+	CodeValidation                = 2000
```
NicolasMahe:
```suggestion
	CodeValidation types.CodeType = 2000
```

NicolasMahe:
An idea: could we both implement cosmos and grpc error at the same times?
it will be really great for the client to be able to receive directly the cosmos error and be able to extract the error code ;)

server/grpc/server.go
```diff
@@ -82,3 +87,19 @@ func (s *Server) register() {
 
 	reflection.Register(s.instance)
 }
+
+func streamCosmosErrorInterceptor(srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
```
NicolasMahe: what about a new package `error` containing a struct `error` that both implement the cosmos error and the grpc status code (`GRPCStatus() *status.Status`)?

krhubert: But why you want to implement that, if you already have the function for doing that.
First of all, we don't have a type for errors in `cosmos/errors.go`, we just retruns `cosmostypes.Error`, but even if there will be type for it, the GRPCStatus interface will looks like this

```

type Error types.ErrorImpl

func (e *Error) GRPCStatus() * status.Status {
   return status.Error(codes.Code(e.Code()), e.Error())
}
```

So exacly the same as interceptor.

So to sumamriazie, we can't and we shound't implement that interface.

PR #1491, opened 2019-11-08
Filter instances on backend side

krhubert:
clsoes https://github.com/mesg-foundation/engine/issues/1467


sdk/instance/backend.go
```diff
@@ -1,6 +1,7 @@
 package instancesdk
 
 import (
+	"encoding/json"
```
NicolasMahe: use codec instead of json

sdk/instance/backend.go
```diff
@@ -46,7 +48,12 @@ func (s *Backend) querier(request cosmostypes.Request, path []string, req abci.R
 		}
 		return s.Get(request, hash)
 	case "list":
-		return s.List(request)
+		var f api.ListInstanceRequest_Filter
+		if err := json.Unmarshal(req.Data, &f); err != nil {
```
NicolasMahe: could you move the unmarshal in:
https://github.com/mesg-foundation/engine/blob/940abdaa67779f39482e03b0bb6afbac0a05b745/cosmos/module.go#L112-L127

this function already marshal the data before sending it back

sdk/instance/sdk.go
```diff
@@ -35,26 +38,17 @@ func (s *SDK) Get(hash hash.Hash) (*instance.Instance, error) {
 }
 
 // List returns all instances.
-func (s *SDK) List(f *Filter) ([]*instance.Instance, error) {
-	var instances []*instance.Instance
-	if err := s.client.Query("custom/"+backendName+"/list", nil, &instances); err != nil {
+func (s *SDK) List(f *api.ListInstanceRequest_Filter) ([]*instance.Instance, error) {
```
NicolasMahe: don't forget to delete sdk.Filter line 15

sdk/instance/sdk.go
```diff
@@ -35,26 +38,17 @@ func (s *SDK) Get(hash hash.Hash) (*instance.Instance, error) {
...
+	fb, err := json.Marshal(f)
```
NicolasMahe: could you move the marshal inside the client.Query. it already unmarshal the received data.

sdk/instance/backend.go
NicolasMahe: should use codec

krhubert:
I don't know why e2e fails, can you help me fix that?

cosmos/client.go
```diff
@@ -34,8 +34,8 @@ func NewClient(node *node.Node, kb keys.Keybase, chainID string) *Client {
 }
 
 // Query is abci.query wrapper with errors check and decode data.
-func (c *Client) Query(path string, data cmn.HexBytes, ptr interface{}) error {
-	result, err := c.ABCIQuery(path, data)
+func (c *Client) Query(path string, data, ptr interface{}) error {
+	result, err := c.ABCIQuery(path, codec.MustMarshalJSON(data))
```
NicolasMahe:
```suggestion
	dataB, err := codec.MarshalJSON(data)
	if err != nil {
		return err
	}
	result, err := c.ABCIQuery(path, dataB)
```

NicolasMahe:
> I don't know why e2e fails, can you help me fix that?

`servicesdk.hash` is using `proto.Marshal`.
```
// Hash returns the calculate hash of a service.
func (s *SDK) Hash(req *api.CreateServiceRequest) (hash.Hash, error) {
	var h hash.Hash
	b, err := proto.Marshal(req)
	if err != nil {
		return nil, err
	}
	if err := s.client.Query("custom/"+backendName+"/hash", cmn.HexBytes(b), &h); err != nil {
		return nil, err
	}
	return h, nil
}
```

NicolasMahe:
> > I don't know why e2e fails, can you help me fix that?
> 
> `servicesdk.hash` is using `proto.Marshal`.
> 
> ```
> // Hash returns the calculate hash of a service.
> func (s *SDK) Hash(req *api.CreateServiceRequest) (hash.Hash, error) {
> 	var h hash.Hash
> 	b, err := proto.Marshal(req)
> 	if err != nil {
> 		return nil, err
> 	}
> 	if err := s.client.Query("custom/"+backendName+"/hash", cmn.HexBytes(b), &h); err != nil {
> 		return nil, err
> 	}
> 	return h, nil
> }
> ```

I fixed it ;)

sdk/service/backend.go
```diff
@@ -66,7 +65,7 @@ func (s *Backend) querier(request cosmostypes.Request, path []string, req abci.R
 		return s.List(request)
 	case "hash":
 		var createServiceRequest api.CreateServiceRequest
-		if err := proto.Unmarshal(req.Data, &createServiceRequest); err != nil {
+		if err := codec.UnmarshalJSON(req.Data, &createServiceRequest); err != nil {
```
antho1404: I would prefer to use proto for the encoding here as we had issues with the json encoding when using `oneof`. For the service it is fine but if we do that for the process we will have an issue. That was the reason why we used proto to marshal here and at the same time it's more efficient

sdk/instance/backend.go
```diff
@@ -91,18 +97,21 @@ func (s *Backend) Exists(request cosmostypes.Request, hash hash.Hash) (bool, err
...
 	)
 
+	// filter results
 	for iter.Valid() {
 		var i *instance.Instance
 		if err := codec.UnmarshalBinaryBare(iter.Value(), &i); err != nil {
 			return nil, err
 		}
-		instances = append(instances, i)
+		if f == nil || (f != nil && (f.ServiceHash.IsZero() || i.ServiceHash.Equal(f.ServiceHash))) {
```
antho1404: `f` will not be nil if we pass the first 'or', also inverted the last condition to be more consistent with the conditions

```suggestion
		if f == nil || f.ServiceHash.IsZero() || f.ServiceHash.Equal(i.ServiceHash) {
```

sdk/service/backend.go
NicolasMahe: we have to choose one encoding for the data. we cannot use proto and amino and json.
Service is not using any oneof, so what is the issue here?

antho1404: I'm fine to use binary encode from amino, let's just try to avoid json

NicolasMahe: @krhubert can you replace by binary unmarshal? don't forget to replace also the marshal ;)

sdk/instance/codec.go
```diff
@@ -0,0 +1,10 @@
+package instancesdk
+
+import (
+	"github.com/mesg-foundation/engine/codec"
+	"github.com/mesg-foundation/engine/protobuf/api"
+)
+
+func init() {
+	codec.Codec.RegisterConcrete(&api.ListInstanceRequest_Filter{}, "mesg.types.ListInstanceRequest_Filter ", nil)
```
NicolasMahe: why do you need to register ListInstanceRequest_Filter to the codec? I tried without and its working well.

sdk/service/codec.go
```diff
@@ -2,8 +2,10 @@ package servicesdk
 
 import (
 	"github.com/mesg-foundation/engine/codec"
+	"github.com/mesg-foundation/engine/service"
 )
 
 func init() {
 	codec.RegisterConcrete(msgCreateService{}, "service/create", nil)
+	codec.RegisterConcrete(&service.Service{}, "mesg.types.Service", nil)
```
NicolasMahe: why do you need to register service to the codec? I tried without and its working well.
Only struct used inside an interface should be registered. The msgs have to be register as they are used in Tx as interface

krhubert:
@antho1404 

Fixed issues with panic. Also add e2e tests for different instance hash and it works as expected. I'm not going to debug grpcurl as it is just a helper tool. From e2e tests and code perespective everything work fine.

PR #1497, opened 2019-11-11
Remove use of map in inputs/outputs tasks and event data

NicolasMahe:
See https://github.com/mesg-foundation/engine/issues/1493 for more details.

Closes https://github.com/mesg-foundation/engine/issues/1493


e2e/testdata/test-service/go.mod
```diff
@@ -3,6 +3,8 @@ module main
 go 1.13
 
 require (
-	github.com/mesg-foundation/engine v0.15.0
+	github.com/Nvveen/Gotty v0.0.0-20120604004816-cd527374f1e5 // indirect
+	github.com/kelseyhightower/envconfig v1.3.0 // indirect
```
krhubert: Those are not needed. I've run go mod tidy

service/service_parameter.go
```diff
@@ -8,25 +8,29 @@ import (
...
 }
 
 // Validate checks if service parameter hash proper types for arrays and objects.
 func (p *Service_Parameter) Validate(value *types.Value) error {
+	isNilOrNullValue := false
 	if value == nil {
+		isNilOrNullValue = true
+	} else {
+		_, isNilOrNullValue = value.GetKind().(*types.Value_NullValue)
+	}
```
krhubert: _, isNilOrNullValue := value.GetKind().(*types.Value_NullValue)
isNilOrNullValue = isNilOrNullValue || value == nil

NicolasMahe:
> What I find really stressful is the fact there will be no more keys... It's very error prone to operate on index (like in processes). Also we lost all the information about key names (in case of error for egzample).

The key still exists in `Service.Parameter` struct. We just don't use them an `index` of another struct. So for errors, we can still display the key with something like `task.Parameter[index].Key`.

> Also I don't like the fact we keep old mesg.yml format because it might be missleading for other dev. We should update it at least in e2e tests.

The current way to implement it is to keep the dev tools as it is (no breaking change on mesg.yml or js lib). But yes I agree we could update it in the future to make it even simpler for everyone ;)

service/service_parameter.go
antho1404: ` || value == nil` is useless because it will panic first if value is `nil` in that case

krhubert:
@NicolasMahe can you confirmt that we can close this PR?

PR #1500, opened 2019-11-14
Test/e2e complex service

krhubert:
More complex service test for https://github.com/mesg-foundation/engine/issues/1446


NicolasMahe:
Why not implementing the complex service directly in the current e2e test service?

krhubert:
To split the tests services and not put everthing into one.

e2e/complex_service_test.go
```diff
@@ -0,0 +1,55 @@
...
+		testServiceHash = resp.Hash
+	})
+
+	t.Run("get", func(t *testing.T) {
+		ctx := metadata.NewOutgoingContext(context.Background(), passmd)
+
+		service, err := client.ServiceClient.Get(ctx, &pb.GetServiceRequest{Hash: testServiceHash})
+		require.NoError(t, err)
+		want := pb.TransformCreateReqToService(req)
+		// TODO: check why the hashes are different
```
NicolasMahe: is it fixed?

e2e/testdata/test-complex-service/main.go
```diff
@@ -0,0 +1,139 @@
...
+	}); err != nil {
+		log.Fatal(err)
+	}
+
+	for {
+		exec, err := stream.Recv()
+		if err != nil {
+			log.Fatal(err)
+		}
+		log.Printf("recive exeuction %s %s\n", exec.TaskKey, exec.Hash)
```
NicolasMahe:
```suggestion
		log.Printf("receive execution %s %s\n", exec.TaskKey, exec.Hash)
```

e2e/complex_service_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+		testServiceHash = resp.Hash
+	})
+
+	t.Run("get", func(t *testing.T) {
+		ctx := metadata.NewOutgoingContext(context.Background(), passmd)
+
+		service, err := client.ServiceClient.Get(ctx, &pb.GetServiceRequest{Hash: testServiceHash})
+		require.NoError(t, err)
+		want := pb.TransformCreateReqToService(req)
```
NicolasMahe: I don't like this `protobuf.TransformCreateReqToService`.
The hash calculation should be done by the service sdk.
And I don't think it makes sense to check if the hash is expected. This is a e2e tests, not a unit test.

NicolasMahe: I suggest to revert this function and not check the expected service hash.

krhubert: Ok we don't need to check it. But `TransformCreateReqToService` should there, but maybe without hash calculation. Sdk pachage shoudn't have logic how transform pb structures

e2e/definition_test.go
```diff
@@ -0,0 +1,142 @@
...
+				Key: "error_ok",
+				Data: []*service.Service_Parameter{
+					{
+						Key:  "msg",
+						Type: "String",
+					},
+				},
+			},
+		},
+		Source: "QmPG1Ze96pH1EgVMWsGKM33jXoG63rigMncSEqZXP7oncq",
```
NicolasMahe: I like to not depend on the cli compilation of the service but we still need to use the cli to archive and upload the source code to ipfs.
Could this hash be at least read from a dedicated file? Or even better, the source code to be uploaded to ipfs every time? Or even, create a local http server to serve the source code when the engine require it (with a modif of the `config.ipfsendpoint`)?

e2e/complex_service_test.go
NicolasMahe: With the discussion of the last meeting, we spoke about bringing the requests closer to the sdk..

e2e/definition_test.go
krhubert: For sure that would be great, but as for now it's kind of overkill feature with questionable benefits.

For dedicated file we need to set up file coping (for local machine, ci). Same for http server.

This could be done but in for sure in differenent PR and with lower priority.

e2e/complex_service_test.go
krhubert: The request yes, but this is transformation and it's different.

krhubert:
@NicolasMahe I've tride to debug it without success, can you take a look on that for few mins?

NicolasMahe:
> @NicolasMahe I've tried to debug it without success, can you take a look on that for few mins?

I improved the test but could not figure how to fix this issue. So the test is now skipped with a FIXME message.

The problem occurs only on the CI and only with a service with at least one dependency. It looks like the dependency is server killed and thus the service is never stopped. I will create an issue and investigate later as it is not blocking.

PR #1502, opened 2019-11-15
Add amino encoding functionality to proto Struct

NicolasMahe:
- Add `amino:"unsafe"` flag to proto `Number` type to make it work with amino.
- Add `MarshalAmino` and `UnmarshalAmino` functions to proto `Struct` type to transform the Fields type from `map[string]*Value` to a key/value value and vice-versa. The key/value array is sorted by keys to guarantee its order and generate the same hash (see order of values in the tests).

Related to https://github.com/mesg-foundation/engine/issues/1493


protobuf/types/struct.go
```diff
@@ -0,0 +1,38 @@
...
+// KeyValue is a simple key/value representation of one field of a Struct.
+type KeyValue struct {
+	Key   string
+	Value *Value
+}
+
+// MarshalAmino transforms the Struct to an array of key/value.
+func (m Struct) MarshalAmino() ([]KeyValue, error) {
+	p := make([]KeyValue, 0)
+	fieldKeys := make([]string, 0)
```
krhubert: Use len(m.Fields) for preallocating space for sliceses. Then used index to insert instead of append

protobuf/types/struct.go
```diff
@@ -0,0 +1,38 @@
...
+			Key:   key,
+			Value: m.Fields[key],
+		})
+	}
+	return p, nil
+}
+
+// UnmarshalAmino transforms the key/value array to a Struct.
+func (m *Struct) UnmarshalAmino(keyValues []KeyValue) error {
+	fields := make(map[string]*Value)
```
krhubert:
```suggestion
	fields := make(map[string]*Value, len(keyValues))
```

protobuf/types/struct_test.go
```diff
@@ -0,0 +1,100 @@
...
+		structValueSort2, err = codec.MarshalBinaryBare(structSort2)
+		require.NoError(t, err)
+		require.Equal(t, hash.Dump(structValueSort1), hash.Dump(structValueSort2))
+	})
+	t.Run("Unmarshal", func(t *testing.T) {
+		require.NoError(t, codec.UnmarshalBinaryBare(structValueSort1, &structUnm1))
+		require.Equal(t, structSort1, structUnm1)
+		require.NoError(t, codec.UnmarshalBinaryBare(structValueSort2, &structUnm2))
+		require.Equal(t, structSort2, structUnm2)
+		require.Equal(t, structUnm1, structUnm2)
```
krhubert:
```suggestion
		require.True(t, structUnm1.Equal(structUnm2))
```

protobuf/types/struct_test.go
```diff
@@ -0,0 +1,100 @@
...
+	t.Run("Marshal", func(t *testing.T) {
+		structValueSort1, err = codec.MarshalBinaryBare(structSort1)
+		require.NoError(t, err)
+		structValueSort2, err = codec.MarshalBinaryBare(structSort2)
+		require.NoError(t, err)
+		require.Equal(t, hash.Dump(structValueSort1), hash.Dump(structValueSort2))
+	})
+	t.Run("Unmarshal", func(t *testing.T) {
+		require.NoError(t, codec.UnmarshalBinaryBare(structValueSort1, &structUnm1))
+		require.Equal(t, structSort1, structUnm1)
```
krhubert:
```suggestion
		require.True(t, structSort1.Equal(structUnm1))
```

protobuf/types/struct_test.go
```diff
@@ -0,0 +1,100 @@
...
+					"struct": {Kind: &Value_StructValue{StructValue: &Struct{Fields: map[string]*Value{
+						"10000":  {Kind: &Value_NumberValue{NumberValue: 10}},
+						"1":      {Kind: &Value_StringValue{StringValue: "valuea"}},
+						"null":   {Kind: &Value_NullValue{NullValue: NullValue_NULL_VALUE}},
+						"number": {Kind: &Value_NumberValue{NumberValue: 10}},
+					}}}},
+				}}}},
+				"list": {Kind: &Value_ListValue{ListValue: &ListValue{Values: []*Value{
+					{Kind: &Value_StringValue{StringValue: "dvaluea"}},
+				}}}},
```
krhubert: todo?

- s/valuea/value/
- change some keys (for example I'm not sure why `StringValue` has key `1` - maybe string_1?)
- Key10000 hodls value 10 - s/10000/10/ ?

protobuf/types/codec.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+func init() {
+	codec.RegisterInterface((*isValue_Kind)(nil), nil)
+	codec.RegisterConcrete(&Value_NullValue{}, "mesg.types.Value_NullValue", nil)
+	codec.RegisterConcrete(&Value_NumberValue{}, "mesg.types.Value_NumberValue", nil)
+	codec.RegisterConcrete(&Value_StringValue{}, "mesg.types.Value_StringValue", nil)
+	codec.RegisterConcrete(&Value_BoolValue{}, "mesg.types.Value_BoolValue", nil)
+	codec.RegisterConcrete(&Value_ListValue{}, "mesg.types.Value_ListValue", nil)
+	codec.RegisterConcrete(&Value_StructValue{}, "mesg.types.Value_StructValue", nil)
```
krhubert: Do we need to register those types?

protobuf/types/struct_test.go
NicolasMahe: The mess with the keys is to try to create the most random ordered map to test if the sort is correctly working.
I find out that for some reason, if the keys are only one alpha characters, then the map is "automatically" sorted. That's why some keys are numeric and some other are just the type of value

protobuf/types/codec.go
NicolasMahe: Yes because they can be used behind an interface

PR #1504, opened 2019-11-15
Fix hash length

NicolasMahe:
Check the length of hash in Marshal and Unmarshal and return always the actual length.

I had the issue where the Marshal function accepted a hash with the wrong length, and the unmarshal function was return an unrecognized EOF error because the length return by the hash was hardcoded...

### DIFF
--- a/hash/hash.go
+++ b/hash/hash.go
@@ -19,6 +19,13 @@ var DefaultHash = sha256.New
 // size is a default size for hashing algorithm.
 var size = DefaultHash().Size()
 
+func checkLen(data []byte) error {
+	if len(data) != size {
+		return fmt.Errorf("hash: invalid length")
+	}
+	return nil
+}
+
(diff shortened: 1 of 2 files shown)


hash/hash.go
```diff
@@ -19,6 +19,13 @@ var DefaultHash = sha256.New
 // size is a default size for hashing algorithm.
 var size = DefaultHash().Size()
 
+func checkLen(data []byte) error {
+	if len(data) != size {
```
krhubert: if this is hash.Hahs length it should be in hash package , what do you think?

NicolasMahe: This function is used with different parameters: 2 with directly `Hash`, 2 with `[]byte`.
So it doesn't make sense to put it as a struct function (`func (h *Hash) checkLen(data []byte) error {`).
> it should be in hash package

it's already in hash package.

PR #1509, opened 2019-11-19
Implement cosmos event and subscriber

NicolasMahe:
@krhubert please improve this PR, checks the TODO and remove the panic and improve the `randStringRunes` function.
I let you choose how you want to manage the errors on chan (new chan? new struct that contain the data and the error?)

### DIFF
--- a/cosmos/client.go
+++ b/cosmos/client.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand"
 	"time"
 
 	"github.com/cosmos/cosmos-sdk/crypto/keys"
 	sdktypes "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/x/auth"
 	"github.com/mesg-foundation/engine/codec"
+	"github.com/mesg-foundation/engine/hash"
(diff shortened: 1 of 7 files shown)


cosmos/client.go
```diff
@@ -129,3 +131,58 @@ func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg, accName, accPassword str
 		return nil, errors.New("i/o timeout")
 	}
 }
+
+// Stream subscribes to the provided query and returns the hash of the matching ressources.
+func (c *Client) Stream(query string) (<-chan hash.Hash, func() error, error) {
```
NicolasMahe: @krhubert the goal of this function is to provide an easier way to use the `Subscribe` functions of cosmos client and also to extract the hash from the event.
This issue I have is it may actually not be worth it as the caller of `Stream` still needs to return a chan, manage its own error (so new error chan or new struct of data) and also need to manage the `closer` function...
Maybe with just some good helper to extract extract the hash and generate a new subscriber name this function is not worth it.
Wait for the next PR to check the how it is used.

krhubert:
closing because all changes will be done in https://github.com/mesg-foundation/engine/pull/1463

PR #1512, opened 2019-11-19
Inject runner hash in docker services

NicolasMahe:
- Inject runner hash in docker services for future use in execution api
- Update e2e service test


e2e/testdata/test-service/main.go
```diff
@@ -19,6 +19,7 @@ const (
 	// env variables for configure mesg client.
 	envMesgEndpoint     = "MESG_ENDPOINT"
 	envMesgInstanceHash = "MESG_INSTANCE_HASH"
+	envMesgRunnerHash   = "MESG_RUNNER_HASH"
```
krhubert: Why we have it here if we don't use it?

e2e/testdata/test-service/main.go
```diff
@@ -136,7 +147,7 @@ func main() {
 		}
 
 		if _, err := client.EventClient.Create(context.Background(), &pb.CreateEventRequest{
-			InstanceHash: exec.InstanceHash,
+			InstanceHash: client.InstanceHash,
```
krhubert: Why change to client? It was here to make sure Execution has proper hash.

sdk/runner/sdk.go
```diff
@@ -67,23 +69,27 @@ func (s *SDK) Create(req *api.CreateRunnerRequest, accountName, accountPassword
 	}
 	instanceEnv := xos.EnvMergeSlices(srv.Configuration.Env, req.Env)
 	envHash := hash.Dump(instanceEnv)
-	// TODO: should be done by instance
+	// TODO: should be done by instance or runner
 	instanceHash := hash.Dump(&instance.Instance{
 		ServiceHash: srv.Hash,
 		EnvHash:     envHash,
 	})
+	expRunnerHash := hash.Dump(&runner.Runner{
```
krhubert:
```suggestion
	runnerHash := hash.Dump(&runner.Runner{
```
?

sdk/runner/sdk.go
```diff
@@ -92,7 +98,14 @@ func (s *SDK) Create(req *api.CreateRunnerRequest, accountName, accountPassword
...
 		return nil, err
 	}
-	return s.Get(tx.Data)
+	run, err := s.Get(tx.Data)
+	if err != nil {
+		return nil, err
+	}
+	if !run.Hash.Equal(expRunnerHash) {
+		return nil, errors.New("calculated runner hash is not the same")
+	}
```
krhubert: Why we have this check here? We are getting runner from our cosmos rest backend - I don't see why this check is here.

Also why we get the runner in the first place when we already have it defined above.

e2e/testdata/test-service/main.go
NicolasMahe: The event should not be linked with the execution.
Both event and execution should use the hash from ENV

sdk/runner/sdk.go
NicolasMahe: It's just to make sure that the transaction created the same hash.
The SDK is done by the client alone as opposed as the backend that is executed by all the node of the network (and more important the validators)

e2e/testdata/test-service/main.go
NicolasMahe: For next PR

sdk/runner/sdk.go
krhubert: But if the tx was already created with wrong hash shoudn't we take some action to remove it? Otherwise I'm not sure we should put it here

e2e/testdata/test-service/main.go
krhubert: so shoudn't be in next PR? :)

sdk/runner/sdk.go
```diff
@@ -92,7 +98,14 @@ func (s *SDK) Create(req *api.CreateRunnerRequest, accountName, accountPassword
 		defer onError()
 		return nil, err
 	}
-	return s.Get(tx.Data)
+	run, err := s.Get(tx.Data)
+	if err != nil {
+		return nil, err
+	}
+	if !run.Hash.Equal(expRunnerHash) {
+		return nil, errors.New("calculated runner hash is not the same")
```
NicolasMahe:
```suggestion
		defer onError()
		return nil, errors.New("calculated runner hash is not the same")
```

sdk/runner/sdk.go
```diff
@@ -92,7 +98,14 @@ func (s *SDK) Create(req *api.CreateRunnerRequest, accountName, accountPassword
 		defer onError()
 		return nil, err
 	}
-	return s.Get(tx.Data)
+	run, err := s.Get(tx.Data)
+	if err != nil {
```
NicolasMahe:
```suggestion
	defer onError()
	if err != nil {
```

sdk/runner/sdk.go
NicolasMahe: > But if the tx was already created with wrong hash shoudn't we take some action to remove it? Otherwise I'm not sure we should put it here

good catch. it now call `onError` function that will stop the runner if there is an issue

e2e/testdata/test-service/main.go
krhubert: but it could left here :)

PR #1513, opened 2019-11-19
Remove service env MESG_TOKEN

NicolasMahe:
Remove service env MESG_TOKEN until a real auth system is put in place


antho1404:
Why this has been merged without approval? 
We shouldn't remove this injected env variable. Without this, the service doesn't know what to communicate back to the engine.
This needs to be reverted also I'm pretty sure that tests will fail on this one but you didn't wait for it

PR #1514, opened 2019-11-19
Remove service env MESG_TOKEN

NicolasMahe:
Remove service env MESG_TOKEN until a real auth system is put in place.


antho1404:
See my comment here https://github.com/mesg-foundation/engine/pull/1513#issuecomment-555428836

We need to improve the authorization system but not remove the current one (which I agree is too simple)

PR #1519, opened 2019-11-20
Add data validation to all resource

krhubert:
close https://github.com/mesg-foundation/engine/issues/1505


protobuf/api/event.proto
```diff
@@ -27,18 +27,22 @@ message StreamEventRequest {
...
     bytes instanceHash = 2 [
+      (gogoproto.moretags) = 'validate:"hash"',
       (gogoproto.customtype) = "github.com/mesg-foundation/engine/hash.Hash",
       (gogoproto.nullable) = false
     ];
 
     // key is the key of the event.
-    string key = 3;
+    string key = 3 [
+      (gogoproto.moretags) = 'hash:"name:3" validate:"printascii"'
```
NicolasMahe: no need for hash
```suggestion
      (gogoproto.moretags) = 'validate:"printascii"'
```

x/xvalidator/validator.go
```diff
@@ -17,7 +21,55 @@ const (
 	envSeparator  = "="
 )
 
-var envNameRegexp = regexp.MustCompile("^[a-zA-Z][a-zA-Z0-9_]*$")
+// New returns a new instance of 'validate' with more validation fields.
```
NicolasMahe: Could you also move the function `validateServiceStruct ` from service validator package and make it generic:
https://github.com/mesg-foundation/engine/blob/6df2444c43929f697e5a41105a3d4cd404a2a2b2/service/validator/validator.go#L28-L45

to something like:
```go
func Validate(s interface{}) error {
	 validate, translator := New()
	var errs xerrors.Errors
	// validate service struct based on tag
	if err := validate.Struct(s); err != nil {
		for _, e := range err.(validator.ValidationErrors) {
			// Remove the name of the struct and the field from namespace
			trimmedNamespace := strings.TrimPrefix(e.Namespace(), namespacePrefix)
			trimmedNamespace = strings.TrimSuffix(trimmedNamespace, e.Field())
			// Only use it when in-cascade field
			namespace := ""
			if e.Field() != trimmedNamespace {
				namespace = trimmedNamespace
			}
			errs = append(errs, fmt.Errorf("%s%s", namespace, e.Translate(translator)))
		}
	}
	return errs.ErrorOrNil()
}
```

I guess having a simple helper that manage improve the error and manage the translation will be better ;)

NicolasMahe: Also, the remaining functions in service/validator package should be move to service package.

protobuf/api/execution.proto
```diff
@@ -31,12 +31,22 @@ service Execution {
...
+  string taskKey = 2 [
+    (gogoproto.moretags) = 'validate:"printascii"'
+  ];
+
   mesg.protobuf.Struct inputs = 3;
-  repeated string tags = 4;
+
+  // tags to executions.
+  repeated string tags = 4 [
+    (gogoproto.moretags) = 'hash:"name:10"'
```
NicolasMahe:
```suggestion
```

x/xvalidator/validator.go
krhubert: For now I left it in the service, because I'm not sure if we want such generic validation everywhere. It was designed for service only. 

But I've moved the creation of `Validate`

krhubert:
I've left validation for cosmos app (backend), because it could be easily done when PR about execution #1463 will be merged.

Then validation will be done in the handler for every message. I haven't found middleware (or some similar concept) for cosmos handlers.

hash/hash.go
```diff
@@ -93,6 +93,15 @@ func Decode(h string) (Hash, error) {
 	return Hash(hash), nil
 }
 
+// DecodeFromBytes decodes hash and checks it length.
+// It returns empty hash on nil slice of bytes.
+func DecodeFromBytes(data []byte) (Hash, error) {
```
NicolasMahe: This function should be called by a bunch of function of this file that use the pattern:
```go
	if len(data) != size {
		return nil, fmt.Errorf("hash: invalid length")
	}
	return Hash(data), nil
```
functions: `Decode`, `Unmarshal`, `UnmarshalJSON` should also

protobuf/api/event.proto
```diff
@@ -54,7 +58,9 @@ message CreateEventRequest {
   ];
 
   // key is the key of the event.
-  string key = 2;
+  string key = 2 [
+    (gogoproto.moretags) = 'validate:"printascii"'
```
NicolasMahe:
```suggestion
    (gogoproto.moretags) = 'validate:"required,printascii"'
```

NicolasMahe: Also add required to `instanceHash` and `data`

protobuf/api/execution.proto
```diff
@@ -31,12 +31,22 @@ service Execution {
 // CreateExecutionRequest defines request to create a single execution.
 message CreateExecutionRequest {
   bytes instanceHash = 1 [
+    (gogoproto.moretags) = 'validate:"hash"',
```
NicolasMahe:
```suggestion
    (gogoproto.moretags) = 'validate:"required,hash"',
```

protobuf/api/execution.proto
```diff
@@ -31,12 +31,22 @@ service Execution {
...
     (gogoproto.customtype) = "github.com/mesg-foundation/engine/hash.Hash",
     (gogoproto.nullable) = false
   ];
-  string taskKey = 2;
+
+    // taskKey to filter executions.
+  string taskKey = 2 [
+    (gogoproto.moretags) = 'validate:"printascii"'
```
NicolasMahe:
```suggestion
    (gogoproto.moretags) = 'validate:"required,printascii"'
```

protobuf/api/execution.proto
```diff
@@ -52,6 +62,7 @@ message CreateExecutionResponse {
 message GetExecutionRequest {
   // Execution's hash to fetch.
   bytes hash = 1 [
+    (gogoproto.moretags) = 'validate:"hash"',
```
NicolasMahe:
```suggestion
    (gogoproto.moretags) = 'validate:"required,hash"',
```

protobuf/api/execution.proto
```diff
@@ -85,6 +101,7 @@ message StreamExecutionRequest{
 message UpdateExecutionRequest {
   // Hash represents execution.
   bytes hash = 1 [
+    (gogoproto.moretags) = 'validate:"hash"',
```
NicolasMahe:
```suggestion
    (gogoproto.moretags) = 'validate:"required,hash"',
```

protobuf/api/instance.proto
```diff
@@ -23,6 +23,7 @@ service Instance {
 // The request's data for the `Get` API.
 message GetInstanceRequest {
   bytes hash = 1 [
+    (gogoproto.moretags) = 'validate:"hash"',
```
```suggestion
    (gogoproto.moretags) = 'validate:"required,hash"',
```

protobuf/api/runner.proto
```diff
@@ -30,6 +30,7 @@ service Runner {
 // The request's data for the `Get` API.
 message GetRunnerRequest {
   bytes hash = 1 [
+    (gogoproto.moretags) = 'validate:"hash"',
```
NicolasMahe:
```suggestion
    (gogoproto.moretags) = 'validate:"required,hash"',
```

protobuf/api/runner.proto
```diff
@@ -63,12 +67,15 @@ message ListRunnerResponse {
 message CreateRunnerRequest {
   // Service's hash to start the runner with.
   bytes serviceHash = 1 [
+    (gogoproto.moretags) = 'validate:"hash"',
```
NicolasMahe:
```suggestion
    (gogoproto.moretags) = 'validate:"required,hash"',
```

protobuf/api/runner.proto
```diff
@@ -84,6 +91,7 @@ message CreateRunnerResponse {
 message DeleteRunnerRequest {
   // Runner's hash
   bytes hash = 1 [
+    (gogoproto.moretags) = 'validate:"hash"',
```
NicolasMahe:
```suggestion
    (gogoproto.moretags) = 'validate:"required,hash"',
```

protobuf/api/service.proto
```diff
@@ -33,33 +33,50 @@ service Service {
 // The request's data for the `Create` API.
 message CreateServiceRequest {
   // Service's sid.
-  string sid = 1;
+  string sid = 1 [
+    (gogoproto.moretags) = 'validate:"required,printascii,max=63,domain"'
```
NicolasMahe: sid is not required in the create request.
if it doesn't exist it is set by the service sdk.
```suggestion
    (gogoproto.moretags) = 'validate:"printascii,max=63,domain"'
```

protobuf/api/service.proto
```diff
@@ -93,6 +110,7 @@ message ListServiceResponse {
 message ExistsServiceRequest {
   // The service's hash of the existing service. This hash is nil if exists is fals.
   bytes hash = 1 [
+    (gogoproto.moretags) = 'validate:"hash"',
```
NicolasMahe:
```suggestion
    (gogoproto.moretags) = 'validate:"required,hash"',
```

NicolasMahe: don't forget to add the validation on `GetServiceRequest`

protobuf/types/event.proto
```diff
@@ -19,14 +19,14 @@ message Event {
 
   // instanceHash is hash of instance that can proceed an execution.
   bytes instanceHash = 2 [
-    (gogoproto.moretags) = 'hash:"name:2"',
+    (gogoproto.moretags) = 'hash:"name:2" validate:"hash"',
```
NicolasMahe:
```suggestion
    (gogoproto.moretags) = 'hash:"name:2" validate:"required,hash"',
```
don't forget to add validation on hash

protobuf/types/event.proto
```diff
@@ -19,14 +19,14 @@ message Event {
...
     (gogoproto.customtype) = "github.com/mesg-foundation/engine/hash.Hash",
     (gogoproto.nullable) = false
   ];
 
   // key is the key of the event.
   string key = 3 [
-    (gogoproto.moretags) = 'hash:"name:3"'
+    (gogoproto.moretags) = 'hash:"name:3" validate:"printascii"'
```
NicolasMahe:
```suggestion
    (gogoproto.moretags) = 'hash:"name:3" validate:"required,printascii"'
```

sdk/runner/msgs.go
```diff
@@ -8,9 +8,9 @@ import (
 
 // msgCreateRunner defines a state transition to create a runner.
 type msgCreateRunner struct {
-	Address     cosmostypes.AccAddress `json:"address"`
-	ServiceHash hash.Hash              `json:"serviceHash"`
-	EnvHash     hash.Hash              `json:"envHash"`
+	Address     cosmostypes.AccAddress `json:"address" validate:"accaddress"`
+	ServiceHash hash.Hash              `json:"serviceHash" validate:"hash"`
```
NicolasMahe:
```suggestion
	ServiceHash hash.Hash              `json:"serviceHash" validate:"required,hash"`
```

sdk/runner/msgs.go
```diff
@@ -8,9 +8,9 @@ import (
 
 // msgCreateRunner defines a state transition to create a runner.
 type msgCreateRunner struct {
-	Address     cosmostypes.AccAddress `json:"address"`
-	ServiceHash hash.Hash              `json:"serviceHash"`
-	EnvHash     hash.Hash              `json:"envHash"`
+	Address     cosmostypes.AccAddress `json:"address" validate:"accaddress"`
```
NicolasMahe:
```suggestion
	Address     cosmostypes.AccAddress `json:"address" validate:"required,accaddress"`
```

sdk/runner/msgs.go
```diff
@@ -58,8 +58,8 @@ func (msg msgCreateRunner) GetSigners() []cosmostypes.AccAddress {
 
 // msgDeleteRunner defines a state transition to delete a runner.
 type msgDeleteRunner struct {
-	Address    cosmostypes.AccAddress `json:"address"`
-	RunnerHash hash.Hash              `json:"runnerHash"`
+	Address    cosmostypes.AccAddress `json:"address" validate:"accaddress"`
```
NicolasMahe:
```suggestion
	Address    cosmostypes.AccAddress `json:"address" validate:"required,accaddress"`
```

sdk/runner/msgs.go
```diff
@@ -58,8 +58,8 @@ func (msg msgCreateRunner) GetSigners() []cosmostypes.AccAddress {
...
+	RunnerHash hash.Hash              `json:"runnerHash" validate:"hash"`
```
NicolasMahe:
```suggestion
	RunnerHash hash.Hash              `json:"runnerHash" validate:"required,hash"`
```

sdk/service/msgs.go
```diff
@@ -9,7 +9,7 @@ import (
 // msgCreateService defines a state transition to create a service.
 type msgCreateService struct {
 	Request *api.CreateServiceRequest `json:"request"`
-	Owner   cosmostypes.AccAddress    `json:"owner"`
+	Owner   cosmostypes.AccAddress    `json:"owner" validate:"accaddress"`
```
NicolasMahe:
```suggestion
	Owner   cosmostypes.AccAddress    `json:"owner" validate:"required,accaddress"`
```

sdk/service/msgs.go
```diff
@@ -9,7 +9,7 @@ import (
 // msgCreateService defines a state transition to create a service.
 type msgCreateService struct {
 	Request *api.CreateServiceRequest `json:"request"`
```
NicolasMahe:
```suggestion
	Request *api.CreateServiceRequest `json:"request"` validate:"required"`
```

server/grpc/server.go
```diff
@@ -69,6 +71,13 @@ func (s *Server) Close() {
 	s.instance.GracefulStop()
 }
 
+func validateInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
```
NicolasMahe: I'm not sure it's the best place here.
I would like the SDK to also check its inputs and a lot of grpc request message are directly forward to the sdk.
maybe check at both place?

x/xvalidator/validator.go
```diff
@@ -17,7 +23,98 @@ const (
...
+}
+
+// New returns a new instance of 'validate' with more validation fields.
+func New() (*validator.Validate, ut.Translator) {
+	return NewWithPrefix("")
+}
+
+// IsHash validates if given field is valid hash.
+// It checks both string and slice of bytes.
+func IsHash(fl validator.FieldLevel) bool {
```
NicolasMahe: `IsHash` should not return false if the hash is not set. if we want to force an hash to be set, then we should use `required` validation.

x/xvalidator/validator.go
```diff
@@ -17,7 +23,98 @@ const (
...
+			break
+		}
+		_, err := hash.DecodeFromBytes(v.Bytes())
+		return err == nil
+	}
+	return false
+}
+
+// IsAccAddress validates if given field is valid cosmos account address.
+func IsAccAddress(fl validator.FieldLevel) bool {
```
NicolasMahe: Same comment for hash, `accAddress` should not return false if it's empty

NicolasMahe:
> Then validation will be done in the handler for every message. I haven't found middleware (or some similar concept) for cosmos handlers.

each cosmos message has to implement `func (msg msgCreateRunner) ValidateBasic() cosmostypes.Error`. the validation should be done there.

hash/hash.go
krhubert: There is an update in Execution SDK that also modifies that. I will refactor it after Execution and this PR will merge

x/xvalidator/validator.go
krhubert: It should, the same was IsDomain works, or any other different validatior. As you said you can use `required` as well as `omitempty` to controls such behaviour.

x/xvalidator/validator.go
krhubert: Same for `IsHash`

server/grpc/server.go
krhubert: Why SDK should check that. For now we use only grpc as entry for sdk. SDK transforms over time and for now, you just can't use it without running engine. So it should stay here for now. And when sdk transforms to the point where you can use it without engine, then we might add the validation here.

hash/hash.go
NicolasMahe: It's merged since 3 days ;)
https://github.com/mesg-foundation/engine/pulls?q=is%3Apr+is%3Aclosed+sort%3Aupdated-desc

server/grpc/server.go
NicolasMahe: I just think that the verification should be done in last entry point of the app.
We also plan to move to SDK the definition of the message, so it will make even more sense to already put the verification in SDK.

krhubert: what is the last entry point? For now, it is the cosmos app and there messages are validated. SDK is a kind of middleware between our grpc api and cosmos api. 

> We also plan to move to SDK the definition of the message, 

Then we will move the validation as well, but for now let's keep it as it is.

hash/hash.go
krhubert: I looked at it and I need to rethink the api. Because some of them (like Marshal) requires make and copy slice,and some don't

sdk/runner/msgs.go
```diff
@@ -58,8 +62,8 @@ func (msg msgCreateRunner) GetSigners() []cosmostypes.AccAddress {
 
 // msgDeleteRunner defines a state transition to delete a runner.
 type msgDeleteRunner struct {
-	Address    cosmostypes.AccAddress `json:"address"`
-	RunnerHash hash.Hash              `json:"runnerHash"`
+	Address    cosmostypes.AccAddress `json:"address" validate:"required,accaddress"`
+	RunnerHash hash.Hash              `json:"runnerHash" validate:"required,hash"`
```
NicolasMahe: don't forget to update
```func (msg msgDeleteRunner) ValidateBasic() cosmostypes.Error {```

File: protobuf/api/event.proto [hunk 28]
```diff
@@ -54,7 +58,9 @@ message CreateEventRequest {
   ];
```
NicolasMahe: Also add required to instanceHash and data

protobuf/types/event.proto
```diff
@@ -19,14 +19,14 @@ message Event {
```
NicolasMahe: don't forget to add validation on hash

e2e/execution_test.go
```diff
@@ -32,6 +32,7 @@ func testExecution(t *testing.T) {
 
 			resp, err := client.ExecutionClient.Create(context.Background(), &pb.CreateExecutionRequest{
 				TaskKey:      "ping",
+				InstanceHash: testInstanceHash,
```
NicolasMahe: this has been removed
```suggestion
```

protobuf/api/execution.proto
```diff
@@ -33,23 +33,46 @@ service Execution {
 
 // CreateExecutionRequest defines request to create a single execution.
 message CreateExecutionRequest {
-  string taskKey = 2;
+  bytes instanceHash = 1 [
```
NicolasMahe: instanceHash has been remove...
you didn't fix the merge correctly

protobuf/api/event.proto
NicolasMahe: in `CreateEventRequest`

protobuf/types/event.proto
krhubert: In different PR

PR #1527, opened 2019-11-25
Move process to cosmos network

krhubert:
closes https://github.com/mesg-foundation/engine/issues/1526


sdk/process/msgs.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	cosmostypes "github.com/cosmos/cosmos-sdk/types"
+	"github.com/mesg-foundation/engine/codec"
+	"github.com/mesg-foundation/engine/hash"
+	"github.com/mesg-foundation/engine/protobuf/api"
+)
+
+// msgCreateProcess defines a state transition to create a service.
+type msgCreateProcess struct {
+	Owner   cosmostypes.AccAddress    `json:"owner"`
```
NicolasMahe: you could directly add the validation like on PR https://github.com/mesg-foundation/engine/pull/1519 on `msgCreateProcess` and `msgDeleteProcess`

sdk/process/backend.go
```diff
@@ -0,0 +1,146 @@
...
+func (s *Backend) handler(req cosmostypes.Request, msg cosmostypes.Msg) (hash.Hash, error) {
+	switch msg := msg.(type) {
+	case msgCreateProcess:
+		p, err := s.Create(req, &msg)
+		if err != nil {
+			return nil, err
+		}
+		return p.Hash, nil
+	case msgDeleteProcess:
+		if err := s.ownership.Delete(req, msg.Owner, msg.Request.Hash); err != nil {
```
NicolasMahe: please create a dedicated function `Delete` even if it's small. Let's not put business logic directly in the handler.

sdk/ownership/backend.go
```diff
@@ -14,20 +14,25 @@ import (
 
 const backendName = "ownership"
 
+const storeHashKey = "ownership_hash"
+
 // Backend is the ownership backend.
 type Backend struct {
-	storeKey *cosmostypes.KVStoreKey
+	storeKey     *cosmostypes.KVStoreKey
+	storeHashKey *cosmostypes.KVStoreKey
```
NicolasMahe: why do you add a new store?
for what i understand it's acting as a reverse index to get the ownership from a resource hash?
I'm not against this, but it's too early to implement it and you didn't implement any getter that use it..

sdk/process/backend.go
```diff
@@ -0,0 +1,146 @@
...
+			if _, err := s.instance.Get(req, n.Event.InstanceHash); err != nil {
+				return nil, err
+			}
+		case *process.Process_Node_Task_:
+			if _, err := s.instance.Get(req, n.Task.InstanceHash); err != nil {
+				return nil, err
+			}
+		}
+	}
+
```
NicolasMahe: Don't forget to call `p.Validate` function:
```suggestion

	if err := p.Validate(); err != nil {	
		return nil, err	
	}

```

process/codec.go
```diff
@@ -0,0 +1,19 @@
+package process
+
+import (
+	"github.com/mesg-foundation/engine/codec"
+)
+
+func init() {
+	codec.RegisterInterface((*isProcess_Node_Type)(nil), nil)
```
NicolasMahe:
```
panic: Unregistered interface process.isProcess_Node_Map_Output_Value
```

sdk/process/backend.go
krhubert: The validation is done in ValidationBasic on msg.

sdk/ownership/backend.go
krhubert: >  I'm not against this, but it's too early to implement it and you didn't implement any getter that use it..

Too early for what? I just didn't want to iterate over the store because it will become bottleneck very fast. Also we don't need any getter, this store is here only for internal purposes.

Also I didn't want to copy functions to get the processes list from the store. I even would get rid of CreateServeiceOnwership.

>  And that's not the purpose of this PR ;) (you already changed the ownership structure)

The purpose of this pr is to move process to cosmos-sdk and create Ownership for them and I did it, just different way.

sdk/ownership/backend.go
```diff
@@ -14,20 +14,25 @@ import (
...
+	storeKey               *cosmostypes.KVStoreKey
+	storeIndexRessourceKey *cosmostypes.KVStoreKey
 }
 
 // NewBackend returns the backend of the ownership sdk.
 func NewBackend(appFactory *cosmos.AppFactory) *Backend {
 	backend := &Backend{
-		storeKey: cosmostypes.NewKVStoreKey(backendName),
+		storeKey:               cosmostypes.NewKVStoreKey(backendName),
+		storeIndexRessourceKey: cosmostypes.NewKVStoreKey(storeIndexRessourceKey),
```
NicolasMahe:
```suggestion
		storeIndexRessourceKey: cosmostypes.NewKVStoreKey(backendName + ".index.resourceHash"),
```

sdk/ownership/backend.go
```diff
@@ -14,20 +14,25 @@ import (
 
 const backendName = "ownership"
 
+const storeIndexRessourceKey = backendName + ".index.resourceHash"
```
NicolasMahe:
```suggestion
```

sdk/ownership/backend.go
NicolasMahe: it's confusing to have 2 variable with the same name. the actual string of the key is only use once.

sdk/ownership/backend.go
```diff
@@ -16,18 +16,21 @@ const backendName = "ownership"
...
+	storeKey               *cosmostypes.KVStoreKey
+	storeIndexRessourceKey *cosmostypes.KVStoreKey
 }
 
 // NewBackend returns the backend of the ownership sdk.
 func NewBackend(appFactory *cosmos.AppFactory) *Backend {
 	backend := &Backend{
-		storeKey: cosmostypes.NewKVStoreKey(backendName),
+		storeKey:               cosmostypes.NewKVStoreKey(backendName),
+		storeIndexRessourceKey: cosmostypes.NewKVStoreKey(backendName + ".index.resourceHash"),
```
NicolasMahe: After some debate with @antho1404 we decide to not implement any index for now.
Let's give us the opportunity to improve the performance and efficiency later.

Also, the relationship between resources and owners is many to many (even if for now it's one to one because of logic limitation). This index you implemented is only one to one.

sdk/ownership/backend.go
```diff
@@ -73,12 +68,28 @@ func (s *Backend) CreateServiceOwnership(request cosmostypes.Request, serviceHas
 	return ownership, nil
 }
 
+// Delete deletes a ownership.
+func (s *Backend) Delete(req cosmostypes.Request, owner cosmostypes.AccAddress, resourceHash hash.Hash) error {
```
NicolasMahe: owner is not used in this function.
it should either delete only the ownership of this specific owner or delete all the ownership associated to a resource. in this last case, remove owner from the parameter.

NicolasMahe:
Let's wait for issue https://github.com/mesg-foundation/engine/issues/1427 to be closed before merging this PR

sdk/process/backend.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	p.Hash = hash.Dump(p)
+	if store.Has(p.Hash) {
+		return nil, fmt.Errorf("process %q already exists", p.Hash)
+	}
+
+	for _, node := range p.Nodes {
+		switch n := node.Type.(type) {
+		case *process.Process_Node_Result_:
+			if _, err := s.instance.Get(req, n.Result.InstanceHash); err != nil {
```
antho1404: We could use the exists function here

```suggestion
			if _, err := s.instance.Exists(req, n.Result.InstanceHash); err != nil {
```

PR #1533, opened 2019-11-28
Split execution to execution request and result

NicolasMahe:
First step in changing the structure of data of execution.
This PR splits internally the executions in execution request and execution result. It transforms the types in the backend SDK.
The behavior of execution is still the same: 1 execution request for 1 execution result max.


sdk/execution/backend.go
```diff
@@ -175,29 +184,59 @@ func (s *Backend) validateExecutionOutput(request cosmostypes.Request, instanceH
...
 		}
-		execs = append(execs, exec)
+		execs = append(execs, execution.ToExecution(execReq, execRes))
 		iter.Next()
 	}
 	iter.Close()
 	return execs, nil
 }
+
+func (s *Backend) getExecutionResult(resStore cosmostypes.KVStore, requestHash hash.Hash) (*execution.ExecutionResult, error) {
```
NicolasMahe: i will implement an index store like on PR https://github.com/mesg-foundation/engine/pull/1527/files

antho1404:
Should we get rid of Execution directly and transform it into ExecutionRequest. I'm afraid that trying to keep both will make things a lot more complicated. What do you think?

NicolasMahe:
> Should we get rid of Execution directly and transform it into ExecutionRequest. I'm afraid that trying to keep both will make things a lot more complicated. What do you think?

ok, let's do it now!
I will update this PR.

PR #1536, opened 2019-11-28
Move node.*.key into node and rename process.key to name

### LINKED ISSUE #1534: Rename process.key to process.name
Rename process.key to process.name to better show that it's only used as an help for human to know what is the process same purpose as service.name.

### LINKED ISSUE #1535: Move Result.key, Event.key, Task.key, Map.key and Filter.key to Node.key
Move Result.key, Event.key, Task.key, Map.key and Filter.key to Node.key.

Every node type redeclare a parameter key. it should centralized in node.
this will require a bit of modification in the orchestrator (like passing the node.key as parameter of some functions) but nothing complicated.

krhubert:
close #1534 
close #1535


orchestrator/orchestrator.go
```diff
@@ -121,12 +121,12 @@ func (s *Orchestrator) execute(filter func(wf *process.Process, node *process.Pr
 
 func (s *Orchestrator) executeNode(wf *process.Process, n *process.Process_Node, exec *execution.Execution, event *event.Event, data *types.Struct) error {
 	logrus.WithField("module", "orchestrator").
-		WithField("nodeID", n.ID()).
+		WithField("nodeID", n.Key).
```
NicolasMahe:
```suggestion
		WithField("node.key", n.Key).
```

orchestrator/orchestrator.go
```diff
@@ -215,7 +215,7 @@ func (s *Orchestrator) processTask(task *process.Process_Node_Task, wf *process.
 		ProcessHash:  wf.Hash,
 		EventHash:    eventHash,
 		ParentHash:   execHash,
-		StepID:       task.Key,
+		StepID:       nodeKey,
```
NicolasMahe: we should rename execution's StepID to stepKey.
@antho1404 what do you think?

process/graph.go
```diff
@@ -36,7 +36,7 @@ func (g Process) FindNodes(filter func(n *Process_Node) bool) []*Process_Node {
 // FindNode return a specific node in a graph identifies by its ID. Returns an error if there is no match or multiple matches
 func (g Process) FindNode(id string) (*Process_Node, error) {
```
NicolasMahe:
```suggestion
func (g Process) FindNode(key string) (*Process_Node, error) {
```

orchestrator/orchestrator.go
antho1404: I'm ok to change it, but why not `nodeKey` like that we always talk about the same thing.

NicolasMahe: `nodeKey` 👍

protobuf/types/process.proto
```diff
@@ -138,37 +118,37 @@ message Process {
...
+    // Key that identifies the node.
+    string key = 1 [
+      (gogoproto.moretags) = 'hash:"name:1" validate:"required"'
+    ];
+
     oneof type {
       // Result is a trigger that listens for a specific result.
-      Result result = 1 [
+      Result result = 2 [
         (gogoproto.moretags) = 'hash:"name:1"'
```
antho1404: if you change the number in proto, don't forget to also change the hash
```suggestion
        (gogoproto.moretags) = 'hash:"name:2"'
```

protobuf/types/process.proto
```diff
@@ -138,38 +118,38 @@ message Process {
...
+    // Key that identifies the node.
+    string key = 1 [
+      (gogoproto.moretags) = 'hash:"name:1" validate:"required"'
+    ];
+
     oneof type {
       // Result is a trigger that listens for a specific result.
-      Result result = 1 [
-        (gogoproto.moretags) = 'hash:"name:1"'
+      Result result = 2 [
```
NicolasMahe: i think you don't need to start from 0 in a oneof

NicolasMahe:
Let's wait for issue https://github.com/mesg-foundation/engine/issues/1427 to be closed before merging this PR

protobuf/types/process.proto
krhubert: from 0?

NicolasMahe: i mean from 1

PR #1540, opened 2019-11-29
Implement nested data on process map

krhubert:
dep https://github.com/mesg-foundation/engine/pull/1536
close https://github.com/mesg-foundation/engine/issues/1304

Changes in process.proto:
- add MapOutput and ListOutput messages in the Map.Output
- Remove Map.Output.Key (not every output has a key, for example array do not have)
- change `repeated Output outputs`  to `map<string, Output> outputs`

Those changes are required to implement both object and array. Doing only NestedObject as @NicolasMahe  suggested is not engouth to implement array type. I don't see any other way to do that.

Missing:

- manual and e2e tests.


NicolasMahe:
- [ ] add e2e tests with nested data on map (dep on https://github.com/mesg-foundation/engine/pull/1547)

protobuf/types/process.proto
```diff
@@ -69,25 +86,43 @@ message Process {
...
+
+          bool bool_const = 4 [
+            (gogoproto.moretags) = 'hash:"name:4"'
+          ];
+
           // Input defined as reference.
-          Reference ref = 2 [
-            (gogoproto.moretags) = 'hash:"name:2" validate:"required_without=Constant"'
+          Reference ref = 5 [
+            (gogoproto.moretags) = 'hash:"name:5" validate:"required_without=Null Constant List Map"'
```
NicolasMahe: Is it working and necessary: `required_without=Null Constant List Map`? it's already in a oneof.

NicolasMahe: Add the key name are wrong..

antho1404: If we receive the data from grpc that shouldn't be necessary but if we use directly the struct then this will trigger and error when we validate the struct.

antho1404: @krhubert can you fix the validation here, we need to add the string/double/bool

PR #1543, opened 2019-12-02
Add lock to cosmos lazy keybase

krhubert:
Adding lock is temp workaroudn until KeyKeyring won't be released.

close https://github.com/mesg-foundation/engine/issues/1539

### DIFF
--- a/cosmos/keybase.go
+++ b/cosmos/keybase.go
@@ -1,10 +1,15 @@
 package cosmos
 
 import (
+	"sync"
+
 	clientkey "github.com/cosmos/cosmos-sdk/client/keys"
 	"github.com/cosmos/cosmos-sdk/crypto/keys"
+	"github.com/cosmos/cosmos-sdk/crypto/keys/hd"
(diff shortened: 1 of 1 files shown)


cosmos/keybase.go
```diff
@@ -1,17 +1,24 @@
...
+	"github.com/cosmos/cosmos-sdk/types"
 	bip39 "github.com/cosmos/go-bip39"
+	"github.com/tendermint/tendermint/crypto"
 )
 
 const mnemonicEntropySize = 256
 
 // Keybase is a standard cosmos keybase.
 type Keybase struct {
 	keys.Keybase
```
antho1404: let's not inherit this if we redefine all the functions and let's make it private too
```suggestion
	keybase keys.Keybase
```

NicolasMahe: agree with antho. Otherwise we could "forget" to implement the lock on a function..

krhubert: I left keybase composed because someone might still want to use it. I can make it private.

PR #1545, opened 2019-12-03
Add path system to reference nested data in process

krhubert:
+ small refactor

dep https://github.com/mesg-foundation/engine/pull/1540
close https://github.com/mesg-foundation/engine/issues/1306


orchestrator/orchestrator.go
```diff
@@ -223,3 +260,42 @@ func (s *Orchestrator) processTask(nodeKey string, task *process.Process_Node_Ta
 	}, s.accountName, s.accountPassword)
 	return err
 }
+
+func query(data *types.Struct, path *process.Process_Node_Map_Output_Reference_Path) (*types.Value, error) {
```
NicolasMahe: Let's rename query:
```suggestion
func resolveRef(data *types.Struct, path *process.Process_Node_Map_Output_Reference_Path) (*types.Value, error) {
```

orchestrator/orchestrator.go
```diff
@@ -223,3 +260,42 @@ func (s *Orchestrator) processTask(nodeKey string, task *process.Process_Node_Ta
...
+
+	v, ok = data.Fields[key.Key]
+	if !ok {
+		return nil, fmt.Errorf("orchestrator: key %s not found", key.Key)
+	}
+
+	for p := path.Path; p.Path != nil; p = p.Path {
+		switch s := p.Selector.(type) {
+		case *process.Process_Node_Map_Output_Reference_Path_Key:
+			str, ok := v.GetKind().(*types.Value_StructValue)
```
NicolasMahe: You should use the helper function generated by proto. something like:
```suggestion
			str, ok := v.GetStructValue()
```

orchestrator/orchestrator.go
```diff
@@ -223,3 +260,42 @@ func (s *Orchestrator) processTask(nodeKey string, task *process.Process_Node_Ta
...
+	}
+
+	v, ok = data.Fields[key.Key]
+	if !ok {
+		return nil, fmt.Errorf("orchestrator: key %s not found", key.Key)
+	}
+
+	for p := path.Path; p.Path != nil; p = p.Path {
+		switch s := p.Selector.(type) {
+		case *process.Process_Node_Map_Output_Reference_Path_Key:
```
NicolasMahe: you could transform this case to a simple if / else if / else.
At least, add a `default` that return an error.

NicolasMahe:
optional path test added

NicolasMahe:
TODO:
- [ ] add e2e test on nested ref. the only e2e test is testing only 1 level of nested ref.

orchestrator/orchestrator.go
```diff
@@ -224,9 +225,9 @@ func (s *Orchestrator) resolveInput(wfHash hash.Hash, exec *execution.Execution,
 		if err != nil {
 			return nil, err
 		}
-		return s.resolveInput(wfHash, parent, nodeKey, outputKey)
+		return s.resolveInput(wfHash, parent, nodeKey, path)
```
antho1404: `resolveInput` wasn't resolving any path, I just added a commit with the fix

PR #1551, opened 2019-12-09
[WIP] Handle get reference result 

krhubert:
dep https://github.com/mesg-foundation/engine/pull/1545


orchestrator/orchestrator.go
```diff
@@ -230,6 +233,20 @@ func (s *Orchestrator) resolveInput(wfHash hash.Hash, exec *execution.Execution,
 	return exec.Outputs.Fields[outputKey], nil
 }
 
+func (s *Orchestrator) resolveResult(wfHash hash.Hash, exec *execution.Execution, instanceHash hash.Hash, outputKey string) (*types.Value, error) {
+	if !wfHash.Equal(exec.ProcessHash) {
```
NicolasMahe: i'm not sure but i think `exec.ProcessHash` will never contains the same `wfHash` as it is created manually or by another process.

orchestrator/orchestrator.go
```diff
@@ -210,6 +210,9 @@ func (s *Orchestrator) outputToValue(output *process.Process_Node_Map_Output, ou
 		if node.GetTask() != nil {
 			return s.resolveInput(wf.Hash, exec, v.Ref.NodeKey, outputKey)
 		}
+		if result := node.GetResult(); result != nil {
+			return s.resolveInput(wf.Hash, exec, v.Ref.NodeKey, outputKey)
```
NicolasMahe: should be `resolveResult` i guess

orchestrator/orchestrator.go
```diff
@@ -230,6 +233,20 @@ func (s *Orchestrator) resolveInput(wfHash hash.Hash, exec *execution.Execution,
 	return exec.Outputs.Fields[outputKey], nil
 }
 
+func (s *Orchestrator) resolveResult(wfHash hash.Hash, exec *execution.Execution, instanceHash hash.Hash, outputKey string) (*types.Value, error) {
+	if !wfHash.Equal(exec.ProcessHash) {
+		return nil, fmt.Errorf("reference's result nodeKey not found")
+	}
+	if !exec.InstanceHash.Equal(instanceHash) {
```
NicolasMahe: i think this is not bad but you should also check that the `TaskKey` are the right one.

orchestrator/orchestrator.go
krhubert: Can you add test about resolve result. It will help me a lot with implementing that logic and future refactor.

NicolasMahe:
Closing this PR as the feature changed

PR #1552, opened 2019-12-09
Add cosmos cli

NicolasMahe:
Usage:
```bash
go run ./cosmos/cli/main.go
```

You have to setup the chain-id (`mesg-dev-chain`) correctly:
From tuto https://tutorials.cosmos.network/nameservice/tutorial/21-build-run.html:
```
# Configure your CLI to eliminate need for chain-id flag
go run ./cosmos/cli/main.go config chain-id mesg-dev-chain
go run ./cosmos/cli/main.go config output json
go run ./cosmos/cli/main.go config indent true
go run ./cosmos/cli/main.go config trust-node true
```

closes https://github.com/mesg-foundation/engine/issues/1563


NicolasMahe:
todo:
- [ ] add CI build and upload to github asset

krhubert:
close https://github.com/mesg-foundation/engine/issues/1563

Makefile
```diff
@@ -51,6 +51,21 @@ dep:
...
+get-ghr:
+	go get -u github.com/tcnksm/ghr
+
+build-cmd-cosmos: dep
+	go build -mod=readonly -o ./bin/mesg-cosmos ./cmd/mesg-cosmos/main.go
+
+publish-cmd-cosmos-dev: get-ghr
+	ghr -u mesg-foundation -r engine -delete -prerelease -n "Developer Release" -b "Warning - this is a developer release, use it only if you know what are doing. Make sure to pull the latest \`mesg/engine:dev\` image. \`\`\`docker pull mesg/engine:dev\`\`\`" release-dev ./bin/mesg-cosmos
+
+publish-cmd-cosmos-prod: get-ghr
```
antho1404: I don't really like the idea to have a release on the engine repo that is not the engine but a CLI to use the engine. It doesn't make much sense.
I see multiple possibilities:
- We could have a separated repo for that but that's not the best
- We could just upload the binary on an s3 or something like that, this way it's not attached to a release in github where the version will not match anything)
- We could just build it

I like the 3rd option as this CLI is for debug or really advance use so I don't see a real need to publish it. This way we can use it as an internal/advanced tool and if there are features that are really necessary, merge them into the `mesg-cli`

What do you think about that?

NicolasMahe: I would like to have an already compiled CLI done by a CI task so we don't have to manage it.
Uploading it to github is simple but yes, maybe not the best place as one of the CLI is a pure client. But at the end, it's a specific client for the engine.
Anyway, I don't want to loose time with this CLI. I will remove the CI task and open a new PR if i feel the need for it.

PR #1555, opened 2019-12-12
Remove account sdk and credential system

NicolasMahe:
Remove account sdk and credential over gRPC to simplify the codebase as we don't use it.

First step of https://github.com/mesg-foundation/engine/issues/1538
Could be reverted/inspired for implementation of https://github.com/mesg-foundation/engine/issues/1484


sdk/execution/sdk.go
```diff
@@ -14,19 +15,19 @@ import (
...
 	runnerSDK   *runnersdk.SDK
 }
 
 // New returns the execution sdk.
-func New(client *cosmos.Client, kb *cosmos.Keybase, serviceSDK *servicesdk.SDK, instanceSDK *instancesdk.SDK, runnerSDK *runnersdk.SDK) *SDK {
+func New(client *cosmos.Client, accAddress cosmostypes.AccAddress, serviceSDK *servicesdk.SDK, instanceSDK *instancesdk.SDK, runnerSDK *runnersdk.SDK) *SDK {
 	sdk := &SDK{
 		client:      client,
-		kb:          kb,
+		accAddress:  accAddress,
```
antho1404: We should get the `accAddress` from the `client` directly with something like `client.GetSignerAddress`

cosmos/client.go
```diff
@@ -68,26 +72,25 @@ func (c *Client) QueryWithData(path string, data []byte) ([]byte, int64, error)
 }
 
 // BuildAndBroadcastMsg builds and signs message and broadcast it to node.
-func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg, accName, accPassword string) (*abci.ResponseDeliverTx, error) {
-	info, err := c.kb.Get(accName)
+func (c *Client) BuildAndBroadcastMsg(msg sdktypes.Msg) (*abci.ResponseDeliverTx, error) {
+	info, err := c.kb.Get(c.accName)
```
antho1404:
```suggestion
	info, err := c.GetAccount()
```

Issue #1537, opened 2019-11-29
 Execute task service error on influxdb


blackinno:
**Describe the bug** 
Listen to the transfer event on ERC-20 then save data to influxdb but all data from transfer events aren't saved to the database.

**To Reproduce**
Run this script :
```js
const Application = require('@mesg/application')

const mesg = new Application()

async function main() {
  const erc20 = await mesg.resolve('ethereum-erc20')
  const runnerHash = await mesg.resolveRunner('influxdb')

  mesg.listenEvent({ filter: { instanceHash: erc20, key: 'transfer' } }).on('data', async event => {
    const transfer = mesg.decodeData(event.data)
    console.log('transfer type: ', typeof transfer, ' transfer', transfer)
    try {
      const result = await mesg.executeTask({
        executorHash: runnerHash,
        eventHash: event.hash,
        taskKey: 'write',
        inputs: mesg.encodeData({
          measurement: 'transfer',
          tags: {
            block: transfer.blockNumber,
            address: transfer.contractAddress,
            from: transfer.from,
            to: transfer.to
          },
          fields: { value: parseFloat(transfer.value) }
        })
      })
      console.log('a result received:', result)
    } catch (error) {
       console.error('an error has occurred:', error)
    }
  })
  console.log('application is running and listening for transfer events')
}

main()
```
**Screenshots**

*log from influxdb service*:
<img width="840" alt="Screen Shot 2562-11-29 at 10 10 23" src="https://user-images.githubusercontent.com/16553272/69841599-7b37c880-1292-11ea-8067-81121211105d.png">

*log from script*:
<img width="1066" alt="Screen Shot 2562-11-29 at 10 26 00" src="https://user-images.githubusercontent.com/16553272/69841677-c2be5480-1292-11ea-84a4-d3236c132c89.png">

NicolasMahe:
has the problem been resolved with https://github.com/mesg-foundation/engine/issues/1539 ?

antho1404:
No, there is still an error because of the sequence. It should be fixed when https://github.com/mesg-foundation/engine/issues/1538 is fixed.

antho1404:
I tested this with https://github.com/mesg-foundation/engine/pull/1556 and the problem is solved.
This issue will be fixed when the PR is merged

PR #1556, opened 2019-12-12
Fix account sequence when signing multiple transactions

NicolasMahe:
This PR save the used sequence in memory and test it against the one from the blockchain to return the max.

Diff without dependent PR: https://github.com/mesg-foundation/engine/pull/1556/files/9dc55511dc71ef07de366741879edcc41ca82212..e1f69c7ce906ab16e4a8232f757b48bc7f602ac6

Dependent on https://github.com/mesg-foundation/engine/pull/1555
Closes https://github.com/mesg-foundation/engine/issues/1538
Should fix https://github.com/mesg-foundation/engine/issues/1537


cosmos/client.go
```diff
@@ -172,3 +170,25 @@ func (c *Client) Stream(ctx context.Context, query string) (chan hash.Hash, chan
...
+	// get remote seq
+	if acc, err := auth.NewAccountRetriever(c).GetAccount(accKb.GetAddress()); err == nil {
+		// replace seq if sup
+		if acc.GetSequence() > seq {
+			seq = acc.GetSequence()
+		}
+	}
+	// increment local sequence
+	sequence = seq + 1
+	return seq, nil
```
krhubert: s/ 	return seq, nil / 	return sequence, nil /

As we return greather value ?

krhubert: Also this function can be simplified a bit like:

```go
func (c *Client) getAccountSequenceAndIncrement() (uint64, error) {
  accKb, err := c.kb.Get(c.accName)
  if err != nil {
    return 0, err
  }
  acc, err := auth.NewAccountRetriever(c).GetAccount(accKb.GetAddress())
  if err != nil {
    return 0, err
  }

  getSequenceMutex.Lock()
  sequence = xmath.MaxUint64(acc.GetSequence(), sequence) + 1
  getSequenceMutex.Unlock()
  return sequence, nil
}
```

e2e/execution_test.go
```diff
@@ -89,6 +89,87 @@ func testExecution(t *testing.T) {
...
+			exec, err := stream.Recv()
+			require.NoError(t, err)
+			require.Equal(t, executionHash2, exec.Hash)
+			require.Equal(t, taskKey, exec.TaskKey)
+			require.Equal(t, hash.Int(3), exec.EventHash)
+			require.Equal(t, executorHash, exec.ExecutorHash)
+			require.Equal(t, execution.Status_Completed, exec.Status)
+			require.True(t, inputs.Equal(exec.Inputs))
+			require.Equal(t, "test", exec.Outputs.Fields["msg"].GetStringValue())
+			require.NotEmpty(t, exec.Outputs.Fields["timestamp"].GetNumberValue())
```
krhubert: I think we don't need to check so many thing there. This test is "over-asserted" and should only compare if two executions were completed. None of  `in progress` tests or any tests for hashes are needed.

NicolasMahe: Is it really the assertion that you don't like or the number of lines it uses? I was thinking to create assertion function for executions.

cosmos/client.go
NicolasMahe: > s/ return seq, nil / return sequence, nil /
> 
> As we return greather value ?

The sequence is increased for the next use. I respect the default implementation of cosmos GetSequence system. First time you get the sequence, it returns 0. Next time (after 1 block) it returns 1.

NicolasMahe: > ```go
> xmath
> ```

`auth.NewAccountRetriever` returns an error if the account doesn't exist (and it's the case when it has not received any token yet). So it need to bypass the error.
But otherwise the implementation looks correct. I will improve mine.

krhubert: > where is xmath.MaxUint64 implemented?

It isn't but it great helper function for our `x` dir :)

> auth.NewAccountRetriever returns an error if the account doesn't exist (and it's the case when it has not received any token yet). 

Oh right, so maybe we can keep the way you wrote :)

e2e/execution_test.go
krhubert: > Is it really the assertion that you don't like or the number of lines it uses? 

Both. The number of lines is tremendous here and many of the assertions are not needed in that test. For me, the test should focus on what is actually tested - nothing more, nothing less.

So:

1) task1 inputs are optional so we don't need to create it here, we can leave it empty
2) receiving in-progress is not needed - we can only check for a completed task
3) assertion of every field is not needed - execution hash probably would be enough (or use Equal function from protobuf as you mentioned)

cosmos/client.go
```diff
@@ -172,3 +170,25 @@ func (c *Client) Stream(ctx context.Context, query string) (chan hash.Hash, chan
...
+	getSequenceMutex.Lock()
+	defer getSequenceMutex.Unlock()
+	accKb, err := c.kb.Get(c.accName)
+	if err != nil {
+		return 0, err
+	}
+	// get local seq
+	seq := sequence
+	// get remote seq
+	if acc, err := auth.NewAccountRetriever(c).GetAccount(accKb.GetAddress()); err == nil {
```
antho1404: we should store in memory the account itself and not the sequence

antho1404:
Broadcast lock added, this solves the issue with the invalid sequence but there is a new problem coming.

When testing with #1537 executions are created correctly but it seems that they are not all processed.

Running this test a few seconds I have:
- 179 executions created
- 163 executions in progress
- 16 executions completed

Even if I wait it doesn't seem to process any more transactions.

NicolasMahe:
> Broadcast lock added, this solves the issue with the invalid sequence but there is a new problem coming.
> 
> When testing with #1537 executions are created correctly but it seems that they are not all processed.
> 
> Running this test a few seconds I have:
> 
> * 179 executions created
> * 163 executions in progress
> * 16 executions completed
> 
> Even if I wait it doesn't seem to process any more transactions.

I will write an e2e tests with thousands of execution to see what's going on

cosmos/client.go
```diff
@@ -162,6 +154,8 @@ func (c *Client) Stream(ctx context.Context, query string) (chan hash.Hash, chan
 				} else {
 					hashC <- hash
 				}
+			case <-eventStream.Cancelled():
+				errC <- eventStream.Err()
```
krhubert: Shoudn't we break a loop here?

cosmos/client.go
```diff
@@ -173,7 +167,48 @@ func (c *Client) Stream(ctx context.Context, query string) (chan hash.Hash, chan
...
+			return nil, err
+		}
+		c.acc = auth.NewBaseAccount(
+			accKb.GetAddress(),
+			nil,
+			accKb.GetPubKey(),
+			AccNumber,
+			0,
+		)
+	}
```
krhubert: This could be moved to a helper method.

cosmos/client.go
```diff
@@ -173,7 +167,48 @@ func (c *Client) Stream(ctx context.Context, query string) (chan hash.Hash, chan
...
+
+// Sign signs a msg and return a tendermint tx.
+func (c *Client) sign(msg sdktypes.Msg) (tenderminttypes.Tx, error) {
+	acc, err := c.GetAccount()
+	if err != nil {
+		return nil, err
+	}
+
+	sequence := acc.GetSequence()
+	acc.SetSequence(acc.GetSequence() + 1)
```
krhubert: I'm not sure, but I think lock might be needed here. We retrive account and the between GetAccount and GetSequence someone might call it. Possible solution is to return sequence from a helper function GetAccountSeq or protect this with mutex.

e2e/execution_test.go
```diff
@@ -89,6 +101,87 @@ func testExecution(t *testing.T) {
 		})
 	})
 
+	t.Run("double execution", func(t *testing.T) {
```
krhubert:
```suggestion
	t.Run("??? execution", func(t *testing.T) {
```

What this test is doing - the description dosen't give a hint what's going on here.

e2e/execution_test.go
```diff
@@ -89,6 +101,87 @@ func testExecution(t *testing.T) {
...
+			require.Equal(t, executionHash2, exec.Hash)
+			require.Equal(t, taskKey, exec.TaskKey)
+			require.Equal(t, hash.Int(3), exec.EventHash)
+			require.Equal(t, executorHash, exec.ExecutorHash)
+			require.Equal(t, execution.Status_Completed, exec.Status)
+			require.True(t, inputs.Equal(exec.Inputs))
+			require.Equal(t, "test", exec.Outputs.Fields["msg"].GetStringValue())
+			require.NotEmpty(t, exec.Outputs.Fields["timestamp"].GetNumberValue())
+		})
+	})
```
krhubert: Do we really need all those checks for this particular test?

e2e/execution_test.go
```diff
@@ -168,6 +261,71 @@ func testExecution(t *testing.T) {
...
+		t.Run("create executions", func(t *testing.T) {
+			wg := sync.WaitGroup{}
+			var mutex sync.Mutex
+			wg.Add(n)
+			for i := 1; i <= n; i++ {
+				go func(i int) {
+					defer wg.Done()
+					resp, err := client.ExecutionClient.Create(context.Background(), &pb.CreateExecutionRequest{
+						TaskKey:      taskKey,
+						EventHash:    hash.Int(1111 + i),
```
krhubert:
```suggestion
						EventHash:    hash.Int(i),
```

- start i from 1000 (with comment)
- or maybe user hash.Random which will fit it here.

e2e/execution_test.go
```diff
@@ -168,6 +261,71 @@ func testExecution(t *testing.T) {
...
+						ExecutorHash: executorHash,
+						Inputs:       inputs,
+					})
+					require.NoError(t, err)
+					mutex.Lock()
+					defer mutex.Unlock()
+					require.NotContains(t, executions, resp.Hash)
+					executions = append(executions, resp.Hash)
+					fmt.Printf("execution %d created\n", i)
```
krhubert:
```suggestion
```

e2e/execution_test.go
```diff
@@ -168,6 +261,71 @@ func testExecution(t *testing.T) {
...
+						TaskKey:      taskKey,
+						EventHash:    hash.Int(1111 + i),
+						ExecutorHash: executorHash,
+						Inputs:       inputs,
+					})
+					require.NoError(t, err)
+					mutex.Lock()
+					defer mutex.Unlock()
+					require.NotContains(t, executions, resp.Hash)
+					executions = append(executions, resp.Hash)
```
krhubert: mutex is not needed here. we can safely make a slice `make([]Executions, n)` and write to every slice index without mutex. After all goroutines end, check for duplication can be performed.

e2e/execution_test.go
```diff
@@ -168,6 +261,71 @@ func testExecution(t *testing.T) {
...
+		})
+		t.Run("check in progress", func(t *testing.T) {
+			execs := make([]hash.Hash, 0)
+			for i := 1; i <= n; i++ {
+				exec, err := streamInProgress.Recv()
+				require.NoError(t, err)
+				require.Contains(t, executions, exec.Hash)
+				require.NotContains(t, execs, exec.Hash)
+				execs = append(execs, exec.Hash)
+				fmt.Printf("execution %d in progress\n", i)
```
krhubert:
```suggestion
```

e2e/execution_test.go
```diff
@@ -168,6 +261,71 @@ func testExecution(t *testing.T) {
...
+		})
+		t.Run("check completed", func(t *testing.T) {
+			execs := make([]hash.Hash, 0)
+			for i := 1; i <= n; i++ {
+				exec, err := streamCompleted.Recv()
+				require.NoError(t, err)
+				require.Contains(t, executions, exec.Hash)
+				require.NotContains(t, execs, exec.Hash)
+				execs = append(execs, exec.Hash)
+				fmt.Printf("execution %d completed\n", i)
```
krhubert:
```suggestion
```

e2e/execution_test.go
```diff
@@ -168,6 +261,71 @@ func testExecution(t *testing.T) {
...
+			execs := make([]hash.Hash, 0)
+			for i := 1; i <= n; i++ {
+				exec, err := streamInProgress.Recv()
+				require.NoError(t, err)
+				require.Contains(t, executions, exec.Hash)
+				require.NotContains(t, execs, exec.Hash)
+				execs = append(execs, exec.Hash)
+				fmt.Printf("execution %d in progress\n", i)
+			}
+			require.Len(t, execs, n)
```
krhubert:
```suggestion
```

This check will be always true

e2e/execution_test.go
```diff
@@ -168,6 +261,71 @@ func testExecution(t *testing.T) {
...
+			execs := make([]hash.Hash, 0)
+			for i := 1; i <= n; i++ {
+				exec, err := streamCompleted.Recv()
+				require.NoError(t, err)
+				require.Contains(t, executions, exec.Hash)
+				require.NotContains(t, execs, exec.Hash)
+				execs = append(execs, exec.Hash)
+				fmt.Printf("execution %d completed\n", i)
+			}
+			require.Len(t, execs, n)
```
krhubert:
```suggestion
```


This check will be always true

e2e/execution_test.go
```diff
@@ -168,6 +261,71 @@ func testExecution(t *testing.T) {
...
+				Fields: map[string]*types.Value{
+					"msg": {
+						Kind: &types.Value_StringValue{
+							StringValue: "test",
+						},
+					},
+				},
+			}
+		)
+		t.Run("create executions", func(t *testing.T) {
```
krhubert:
```suggestion
		t.Run("??? executions", func(t *testing.T) {
```

More description, as this test is not only about create exeucitons

cosmos/client.go
NicolasMahe: Yes you're right. We actually plan to move all the account with signing logic to a dedicated struct in a future PR.

cosmos/client.go
NicolasMahe: The lock is in the function `BuildAndBroadcastMsg` around the call of `sign`. This may not be the best but we plan to move all the account + sign logic to a dedicated structure.

NicolasMahe: issue created: <https://github.com/mesg-foundation/engine/issues/1565>

cosmos/client.go
NicolasMahe: issue created: <https://github.com/mesg-foundation/engine/issues/1565>

e2e/execution_test.go
NicolasMahe: i removed this test as its redundant with "many executions in parallel" test

e2e/execution_test.go
NicolasMahe: i removed this test as its redundant with "many executions in parallel" test

e2e/execution_test.go
NicolasMahe: this specific sub-test is only creating the executions.

e2e/execution_test.go
NicolasMahe: i would really like to keep the check on the length of arrays so we verify that n executions has been created, then in progress, then completed.
This is a way to make sure the stream is receiving ALL expected events.

e2e/execution_test.go
NicolasMahe: see https://github.com/mesg-foundation/engine/pull/1556/files#r359079391

e2e/execution_test.go
NicolasMahe: see https://github.com/mesg-foundation/engine/pull/1556/files#r359079391

NicolasMahe:
@krhubert can you do a last review and merge if your ok :)

PR #1558, opened 2019-12-12
Add process refKey

krhubert:
close https://github.com/mesg-foundation/engine/issues/1557

From what I see in this implementation `nodeKey` (in Reference) was also needed to find the right execution, @NicolasMahe any thoughts on that?

+ I've updated all messages id, because the counting was started form 2


protobuf/types/process.proto
```diff
@@ -14,33 +14,38 @@ message Process {
...
-        (gogoproto.moretags) = 'hash:"name:3" validate:"printascii,required"'
+      string eventKey = 2 [
+        (gogoproto.moretags) = 'hash:"name:2" validate:"printascii,required"'
       ];
     }
 
     message Task {
+      // RefKey of this execution's node. Must to set to reference an output of this task from a map node.
+      string refKey = 1 [
+        (gogoproto.moretags) = 'hash:"name:1" validate:"required"'
```
NicolasMahe: it's not required.
```suggestion
        (gogoproto.moretags) = 'hash:"name:1" validate:"printascii"'
```

protobuf/types/process.proto
```diff
@@ -75,8 +80,8 @@ message Process {
         }
 
         message Reference {
-          // Key of the node in the graph. If empty, will be using the src of the edge.
-          string nodeKey = 1 [
+          // RefKey of the execution's node in the graph.
+          string refKey = 1 [
             (gogoproto.moretags) = 'hash:"name:1" validate:"required"'
```
NicolasMahe:
```suggestion
            (gogoproto.moretags) = 'hash:"name:1" validate:"required,printascii"'
```

orchestrator/orchestrator.go
```diff
@@ -202,14 +202,14 @@ func (s *Orchestrator) outputToValue(output *process.Process_Node_Map_Output, wf
 			},
 		}, nil
 	case *process.Process_Node_Map_Output_Ref:
-		node, err := wf.FindNode(v.Ref.NodeKey)
-		if err != nil {
-			return nil, err
-		}
-		if node.GetTask() != nil {
-			return s.resolveInput(wf.Hash, exec, v.Ref.NodeKey, v.Ref.Key)
+		nodes := wf.FindNodes(func(n *process.Process_Node) bool {
```
NicolasMahe: I correct the related issue https://github.com/mesg-foundation/engine/issues/1557
so `refKey` is optional
if not set, it the ref should be done on the outputs of the previous node (`return data.Fields[v.Ref.Key]`)

NicolasMahe:
TODO:
- [ ] rename `execution.nodeKey ` to `execution.refKey`

orchestrator/orchestrator.go
krhubert: we don't have a `Key` in Ref. And we set up that the ref key is not optional (same with path) or I missunderstood something

orchestrator/orchestrator.go
```diff
@@ -201,34 +202,43 @@ func (s *Orchestrator) outputToValue(output *process.Process_Node_Map_Output, wf
...
-			return nil, err
+		if v.Ref.RefKey == "" {
+			return resolveRef(data, v.Ref.Path)
 		}
-		if node.GetTask() != nil {
-			return s.resolveInput(wf.Hash, exec, v.Ref.NodeKey, v.Ref.Path)
+		// the following code is only for catching checking the node exist.
+		nodes := wf.FindNodes(func(n *process.Process_Node) bool {
+			task := n.GetTask()
+			log.Println("getting task", n.Key)
```
antho1404: A few logs to remove or if useful to output with logrus

process/process.go
```diff
@@ -21,8 +21,8 @@ func (w *Process) Validate() error {
 		mapNode := node.GetMap()
 		if mapNode != nil {
 			for _, output := range mapNode.Outputs {
-				if ref := output.GetRef(); ref != nil {
-					if _, err := w.FindNode(ref.NodeKey); err != nil {
+				if ref := output.GetRef(); ref != nil && ref.RefKey != "" {
+					if _, err := w.FindNode(ref.RefKey); err != nil {
```
antho1404: I don't think this is right, `findNode` with check the `nodeKey`, in our case `nodeKey` is different than `refKey`.
This should be a find with a filter that the node is a task and has the same `refKey` or am I missing something here?

orchestrator/orchestrator.go
```diff
@@ -90,7 +91,7 @@ func (s *Orchestrator) dependencyFilter(exec *execution.Execution) func(wf *proc
 		if len(parents) > 1 {
 			return false, fmt.Errorf("multi parents not supported")
 		}
-		return parents[0] == exec.NodeKey, nil
+		return parents[0] == exec.RefKey, nil
```
antho1404: `parents` are the list of parent nodes, so the list of `nodeKey`, I don't think this is right

e2e/orchestrator_ref_grand_parent_task_test.go
```diff
@@ -33,6 +33,7 @@ func testOrchestratorRefTask(executionStream pb.Execution_StreamClient, instance
 						Key: "n1",
 						Type: &process.Process_Node_Task_{
 							Task: &process.Process_Node_Task{
+								RefKey:       "n1",
```
antho1404: Let's use a different name for `key` and `refKey` to avoid false positive in the tests

```suggestion
								RefKey:       "ref1",
```

NicolasMahe:
@antho1404 will finish the modif of this PR

process/process.go
```diff
@@ -22,9 +22,13 @@ func (w *Process) Validate() error {
...
 			for _, output := range mapNode.Outputs {
 				if ref := output.GetRef(); ref != nil && ref.RefKey != "" {
-					if _, err := w.FindNode(ref.RefKey); err != nil {
+					refNode, err := w.FindTaskNodeFromRef(node.Key, ref.RefKey)
+					if err != nil {
 						return err
 					}
+					if refNode == nil {
+						return fmt.Errorf("%q reference not found", ref.RefKey)
+					}
```
krhubert: Shouldn't this error be returned by `FindTaskNodeFromRef` ?

process/process.go
```diff
@@ -45,3 +49,25 @@ func (w *Process) Trigger() (*Process_Node, error) {
 	}
 	return triggers[0], nil
 }
+
+// FindTaskNodeFromRef returns the node associated to a reference in the nodeKey's parents, returns an nil if not found
+func (w *Process) FindTaskNodeFromRef(nodeKey string, refKey string) (*Process_Node, error) {
```
krhubert:
```suggestion
func (w *Process) FindTaskNodeFromRef(nodeKey, refKey string) (*Process_Node, error) {
```

Not mandatory

process/process.go
```diff
@@ -45,3 +49,25 @@ func (w *Process) Trigger() (*Process_Node, error) {
...
+	node, err := w.FindNode(nodeKey)
+	if err != nil {
+		return nil, err
+	}
+	if node.GetTask() != nil && node.GetTask().RefKey == refKey {
```
krhubert:
```suggestion
	if task := node.GetTask(); task != nil && task.RefKey == refKey {
```

process/process.go
```diff
@@ -45,3 +49,25 @@ func (w *Process) Trigger() (*Process_Node, error) {
...
+	}
+	var refNode *Process_Node
+	for _, parent := range w.ParentKeys(nodeKey) {
+		r, err := w.FindTaskNodeFromRef(parent, refKey)
+		if err != nil {
+			return nil, err
+		}
+		if refNode == nil && r != nil {
+			refNode = r
+		}
```
krhubert: shoudn't we `break` here on refNode found?

antho1404:
I think this PR adds too much complexity for no real benefits. I created another PR (https://github.com/mesg-foundation/engine/pull/1567) with a simple check on the parent and the ref that solves the data availability issue and still resolve based on the `nodeKey`.

This way we have a single system (nodeKey) to construct the graph and reference values but with additional checks.

There are more checks to do on the process that is listed on this issue https://github.com/mesg-foundation/engine/issues/1568

PR #1580, opened 2020-01-03
Engine execution stats

antho1404:
I needed to add some stats in order to debug https://github.com/mesg-foundation/engine/issues/1570.

This is not a final version, let's discuss first what and how we need to implements metrics but this is a way to start.

#### Start prometheus and grafana

```
docker service create -p 3001:3000 --network=engine grafana/grafana
docker service create -p 9090:9090 --network=engine --name=prometheus --mount type=bind,source=$(pwd)/prometheus.yml,destination=/etc/prometheus/prometheus.yml prom/prometheus
```

#### Configure grafana
- go to http://localhost:3001/
- add a datasource with the url `http://prometheus:9090`
- create a dashboard with 4 different metrics
  - mesg_execution_created
  - mesg_execution_signed
  - mesg_execution_in_progress
  - mesg_execution_completed

Then you can create an execution the way you want and you will see the 4 different steps of the lifecycle of an execution


prometheus.yml
```diff
@@ -0,0 +1,13 @@
+global:
```
NicolasMahe: is this file necessary here?
It's a config file for prometheus, not for the engine.

server/grpc/server.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	grpc_logrus "github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus"
+	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
```
NicolasMahe: nice

sdk/execution/metrics.go
```diff
@@ -0,0 +1,56 @@
+package executionsdk
+
+import (
+	"github.com/go-kit/kit/metrics"
+	prometheus "github.com/go-kit/kit/metrics/prometheus"
+	stdprometheus "github.com/prometheus/client_golang/prometheus"
+)
+
+var m *metric
```
NicolasMahe: How do this object is shared to tendermint's prometheus? Is it working with public singleton?

sdk/execution/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+
+type metric struct {
+	Created    metrics.Counter
```
NicolasMahe: We should prefix with the name of the method:
```suggestion
	CreateCreated    metrics.Counter
```

sdk/execution/metrics.go
antho1404: just register globally when initializing it with `prometheus.NewCounterFrom`

sdk/execution/metrics.go
antho1404: they are already prefixed with the namespace and subsystem so they will have a label like `mesg_execution_created`. These are metrics about the execution itself, not the function of the sdk.

Not so sure why you want to prefix

prometheus.yml
antho1404: Agree, this is not absolutely necessary but really good to have especially if we want in dev to run some analytics and maybe the dev script should start it too and in that case, it would be useful.

What do you think is the best?
- Keep it in order to have it in dev too
- Make it an example > `prometheus.example.yml`
- Move it to another config repo like https://github.com/mesg-foundation/testnet-ansible
- Add it to the documentation
- Just get rid of it

sdk/execution/metrics.go
NicolasMahe: It's only for Go style. These metrics are only used for the creation of an execution. How to name them when we add metric about the update of execution?
It's not really important as it's only a before compilation issue.

prometheus.yml
NicolasMahe: Move it to ansible repo ;) the `testnet-ansible` should also be renamed to just `ansible`.

sdk/execution/metrics.go
antho1404: Not these stats are for the whole lifecycle of the execution:
- Created
- PreSigned (not really useful, mostly for tests)
- Signed
- InProgress
- Completed

That's all the status of the execution from created to completed

prometheus.yml
antho1404: done

PR #1608, opened 2020-01-16
Change default path for the engine

antho1404:
This allows us to store in the same `.mesg` multiple networks. With the dev script we write on the dev chain but we could run the testnet still in the same `.mesg` directory


NicolasMahe:
Closing as we choose not to create a special path system for the network and just keep the original one.

PR #1609, opened 2020-01-17
Force encoding in json everywhere in the engine

antho1404:
By changing encoding to json we make tendermint compatible from any client.
With that the CLI or any UI can interact directly with tendermint without interacting with any engine's api


NicolasMahe:
@NicolasMahe check the cosmos module to see how they encode the ABCI messages

NicolasMahe:
> @NicolasMahe check the cosmos module to see how they encode the ABCI messages

Cosmos module are encoding in JSON in the queriers function. we should do the same.
But let's keep binary encoding for the stores.
issue created: https://github.com/mesg-foundation/engine/issues/new

PR #1626, opened 2020-01-31
Emit only one event with multiple attributes in cosmos module handler

NicolasMahe:
Fixes https://github.com/mesg-foundation/engine/issues/1624

New event output:
```json
[ { "key": "module", "value": "runner" }, { "key": "hash", "value": "9CxqDUrJaXj8Vea6kL5dMzyn7sD6nx6htePkbLitNxfq" } ]
```

Screen Shot 2020-01-31 at 15 11 08

----

Should we also create a new type of event. Currently we are using "message" that is define by cosmos.

### DIFF
--- a/cosmos/module.go
+++ b/cosmos/module.go
@@ -114,17 +114,17 @@ func (m *AppModule) NewHandler() cosmostypes.Handler {
 			return cosmostypes.ErrInternal(err.Error()).Result()
 		}
 
-		request.EventManager().EmitEvent(cosmostypes.NewEvent(
+		event := cosmostypes.NewEvent(
 			cosmostypes.EventTypeMessage,
 			cosmostypes.NewAttribute(cosmostypes.AttributeKeyModule, m.name),
-		))
-
+		)
 		if hash != nil {
-			request.EventManager().EmitEvent(cosmostypes.NewEvent(
-				cosmostypes.EventTypeMessage,
+			event = event.AppendAttributes(
 				cosmostypes.NewAttribute(AttributeKeyHash, hash.String()),
-			))
+			)
 		}
+		request.EventManager().EmitEvent(event)
+
 		return cosmostypes.Result{
 			Data:   hash,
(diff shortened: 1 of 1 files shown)


NicolasMahe:
> As I remember engine has a query for recaive this event somewhere. Is it still valid?

yes it's still working

antho1404:
I think we should have events based on the module name instead of message (that is the name used by cosmos to have an event based on the messages processed) so if we want specific events we should make them dedicated for what we need.

```
execution: [{ key: "hash", value: "xxxx" }, { key: "action", value: "create" }]
```

This way we will be able to query all resources that has been created/updated and we can always query multiple resources with `AND` or `OR` in the query. https://docs.tendermint.com/master/rpc/#/Websocket/subscribe

To avoid conflicts we could prefix the events with mesg so `mesg:execution`, `mesg:service`.

Another solution is to also embed the action in the event name `mesg:execution:create`. We would still be able to query all execution related events with the `CONTAINS` instruction but might be harder.

NicolasMahe:
> I think we should have events based on the module name instead of message (that is the name used by cosmos to have an event based on the messages processed) so if we want specific events we should make them dedicated for what we need.
> 
> ```
> execution: [{ key: "hash", value: "xxxx" }, { key: "action", value: "create" }]
> ```
> 
> This way we will be able to query all resources that has been created/updated and we can always query multiple resources with `AND` or `OR` in the query. https://docs.tendermint.com/master/rpc/#/Websocket/subscribe
> 
> To avoid conflicts we could prefix the events with mesg so `mesg:execution`, `mesg:service`.
> 
> Another solution is to also embed the action in the event name `mesg:execution:create`. We would still be able to query all execution related events with the `CONTAINS` instruction but might be harder.

let's continue the discussion when the codebase is updated to cosmos v0.38 and we refactor the sdk to cosmos modules. See https://github.com/mesg-foundation/engine/issues/1639

PR #1627, opened 2020-02-02
Simplify txbuilder and improve keybase signing

NicolasMahe:
This PR remove the custom TxBuilder that we have in favour of using the default one directly.
The custom keybase signing has been re-implemented to add a cache on the plain private key of the user. This impressively improves the performance! The engine can sign 1,000 of txs in less than 1 sec (before only around 3 or 4 per sec)

Related to https://github.com/mesg-foundation/engine/issues/1570

Next step is to investigate mempool issue by running the monitoring tool


NicolasMahe:
> I don't really like the management of the cache for the private key.
> 
> * The delay might be unnecessary for now
> * The management of the `lastSign` is not clear. We should support have a map to cache all of them or none of them.

I remove the cache based on time to a cache using a map to store all private keys used.

---

Performance comparison with the e2e tests `execution/many_executions_in_parallel` with a block time of 1sec that creates 10 executions:
before: 13.31s
after: 2.10s

🍾

cosmos/keybase.go
```diff
@@ -81,10 +85,25 @@ func (kb *Keybase) Delete(name, passphrase string, skipPass bool) error {
 }
 
 // Sign is a lock protected version of keys.Sign
+// it also keeps the last private key used in memory for the time set in `keepPrivTime` for performance improvement.
 func (kb *Keybase) Sign(name, passphrase string, msg []byte) ([]byte, crypto.PubKey, error) {
 	kb.mx.Lock()
 	defer kb.mx.Unlock()
-	return kb.kb.Sign(name, passphrase, msg)
+	hash := sha256.Sum256([]byte(name + passphrase))
+	priv := kb.privCache[hash]
```
antho1404: Can we just check the presence of the key instead of the fetched value

```suggestion
	priv, ok := kb.privCache[hash]
	if ok {
```

cosmos/keybase.go
```diff
@@ -16,9 +17,9 @@ const mnemonicEntropySize = 256
 
 // Keybase is a standard cosmos keybase.
 type Keybase struct {
-	kb keys.Keybase
-
-	mx sync.Mutex
+	kb        keys.Keybase
+	mx        sync.Mutex
+	privCache map[[sha256.Size]byte]crypto.PrivKey
```
krhubert:
```suggestion
	privKeyCache map[[sha256.Size]byte]crypto.PrivKey
```

or just `privKeys`, but privCache expands to private cache 😕

cosmos/keybase.go
```diff
@@ -81,10 +85,25 @@ func (kb *Keybase) Delete(name, passphrase string, skipPass bool) error {
 }
 
 // Sign is a lock protected version of keys.Sign
+// it also keeps the last private key used in memory for the time set in `keepPrivTime` for performance improvement.
 func (kb *Keybase) Sign(name, passphrase string, msg []byte) ([]byte, crypto.PubKey, error) {
 	kb.mx.Lock()
 	defer kb.mx.Unlock()
-	return kb.kb.Sign(name, passphrase, msg)
+	hash := sha256.Sum256([]byte(name + passphrase))
```
krhubert: We need to find different way for calc hash. Imagine you have two users:

- bob:pass
- bobp:ass

They both will have the same hash which can lead to unexpected behaviour.

I think adding join string will solve the problem

```suggestion
	hash := sha256.Sum256([]byte(name + ":" + passphrase))
```

cosmos/keybase_test.go
```diff
@@ -40,4 +41,47 @@ func TestKeybase(t *testing.T) {
...
+				}
+				sig, pub, err := kb.Sign(nameToUse, password, []byte{})
+				require.NoError(t, err)
+				require.NotEmpty(t, sig)
+				require.NotEmpty(t, pub)
+			}
+			require.NotEmpty(t, kb.privCache[hash])
+			require.NotEmpty(t, kb.privCache[hash2])
+			require.NotEqual(t, kb.privCache[hash], kb.privCache[hash2])
+		})
```
krhubert:
```go
    t.Run("Many times", func(t *testing.T) {
      hash := sha256.Sum256([]byte(name + password))
      hash2 := sha256.Sum256([]byte(name2 + password))
      for i := 0; i < 1000; i++ {
        nameToUse := name
        if i%2 == 0 {
          nameToUse = name2
        }
        _, _, err := kb.Sign(nameToUse, password, []byte{})
        require.NoError(t, err)
      }
      require.NotEqual(t, kb.privCache[hash], kb.privCache[hash2])
    })
```

Checks duplication, can be simplified a bit. Also you might take a look on `testing/quick` package, as it might introduce some randomness into such tests.

cosmos/keybase.go
```diff
@@ -81,10 +85,25 @@ func (kb *Keybase) Delete(name, passphrase string, skipPass bool) error {
...
+			return nil, nil, err
+		}
+		kb.privCache[hash] = priv
+	}
+	sig, err := priv.Sign(msg)
+	if err != nil {
+		return nil, nil, err
+	}
+	pub := priv.PubKey()
+	return sig, pub, nil
```
krhubert:
```suggestion
	return sig, priv.PubKey(), nil
```

cosmos/client.go
```diff
@@ -232,9 +244,20 @@ func (c *Client) sign(msg sdktypes.Msg) (tenderminttypes.Tx, error) {
...
-	signedTx, err := txBuilder.BuildAndSignStdTx(msg, c.accName, c.accPassword)
+	// create StdSignMsg
+	stdSignMsg, err := txBuilder.BuildSignMsg(msgs)
 	if err != nil {
 		return nil, err
 	}
-	return txBuilder.Encode(signedTx)
+
+	// create StdTx
+	stdTx := authtypes.NewStdTx(stdSignMsg.Msgs, stdSignMsg.Fee, []authtypes.StdSignature{}, stdSignMsg.Memo)
```
krhubert:
```suggestion
	stdTx := authtypes.NewStdTx(stdSignMsg.Msgs, stdSignMsg.Fee, nil, stdSignMsg.Memo)
```

if `false` is passed to SignStdTx method, then we don't need an empty sigs here.

cosmos/keybase_test.go
```diff
@@ -40,4 +41,47 @@ func TestKeybase(t *testing.T) {
...
+			require.NoError(t, err)
+			require.NotEmpty(t, sig2)
+			require.NotEmpty(t, pub2)
+			require.NotEmpty(t, kb.privCache[hash2])
+			// not the same
+			require.NotEqual(t, sig, sig2)
+			require.NotEqual(t, pub, pub2)
+			require.NotEqual(t, kb.privCache[hash], kb.privCache[hash2])
+		})
+		t.Run("Many times", func(t *testing.T) {
```
krhubert: What is the purpose for tests "Two accounts" and "Many times", are they checking different thing? can't we just have one of them?

cosmos/keybase.go
```diff
@@ -81,10 +85,25 @@ func (kb *Keybase) Delete(name, passphrase string, skipPass bool) error {
 }
 
 // Sign is a lock protected version of keys.Sign
+// it also keeps the last private key used in memory for the time set in `keepPrivTime` for performance improvement.
 func (kb *Keybase) Sign(name, passphrase string, msg []byte) ([]byte, crypto.PubKey, error) {
 	kb.mx.Lock()
```
krhubert: Shoudn't mutex only surround `kb.kb.ExportPrivateKeyObject`, should it?

I guess engine losses a lot of time while waiting for Sing  the transaction (and a bit for calc the hash ;)). It dosen't need a mutex there.

NicolasMahe: signing is really super fast. but yeah i will make the change

cosmos/keybase_test.go
NicolasMahe: i don't see how i could use `testing/quick`. do you have suggestion?

NicolasMahe:
Here is the result before and after with the monitoring metrics:
before:
Screen Shot 2020-02-04 at 14 50 18
everything is stuck and i had to stop the process

after:
Screen Shot 2020-02-04 at 15 47 54
no issue at all. ALL erc20 events are saved to influxdb!!
45,000 txs executed in 45min. and that's not even the maximum!

@antho1404 check again now we have the dashboard in place!

PR #1637, opened 2020-02-04
Always start monitoring with dev

NicolasMahe:
- Always start monitoring with `make dev`
- Remove `make dev-mon`

Dependant on https://github.com/mesg-foundation/engine/pull/1636


Makefile
```diff
@@ -37,9 +37,6 @@ docker-tools:
 	docker build -t mesg/tools:local -f Dockerfile.tools .
 
 dev: docker-dev
-	- ./scripts/dev.sh
```
krhubert: 👎 

I don't like this change. It takes ~20 seconds to start monitoring and it is not needed in many cases. I prefer to have seperate target for monitoring.

NicolasMahe: i agree it's very boring to have to wait 20 seconds more to run the engine but the argument here is when we will have an issue we always have the monitoring tool in place to be able to check it. but at the same time, if the dev command delete the monitoring tools, then it's useless..

what about a version that start it only if it doesn't already run, but it doesn't delete it on stop?

PR #1640, opened 2020-02-04
Update to cosmos v0.38

NicolasMahe:
Update to cosmos v0.38 🎊 

but we had to set an hack (https://github.com/mesg-foundation/engine/pull/1640/commits/fae99d69bc69faa794819856f6aea28c95e9c724) to force the sequence to increment even in when tx is being checked

closes https://github.com/mesg-foundation/engine/issues/1612


krhubert:
Needs test a bit more, but it will be done tommorow. Otherwise looking good.

cosmos/errors.go
```diff
@@ -1,26 +1,37 @@
...
+var (
+	ErrValidation = sdkerrors.Register(CodespaceMesg, CodeValidation, "validation failed")
+	ErrMarshal    = sdkerrors.Register(CodespaceMesg, CodeMarshal, "failed to marshal")     // TODO: to replace by cosmoserrors.ErrJSONMarshal if it makes sense
+	ErrUnmarshal  = sdkerrors.Register(CodespaceMesg, CodeUnmarshal, "failed to unmarshal") // TODO: to replace by cosmoserrors.ErrJSONUnmarshal if it makes sense
 )
 
 // NewMesgErrorf creates error with given code type and mesg codespace.
-func NewMesgErrorf(ct types.CodeType, format string, a ...interface{}) types.Error {
-	return types.NewError(CodespaceMesg, ct, format, a...)
+func NewMesgErrorf(code uint32, format string, a ...interface{}) *sdkerrors.Error {
```
NicolasMahe: it looks like this function is not used anymore. can be deleted.

cosmos/errors.go
```diff
@@ -1,26 +1,37 @@
...
+	return sdkerrors.New(CodespaceMesg, code, fmt.Sprintf(format, a...))
 }
 
 // NewMesgWrapError creates error with given code type and mesg codespace.
-func NewMesgWrapError(ct types.CodeType, err error) types.Error {
-	return types.NewError(CodespaceMesg, ct, err.Error())
+func NewMesgWrapError(code uint32, err error) *sdkerrors.Error {
```
NicolasMahe: same here. can be deleted

krhubert: This one is used.

PR #1649, opened 2020-02-09
Remove/move all x packages

antho1404:
In order to prepare for the new cosmos modules in the `x` directory, I moved/deleted all the `x` modules.
The only duplication from these changes is the `x-signal` that is duplicated in the main and the e2e service, everything is else is just reorganized with no duplication.

- `x-string`: code placed where needed
- `x-signal`: code duplicated on main and e2e service
- `x-reflect`: code moved in the cosmos client
- `x-rand`: code moved in the main
- `x-os`: create a new package `env`
- `x-validator`: create a new validator package
- `x-net`: domain name check moved to validator package and port check on the main
- `x-error`: create a new `errors` package that we can use with `errors.New`

With this, we have nothing left on the `x` directory and can be fully used for cosmos modules.


krhubert:
I'm very against that change. What you did is just negation why the `/x` package was introduce in the first place. To keep all those functions and tests in a sepearate packages (easy to test, maintain and develop)

Just to be clear, I've started to put cosmos modules under `/module` directory. 
if we want to change `/x` we can just rename it, but moving all functions around code base + removing tests (!!!) only makes engine to step backward.

As I see it, we can rename `/x` to `/ext` or keep cosmos modules under `module` dir which is more descriptive.

NicolasMahe:
Let's not merge this PR for now. We are about to move a lot of code, so let's come back to this PR once its done.
Let's do a new PR that just move `x` packages to `ext` for now.

NicolasMahe:
closing because too old but referenced in https://github.com/mesg-foundation/engine/issues/1639

Issue #62, opened 2018-03-23
Possibility to backup the address with an export system


antho1404:
#26 related

PR #1651, opened 2020-02-10
Use template to move ownership cosmos module to x/ dir

krhubert:
Unfortunately, I stuck on the problem with process deletion and I can't track it :/

@ NicolasMahe can you take a look (e2e tests)? 

related to https://github.com/mesg-foundation/engine/issues/1639


app/app.go
```diff
@@ -84,6 +84,7 @@ func MakeCodec() *codec.Codec {
 
 	ModuleBasics.RegisterCodec(mesgcodec.Codec)
 	vesting.RegisterCodec(mesgcodec.Codec)
+	ownership.RegisterCodec(mesgcodec.Codec)
```
NicolasMahe: this should not be done here.
the call to `ModuleBasics.RegisterCodec(mesgcodec.Codec)` should iterate over all the module and register the codec against each code. The module has to itself on the codec by implementing [func (AppModuleBasic) RegisterCodec(cdc *codec.Codec) {](https://github.com/mesg-foundation/engine/pull/1651/files#diff-a1fb0f761753d7736d7a16cbeed04b6aR34) but you already did it.

app/app.go
```diff
@@ -59,7 +59,7 @@ var (
 		supply.AppModuleBasic{},
 
 		// Engine's AppModuleBasic
-		cosmos.NewAppModuleBasic(ownershipsdk.ModuleName),
+		cosmos.NewAppModuleBasic(ownership.ModuleName),
```
NicolasMahe: We don't want to use any custom code like we previously did. Don't call `cosmos.NewAppModuleBasic` as it uses the moduleBasic define in `cosmos/module.go` and don't implement much stuff, as opposed as your new version in `x/ownership/module.go`
```suggestion
		ownership.AppModuleBasic{},
```

app/app.go
NicolasMahe: In your currently implementation, this is required because your don't register the right ownership AppModuleBasic. See comment line #62 of this same file

app/app.go
```diff
@@ -242,7 +243,7 @@ func NewInitApp(
 	)
 
 	// Engine's module keepers
-	app.ownershipKeeper = ownershipsdk.NewKeeper(keys[ownershipsdk.ModuleName])
+	app.ownershipKeeper = ownership.NewKeeper(app.cdc, keys[ownership.ModuleName])
```
NicolasMahe: the ownership module should declare its own `StoreKey` like other cosmos module. Eg:
```suggestion
	app.ownershipKeeper = ownership.NewKeeper(app.cdc, keys[ownership.StoreKey])
```

sdk/process/keeper.go
```diff
@@ -6,21 +6,21 @@ import (
 	cosmostypes "github.com/cosmos/cosmos-sdk/types"
 	"github.com/mesg-foundation/engine/codec"
 	"github.com/mesg-foundation/engine/hash"
-	"github.com/mesg-foundation/engine/ownership"
+	ownershippb "github.com/mesg-foundation/engine/ownership"
```
NicolasMahe: As said in https://github.com/mesg-foundation/engine/issues/1639#issuecomment-584042783, we can also move the proto go struct directly in `x/module/ownership` package or sub-package:
>  execution, instance, ownership, process, runner, service packages should move to x/ packages. All of the code is about data definition or validation.

sdk/process/keeper.go
```diff
@@ -6,21 +6,21 @@ import (
...
 	instancesdk "github.com/mesg-foundation/engine/sdk/instance"
-	ownershipsdk "github.com/mesg-foundation/engine/sdk/ownership"
+	"github.com/mesg-foundation/engine/x/ownership"
 )
 
 // Keeper is the service keeper.
 type Keeper struct {
 	storeKey        *cosmostypes.KVStoreKey
-	ownershipKeeper *ownershipsdk.Keeper
+	ownershipKeeper ownership.Keeper
```
NicolasMahe: If we really want to do it nicely, the process module should declare an `OwershipKeeper` interface. But i'm ok to keep it like this for now.

x/ownership/handler.go
```diff
@@ -0,0 +1,17 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/mesg-foundation/engine/x/ownership/internal/types"
+)
+
+// NewHandler creates an sdk.Handler for all the instance type messages
+func NewHandler(k Keeper) sdk.Handler {
+	return func(ctx sdk.Context, msg sdk.Msg) (*sdk.Result, error) {
+		errMsg := fmt.Sprintf("unrecognized %s message type: %T", types.ModuleName, msg)
```
NicolasMahe: The module should also use its alias.
The alias system is made to allow customisation of the module. So the main package `x/ownership` should use them, i think.
```suggestion
		errMsg := fmt.Sprintf("unrecognized %s message type: %T", ModuleName, msg)
```

x/ownership/client/cli/query.go
```diff
@@ -0,0 +1,51 @@
...
+// GetCmdListOwnerships returns command listing ownerships
+func GetCmdListOwnerships(queryRoute string, cdc *codec.Codec) *cobra.Command {
+	return &cobra.Command{
+		Use:   "list",
+		Short: "list",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cliCtx := context.NewCLIContext().WithCodec(cdc)
+			res, _, err := cliCtx.QueryWithData(fmt.Sprintf("custom/%s/"+types.QueryListOwnerships, queryRoute), nil)
+			if err != nil {
+				fmt.Printf("could not get scavenges\n%s\n", err.Error())
```
NicolasMahe:
```suggestion
				fmt.Printf("could not get ownerships\n%s\n", err.Error())
```

x/ownership/client/rest/query.go
```diff
@@ -0,0 +1,39 @@
+package rest
```
NicolasMahe: just a reminder: this implementation is not done.

x/ownership/internal/keeper/keeper.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	ownership := &ownership.Ownership{
+		Owner:        owner.String(),
+		Resource:     resource,
+		ResourceHash: resourceHash,
+	}
+	ownership.Hash = hash.Dump(ownership)
+
+	store.Set(ownership.Hash, k.cdc.MustMarshalBinaryLengthPrefixed(ownership))
```
NicolasMahe: Let's avoid to use `MustXXX` functions and return the error.

x/ownership/internal/types/key.go
```diff
@@ -0,0 +1,15 @@
...
+	ModuleName = "ownership"
+
+	// StoreKey to be used when creating the KVStore
+	StoreKey = ModuleName
+
+	// RouterKey to be used for routing msgs
+	RouterKey = ModuleName
+
+	// QuerierRoute to be used for routing
+	QuerierRoute = ModuleName
```
NicolasMahe: let's use those const instead of ModuleName when it make sense.

server/grpc/api/ownership.go
```diff
@@ -3,26 +3,32 @@ package api
...
+		client: client,
+	}
 }
 
 // List returns all ownerships.
 func (s *OwnershipServer) List(ctx context.Context, req *protobuf_api.ListOwnershipRequest) (*protobuf_api.ListOwnershipResponse, error) {
-	ownerships, err := s.sdk.Ownership.List()
-	if err != nil {
+	var ownerships []*ownershippb.Ownership
+	if err := s.client.QueryJSON("custom/"+ownership.ModuleName+"/"+ownership.QueryListOwnerships, nil, &ownerships); err != nil {
```
NicolasMahe: As said in other comment, we should use specific constant as much as possible. Here, replace `ModuleName` by `QuerierRoute`.
```suggestion
	if err := s.client.QueryJSON("custom/"+ownership.QuerierRoute +"/"+ownership.QueryListOwnerships, nil, &ownerships); err != nil {
```

x/ownership/internal/keeper/keeper.go
```diff
@@ -0,0 +1,104 @@
...
+
+// hasOwner checks if given resource has owner. Returns all ownership hash and true if has one
+// nil and false otherwise.
+func (k Keeper) findOwnerships(store sdk.KVStore, owner string, resourceHash hash.Hash) []hash.Hash {
+	var ownerships []hash.Hash
+	iter := store.Iterator(nil, nil)
+
+	for iter.Valid() {
+		var o *ownership.Ownership
+		if err := k.cdc.UnmarshalBinaryBare(iter.Value(), &o); err == nil {
```
NicolasMahe: The e2e issue is coming from here. Wrong `Unmarshal` function used:
```suggestion
		if err := k.cdc.UnmarshalBinaryLengthPrefixed(iter.Value(), &o); err == nil {
```

NicolasMahe:
> Unfortunately, I stuck on the problem with process deletion and I can't track it :/
> 
> @ NicolasMahe can you take a look (e2e tests)?

Found the issue. See https://github.com/mesg-foundation/engine/pull/1651/files#r377441190

cosmos/client.go
```diff
@@ -55,6 +55,24 @@ func NewClient(node *node.Node, kb keys.Keybase, chainID, accName, accPassword,
 	}, nil
 }
 
+// QueryJSON is abci.query wrapper with errors check and decode data.
+func (c *Client) QueryJSON(path string, qdata, ptr interface{}) error {
```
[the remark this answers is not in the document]
krhubert: Exacly!

sdk/process/keeper.go
krhubert: I didn't want to put that in this PR, but I can if you want to.

sdk/process/keeper.go
krhubert: Yes, that's the standard way for the framework and it will be done when process/service module will be transform to it.

x/ownership/handler.go
krhubert: It's taken from the template. I think the alias is for other packages, because the module can use `internal/` safely.

x/ownership/client/rest/query.go
krhubert: What do you mean? It is done. Here only parameters are exported, but the `list/` route is already exported as well. Open & write what missng in your opinion.

x/ownership/internal/keeper/keeper.go
krhubert: This one is from the template. I think it's good to have `Must` as this will only panic if the concreate type is not register (same with json).

x/ownership/internal/types/key.go
krhubert: Ok. Given resolve as you already proposed changes in the client.

x/ownership/client/rest/query.go
NicolasMahe: where is the route query exported?
don't confuse handler querier (backend) and rest query (client)

x/ownership/internal/keeper/keeper.go
NicolasMahe: but this function already return an error.. why not using it?

x/ownership/client/rest/query.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/cosmos/cosmos-sdk/client/context"
+	"github.com/cosmos/cosmos-sdk/types/rest"
+	"github.com/gorilla/mux"
+	"github.com/mesg-foundation/engine/x/ownership/internal/types"
+)
+
+func registerQueryRoutes(cliCtx context.CLIContext, r *mux.Router) {
+	r.HandleFunc(
+		"/ownership/list",
+		queryParamsHandlerFn(cliCtx),
```
NicolasMahe:
```suggestion
		queryListHandlerFn(cliCtx),
```

x/ownership/client/rest/query.go
```diff
@@ -0,0 +1,62 @@
...
+	).Methods(http.MethodGet)
+	r.HandleFunc(
+		"/ownership/parameters",
+		queryListHandlerFn(cliCtx),
```
NicolasMahe:
```suggestion
		queryParamsHandlerFn(cliCtx),
```

x/ownership/client/cli/query.go
```diff
@@ -0,0 +1,51 @@
...
+	ownershipQueryCmd := &cobra.Command{
+		Use:                        types.ModuleName,
+		Short:                      fmt.Sprintf("Querying commands for the %s module", types.ModuleName),
+		DisableFlagParsing:         true,
+		SuggestionsMinimumDistance: 2,
+		RunE:                       client.ValidateCmd,
+	}
+
+	ownershipQueryCmd.AddCommand(
+		flags.GetCommands()...,
```
NicolasMahe: You need to add the `GetCmdListOwnerships` to this list

x/ownership/client/rest/query.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func queryListHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		cliCtx, ok := rest.ParseQueryHeightOrReturnBadRequest(w, cliCtx, r)
+		if !ok {
+			return
+		}
+
+		route := fmt.Sprintf("custom/%s/list", types.QuerierRoute)
```
NicolasMahe:
```suggestion
		route := fmt.Sprintf("custom/%s/%s", types.QuerierRoute, types.QueryListOwnerships)
```

PR #1670, opened 2020-02-18
Implement first iteration of the executions ecomomy

krhubert:
TODO:

I need to check why account has insufficient founds.

```
ERRO[0015] finished unary call with code Unknown         error="an error occurred in transaction: insufficient funds: insufficient account funds;  < 1atto: failed to execute message; message index: 0" grpc.code=Unknown grpc.method=Update grpc.service=mesg.api.Execution grpc.start_time="2020-02-18T19:38:25Z" grpc.time_ms=1020.083 module=grpc span.kind=server system=grpc
```

- I don't like append in the execution store, but I think we don't have any better option and we have to store the execution price.
- split price accros all service owners.


x/execution/internal/keeper/keeper.go
```diff
@@ -86,10 +92,21 @@ func (k *Keeper) Create(ctx sdk.Context, msg types.MsgCreateExecution) (*executi
...
 	if !ctx.IsCheckTx() {
 		M.InProgress.Add(1)
 	}
+
+	if err := k.supplyKeeper.DelegateCoinsFromAccountToModule(ctx, msg.Signer, types.ModuleName, msg.Price); err != nil {
+		return nil, err
+	}
+
 	store.Set(exec.Hash, value)
+	store.Set(append(exec.Hash, []byte("_price")...), priceBz)
```
NicolasMahe: what is this?
do we need to keep the price here? can't we just add it to the execution struct (`exec.Price`)?

NicolasMahe:
> I need to check why account has insufficient founds.
> 
> ```
> ERRO[0015] finished unary call with code Unknown         error="an error occurred in transaction: insufficient funds: insufficient account funds;  < 1atto: failed to execute message; message index: 0" grpc.code=Unknown grpc.method=Update grpc.service=mesg.api.Execution grpc.start_time="2020-02-18T19:38:25Z" grpc.time_ms=1020.083 module=grpc span.kind=server system=grpc
> ```

what function's call is returning this error?
maybe it's a rounding issue. Coins don't support decimal and round the amount when using div.

x/execution/internal/keeper/keeper.go
krhubert: we can. This it the second option I wanted to discuss.

NicolasMahe: let's add the price to the execution itself. like this it will be available in the ABCI and thus easy for client to display it ;)

x/execution/internal/keeper/keeper.go
```diff
@@ -170,3 +249,60 @@ func (k *Keeper) List(ctx sdk.Context) ([]*executionpb.Execution, error) {
...
+			return nil, err
+		}
+		out = append(out, addr)
+	}
+	return out, nil
+}
+
+// coinsQuoRem divides coins and returns quotient and remainder.
+// NOTE: if reminder is zero then it will be set to quotient.
+func coinsQuoRem(coins sdk.Coins, div int) (sdk.Coins, sdk.Coins) {
```
NicolasMahe: can't we use directly the arithmetic function on the type `coins` or `coin`?

e2e/execution_test.go
```diff
@@ -189,6 +189,35 @@ func testExecution(t *testing.T) {
...
+			Inputs:       inputs,
+		})
+		require.NoError(t, err)
+
+		_, err = streamInProgress.Recv()
+		require.NoError(t, err)
+
+		exec, err := streamCompleted.Recv()
+		require.NoError(t, err)
+		require.Equal(t, resp.Hash, exec.Hash)
```
NicolasMahe: we should check the balances of the accounts ;)

protobuf/api/execution.proto
```diff
@@ -70,6 +70,11 @@ message CreateExecutionRequest {
     (gogoproto.customtype) = "github.com/mesg-foundation/engine/hash.Hash",
     (gogoproto.nullable) = false
   ];
+
+  // price of running the execution.
+  string price = 10 [
```
antho1404: Can't we use the type `Coin` from cosmos here instead of having to parse it, this will be much easier to interact with if we have the same struct for all the messages/resources

app/app.go
```diff
@@ -385,6 +386,11 @@ func (app *NewApp) ModuleAccountAddrs() map[string]bool {
 	return modAccAddrs
 }
 
+// BankKeeper returns a keeper.
+func (app *NewApp) BankKeeper() bank.Keeper {
```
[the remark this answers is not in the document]
krhubert: Remaings, to remove.

x/execution/internal/keeper/keeper.go
```diff
@@ -3,36 +3,47 @@ package keeper
...
 	"github.com/mesg-foundation/engine/hash"
+	ownershippb "github.com/mesg-foundation/engine/ownership"
 	"github.com/mesg-foundation/engine/protobuf/api"
 	typespb "github.com/mesg-foundation/engine/protobuf/types"
 	"github.com/mesg-foundation/engine/x/execution/internal/types"
 	"github.com/tendermint/tendermint/libs/log"
 )
 
+// price share for the execution runner
+const runnerShare = 0.9
```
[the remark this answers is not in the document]
krhubert: It can be done even now, if we want that. This change is only about adding one field to service type.

protobuf/api/execution.proto
krhubert: I thought about that, but I haven't seen Unmarshal for type `Coins`. I can test it today.

e2e/execution_test.go
krhubert: Yes, the good e2e tests are missing... It's not only balance, but also service with multiple owners.

x/execution/internal/keeper/keeper.go
krhubert: I spot `DecCoins` yesterday as well, but I know the api of `math/big` and It was easier for me to write, because DecCoins lacks docs... 😠 But I can change it to deccoins today.

krhubert:
@antho1404 
>I have one issue with that, the money is transferred directly to the owner of the service and not to the service directly. ...

I thought we were talking about moving the money to the service owner as a first draft. transfer it to service is not that simple I guess.

1) the service doesn't have any wallet/address (except the owners one) - so how we want to create a service account
2) how we want to transfer money from service back to the user - we don't have any api to do so (including api for (automatic distribution, delegating payment to someone else, valuating the service based on its balance...)) ...

To be clear we can and probably should do that, but we need to talk about how we approach that.

x/execution/internal/keeper/keeper.go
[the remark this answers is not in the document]
antho1404: I like to have it hardcoded for now, don't want to complexify it now ;)

antho1404:
I had a quick look at this supply keeper and it seems that there is functions to generate accounts per module. Actually the functions you use are creating such accounts and it doesn't seem that there is any restriction in the module name so just having the hash of the service instead of the module name would do the job I think.
That might be enough but ultimately, I think it might be better to use a `Coin` attribute on each service, this would probably simplify a lot of things

antho1404:
For the part to transfer back to the user, I'm totally fine to have the token on the service only and that there is no way to get it back yet but at least we can see that the balance increase. Users have the ownership of these token so indirectly it's theirs and we can show the amount of token that every service received which might be hard if we distribute to the user (the user can receive from multiple services or direct transfers)

x/execution/internal/keeper/keeper.go
NicolasMahe: Please try with de coins ;) the go doc is your api :p

protobuf/api/execution.proto
krhubert: I'm waiting for the status PR status https://github.com/cosmos/cosmos-sdk/issues/5685. Without it we have drop support for equal_all in executions.proto.

NicolasMahe:
@krhubert please check https://github.com/mesg-foundation/engine/pull/1676

NicolasMahe:
let's add e2e test that check the balance of the execution, service, runner and user before merging!

app/app.go
```diff
@@ -71,6 +71,7 @@ var (
 	maccPerms = map[string][]string{
 		auth.FeeCollectorName:     nil,
 		distr.ModuleName:          nil,
+		execution.ModuleName:      {supply.Staking},
```
NicolasMahe: this is useless now

protobuf/api/execution.proto
antho1404: Looks like the support of `equal_all` has been added 🎉 https://github.com/cosmos/cosmos-sdk/pull/5688
For coin but also all the other cosmos types.

antho1404: Can we at least use their proto types instead of a string?

app/app.go
```diff
@@ -247,7 +248,7 @@ func NewInitApp(
 	app.processKeeper = process.NewKeeper(app.cdc, keys[process.StoreKey], app.instanceKeeper, app.ownershipKeeper)
 	app.serviceKeeper = service.NewKeeper(app.cdc, keys[service.StoreKey], app.ownershipKeeper)
 	app.runnerKeeper = runner.NewKeeper(app.cdc, keys[runner.StoreKey], app.instanceKeeper)
-	app.executionKeeper = execution.NewKeeper(app.cdc, keys[execution.StoreKey], app.serviceKeeper, app.instanceKeeper, app.runnerKeeper, app.processKeeper)
+	app.executionKeeper = execution.NewKeeper(app.cdc, keys[execution.StoreKey], app.bankKeeper, app.serviceKeeper, app.instanceKeeper, app.runnerKeeper, app.processKeeper, app.ownershipKeeper)
```
antho1404: I think `ownershipKeeper` is not useful anymore

x/execution/internal/keeper/keeper.go
```diff
@@ -6,35 +6,45 @@ import (
...
-	processKeeper  types.ProcessKeeper
+	storeKey sdk.StoreKey
+	cdc      *codec.Codec
+
+	bankKeeper      types.BankKeeper
+	serviceKeeper   types.ServiceKeeper
+	instanceKeeper  types.InstanceKeeper
+	runnerKeeper    types.RunnerKeeper
+	processKeeper   types.ProcessKeeper
+	ownershipKeeper types.OwnershipKeeper
```
antho1404: to delete

x/execution/internal/types/expected_keepers.go
```diff
@@ -37,3 +45,8 @@ type RunnerKeeper interface {
 type ProcessKeeper interface {
 	Get(ctx sdk.Context, hash hash.Hash) (*processpb.Process, error)
 }
+
+// OwnershipKeeper module interface.
+type OwnershipKeeper interface {
```
antho1404: to remove

x/ownership/internal/keeper/keeper.go
```diff
@@ -31,6 +31,25 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 	return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
 }
 
+// GetOwners returns the ownership from the keeper.
+func (k Keeper) GetOwners(ctx sdk.Context, resourceHash hash.Hash) ([]*ownership.Ownership, error) {
```
antho1404: to remove

x/execution/internal/types/msg.go
```diff
@@ -34,8 +34,15 @@ func (msg MsgCreateExecution) Type() string {
 
 // ValidateBasic runs stateless checks on the message.
 func (msg MsgCreateExecution) ValidateBasic() error {
+	price, err := sdk.ParseCoins(msg.Request.Price)
```
antho1404: not needed if we use the coin type

x/execution/internal/keeper/keeper.go
```diff
@@ -123,13 +145,59 @@ func (k *Keeper) Update(ctx sdk.Context, msg types.MsgUpdateExecution) (*executi
...
+	if err := k.distributePriceShares(ctx, execAddress, runnerAddress, serviceAddress, exec.Price); err != nil {
+		return nil, err
+	}
+
 	store.Set(exec.Hash, value)
 	return exec, nil
 }
 
+func (k *Keeper) distributePriceShares(ctx sdk.Context, execAddress, runnerAddress, serviceAddress sdk.AccAddress, price string) error {
+	coins, err := sdk.ParseCoins(price)
```
antho1404: not needed if we use the coin type

x/execution/internal/keeper/keeper.go
```diff
@@ -44,6 +54,10 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 
 // Create creates a new execution from definition.
 func (k *Keeper) Create(ctx sdk.Context, msg types.MsgCreateExecution) (*executionpb.Execution, error) {
+	price, err := sdk.ParseCoins(msg.Request.Price)
```
antho1404: not needed if we use the coin type

ext/xvalidator/validator.go
```diff
@@ -67,7 +67,15 @@ func NewWithPrefix(prefix string) (*validator.Validate, ut.Translator) {
...
 	validate.RegisterValidation("accaddress", IsAccAddress)
 	validate.RegisterTranslation("accaddress", trans, func(ut ut.Translator) error {
-		return ut.Add("hash", "{0} must be a valid address", false)
+		return ut.Add("address", "{0} must be a valid address", false)
+	}, func(ut ut.Translator, fe validator.FieldError) string {
+		t, _ := ut.T(fe.Tag(), fe.Field(), prefix)
+		return t
+	})
+
+	validate.RegisterValidation("coins", IsCoins)
```
antho1404: not needed if we use the coin type

NicolasMahe: I think this will style be useful to check the denom and the amount is correctly formated. but it need to be updated to use directly the `Coins` type.
Or does the marshalling function of `Coins` already check the format?

krhubert:
Ok, left note - we can change Execution.Price to Coins in next cosmos-sdk release (please check CI for error). Also we can include Coin directly in protobuf as they move some types to proto (including coin)

So this is ready to merge, just need to reset hard one last commit.

NOTE:
https://github.com/cosmos/cosmos-sdk/blob/master/types/types.proto#L14 <- Coin definition that is lacking in our current cosmos-sdk version.

protobuf/api/execution.proto
```diff
@@ -70,6 +70,12 @@ message CreateExecutionRequest {
...
```
antho1404: Are you sure about this type in string?
I would import and use this type https://github.com/cosmos/cosmos-sdk/blob/master/types/types.proto#L14

protobuf/types/execution.proto
```diff
@@ -109,4 +108,11 @@ message Execution {
     (gogoproto.customtype) = "github.com/mesg-foundation/engine/hash.Hash",
     (gogoproto.nullable) = false
   ];
+
+  // price of running the exeuction.
+  string price = 14 [
+    (gogoproto.moretags) = 'hash:"name:14"',
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Coins",
+    (gogoproto.nullable) = false
```
NicolasMahe:
```suggestion
```

NicolasMahe:
> NOTE:
> https://github.com/cosmos/cosmos-sdk/blob/master/types/types.proto#L14 <- Coin definition that is lacking in our current cosmos-sdk version.

We can copy paste the coins type into our own proto until they release it.
Like this the api / clients will not have to change again?

protobuf/types/execution.proto
krhubert: Why you want to remove nullable? sdk.Coins is an array so you can still set null on it.

krhubert:
> We can copy paste the coins type into our own proto until they release it.
Like this the api / clients will not have to change again?

We can, but it will be diffrent type and we will have to cast the types.

> e2e are failling:

I know. I left that PR at this state to show you where we are with using custom types.

PR #1680, opened 2020-02-24
add block height to execution

krhubert:
close https://github.com/mesg-foundation/engine/issues/1631

dep https://github.com/mesg-foundation/engine/pull/1670


protobuf/types/execution.proto
```diff
@@ -114,4 +114,9 @@ message Execution {
   string price = 14 [
     (gogoproto.moretags) = 'hash:"name:14" validate:"coins"'
   ];
+
+  // blockHeight where the execution was included into blockchain.
+  int64 blockHeight = 15 [
+    (gogoproto.moretags) = 'hash:"-"'
```
NicolasMahe: this should be part of the hash, no?

x/execution/internal/keeper/keeper.go
```diff
@@ -88,6 +88,8 @@ func (k *Keeper) Create(ctx sdk.Context, msg types.MsgCreateExecution) (*executi
 		msg.Request.Tags,
 		msg.Request.ExecutorHash,
 	)
+	exec.BlockHeight = ctx.BlockHeight()
```
NicolasMahe: could be include in the function `executionpb.New(`

NicolasMahe:
what about other resources?
We could have exactly the same on all resources.
@antho1404 @krhubert what do you think?

protobuf/types/execution.proto
antho1404: If we do that, we cannot be sure that the hash calculated on the client or even during the processing will be the same. What happens if the execution is created on one block but in the mempool for 5 blocks.

x/execution/internal/keeper/keeper.go
antho1404: I like that here it's really something extra that is not really part of the execution, just some kind of metadata

antho1404:
Let's do it on the executions first and see how it goes, I feel this is a bit hacky and we might not need it so I would like to see first with the execution if it's something really required before we touch other resources if we don't need it yet

krhubert:
> what about other resources?
We could have exactly the same on all resources.

We don't have a need for it right now. We can always add it later.

PR #1683, opened 2020-02-26
Make sure to call end block for each modules

antho1404:
The `EndBlocker` of the module was not called by the module and also these `EndBlock` for each module were not called because of a wrong initialization of the app.

### DIFF
--- a/x/execution/module.go
+++ b/x/execution/module.go
@@ -134,6 +134,7 @@ func (am AppModule) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) {
 
 // EndBlock returns the end blocker for the execution module. It returns no validator
 // updates.
-func (AppModule) EndBlock(_ sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+	EndBlocker(ctx, am.keeper)
 	return []abci.ValidatorUpdate{}
 }
--- a/x/instance/module.go
+++ b/x/instance/module.go
@@ -134,6 +134,7 @@ func (am AppModule) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) {
 
 // EndBlock returns the end blocker for the instance module. It returns no validator
 // updates.
-func (AppModule) EndBlock(_ sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+	EndBlocker(ctx, am.keeper)
 	return []abci.ValidatorUpdate{}
 }
--- a/x/ownership/module.go
+++ b/x/ownership/module.go
@@ -134,6 +134,7 @@ func (am AppModule) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) {
 
 // EndBlock returns the end blocker for the ownership module. It returns no validator
 // updates.
-func (AppModule) EndBlock(_ sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+	EndBlocker(ctx, am.keeper)
 	return []abci.ValidatorUpdate{}
 }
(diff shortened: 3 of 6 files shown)


app/app.go
```diff
@@ -269,13 +269,6 @@ func NewInitApp(
 
 		staking.NewAppModule(app.stakingKeeper, app.accountKeeper, app.supplyKeeper),
 	)
-	// During begin block slashing happens after distr.BeginBlocker so that
-	// there is nothing left over in the validator fee pool, so as to keep the
-	// CanWithdrawInvariant invariant.
-
-	app.mm.SetOrderBeginBlockers(distr.ModuleName, slashing.ModuleName)
```
antho1404: This is deleted because `module.NewManager` is already creating ordering this based on the order of the module we give.

x/execution/module.go
```diff
@@ -134,6 +134,7 @@ func (am AppModule) BeginBlock(ctx sdk.Context, req abci.RequestBeginBlock) {
 
 // EndBlock returns the end blocker for the execution module. It returns no validator
 // updates.
-func (AppModule) EndBlock(_ sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+	EndBlocker(ctx, am.keeper)
```
krhubert: Hum end blocker is empty in each module. What is the point of calling it?

app/app.go
krhubert: But it won't be the same order.

krhubert:
> The EndBlocker of the module was not called by the module and also these EndBlock for each module were not called because of a wrong initialization of the app.

Hum, I'm not sure why this code works and how, but EndBlock is empty and dose nothing in each module.

Also you have change BeginBlockers order which on which I'm not sure if they are good.
If you take a look on cosmos/sdk-tutorial , the line is missing is in fact:

```go
  app.mm.SetOrderBeginBlockers(distr.ModuleName, slashing.ModuleName)
  >>> app.mm.SetOrderEndBlockers(staking.ModuleName)
```

So I would rather add only this one and check if its' working.

x/execution/module.go
antho1404: Making sure that it's called when we need it, I was experimenting with the pagination and found this issue

app/app.go
antho1404: Is there any specific reason why we have this order and why we remove other modules in the first place?

If you look at all the different modules, half of them are implementing the `BeginBlock` function 
- [distribution](https://github.com/cosmos/cosmos-sdk/blob/165e134405c5e4f2a1115e83321abd1328feb9c2/x/distribution/module.go#L146)
- [evidence](https://github.com/cosmos/cosmos-sdk/blob/165e134405c5e4f2a1115e83321abd1328feb9c2/x/evidence/module.go#L160)
- [mint](https://github.com/cosmos/cosmos-sdk/blob/165e134405c5e4f2a1115e83321abd1328feb9c2/x/mint/module.go#L129)
- [slashing](https://github.com/cosmos/cosmos-sdk/blob/165e134405c5e4f2a1115e83321abd1328feb9c2/x/slashing/module.go#L143)
- [staking](https://github.com/cosmos/cosmos-sdk/blob/165e134405c5e4f2a1115e83321abd1328feb9c2/x/staking/module.go#L169)
- [upgrade](https://github.com/cosmos/cosmos-sdk/blob/165e134405c5e4f2a1115e83321abd1328feb9c2/x/upgrade/module.go#L128)

Why do we only activate the distribution and slashing in the first place?

antho1404: Also, the order is the same as we add `distribution` before `slashing` in the `module.NewManager`

antho1404: code reverted for now, we can think about that later

antho1404:
I just reverted the modification on the begin/end order. Now the PR just call the `EndBlocker`. Not something critical as it is not called anymore but at least it's ready when we want to call it.

krhubert:
What about calling Order explicit like we disscuss? 

I'm ok to merge this one too, although as @NicolasMahe point it's not necessary. Also if you want to be consistent you should call BeginBlock as well :P

antho1404:
I just wanted to revert it and maybe put it in another PR as it creates too much discussion and I want to try to have more details about that.
`BeginBlock` calls `BeginBlocker` correctly in all the modules, only the `EndBlocker` was not called.
I agree it's not a "necessary" PR for now but I was playing with the end of blocks and got stuck for a long time because of this bug so if it helps for next time not being stuck for nothing it's good isn't it? This is still a bug as a function of our module is never called and will never be called without this fix.

PR #1684, opened 2020-02-27
Execution pagination 

antho1404:
This is a really basic proposal for the pagination just to show the idea I have in mind for the pagination (and code might be better than words for that).

Basically, at every end of blocks, iterate on the events and index the one we want based on specific criteria. Of course, the "database" here is just an array but it should be a persistent database like a leveldb or tmdb.

We can give this database to the keeper so that the keeper can access this index too.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -126,6 +126,8 @@ type NewApp struct {
 
 	// simulation manager
 	sm *module.SimulationManager
+
+	executionIndexer []string
 }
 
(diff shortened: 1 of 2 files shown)


antho1404:
Archiving this PR. We implemented a basic pagination at the API/CLI level for now and will probably move to a general index (maybe even outside of the cosmos app).

PR #1688, opened 2020-02-27
Handle withdraw from resources

krhubert:
About test with cli:

- there is no possible way to test full path with cli, because we don't support tx (besides ownership)
- for ownership tx cli uses `keys.Keyring` and cosmos still uses old Kyebase which doesn't work together
- for cli we can only use `query` subcommands
- we can test only path to create a message by running:

```
➜  engine git:(feature/withdraw-coins) ✗ ./bin/mesg-cli keys add test-user

- name: test-user
  type: local
  address: mesgtest1h5puw4ec89ntgzsml7uq64xc9ez996wuxxrdl8
  pubkey: mesgtestpub1addwnpepq09gg39ttn6lpg0aml9sdzvm90gdtgka5yw5arkkpn52578rfalxxgf835s
  mnemonic: ""
  threshold: 0
  pubkeys: []


**Important** write this mnemonic phrase in a safe place.
It is the only way to recover your account if you ever forget your password.

aerobic inside mail main scrub rigid wedding body seed lesson nose venue swamp chicken panther debate glow know novel horror unknown try sell build
➜  engine git:(feature/withdraw-coins) ✗ ./bin/mesg-cli tx ownership withdraw-coins AXg1nPSCGsdkzbr5N5wLq2Z5CXwxYneXFXHTQfUAUqx5 1atto --from test-user
ERROR: unknown address: account mesgtest1h5puw4ec89ntgzsml7uq64xc9ez996wuxxrdl8 does not exist
```

About test with rest:

```
+                       fmt.Fprintf(os.Stderr, "%s\n", acc.GetAddress().String())
+                       fmt.Fprintf(os.Stderr, "%s\n", testServiceHash.String())
+                       fmt.Fprintf(os.Stderr, "%s\n", sdk.AccAddress(crypto.AddressHash(testServiceHash)).String())
+                       fmt.Fprintf(os.Stderr, "sleep")
+                       time.Sleep(5 * time.Minute)
```

Put that in e2e tests (under `"withdraw from service"` tests) so it will configure all process and then you can send it with

```
curl -i --location --request POST '127.0.0.1:1317/ownership/withdraw-coins' \
--header 'Content-Type: application/json' \
--data-raw '{
        "amount": "700atto",
        "hash": "B8MQUF4egu9Fjfz3TuYB5ZhUtvRjhEe9XuQHAeiuo5fi",
        "base_req": {
                "memo": "glimpse upon body vast economy give taxi yellow rabbit come click ranch chronic hammer sport near rotate  charge lumber chicken cloud base thing forum",
                "account_number": "0",
                "gas": "200000",
                "sequence": "1",
                "gas_adjustment": "1.2",
                "chain_id": "mesg-dev-chain",
                "from": "mesgtest1sfjx3xmrnrllhc86jhfzdpqpckn4s69a5tpr4f"
        }
}'
```

You can check balances with :
```
curl --location --request GET '127.0.0.1:1317/bank/balances/mesgtest1sfjx3xmrnrllhc86jhfzdpqpckn4s69a5tpr4f'
```
```
curl --location --request GET '127.0.0.1:1317/bank/balances/mesgtest1vf42ul0gc39p5ml29rmvk46fesdakjuk8hs9fc'

```

Hum, as now I can see this tx response is not processed by network... need to debug why. In logs it occurs as recived.


antho1404:
Didn't we say that the withdraw was a resource-based handler? so like that the api is simpler for users `/service/withdraw` etc...?

Just to clarify, am I correct with the workflow of this feature?
- The user wants to withdraw a token from a resource
- Get the hash of the resource
- Find the ownership associated with his address/resource hash
- Call the `/ownership/withdraw` with the hash of the ownership

Am I right?

Isn't it easier to have something like:
- The user wants to withdraw a token from a resource
- Get the hash of the resource
- Call the `/[resource]/withdraw` with the hash of the resource

Or

- The user wants to withdraw a token from a resource
- Get the hash of the resource
- Call the `/ownership/withdraw` with the hash and type of the resource

krhubert:
> Didn't we say that the withdraw was a resource-based handler? so like that the api is simpler for users /service/withdraw etc...?

There was 3 propositions:

- per resource (suggested by @antho1404)
- separate module (suggested by @krhubert)
- and ownership (suggested by @NicolasMahe)

> Am I right?

Yep :)

> Isn't it easier to have something like:

I'm not sure if I follow. In above 3 types of withdraw-coins modules we need to pass `Hash` and `Amount` so what do think is exactly easier?

the bullet you skiped `Find the ownership associated with his address/resource hash` has to be done every time because it validates if caller Owner is in fact the owner of the resource.

antho1404:
The part to find the ownership is required for verification but the API might not need it.
We can have either
- `/ownership/withdraw` with `amount`, `resource_type` and `hash` (that will raise an error if you don't have ownership on this resource)
- `/service/withdraw` with `amount`, `hash` (that will raise an error if you don't have ownership on this resource)

krhubert:
I'll copy-paste talk with @NicolasMahe on discord.

`"/ownership/withdraw"`
If we had module like `bank` `coins` etc then withdraw makes sense because it gives context
otherwise, we need to withdraw-coins (or something), because we can withdraw ownership for example and this is not what this endpoint dose.

Withdraw alone isn't related only to money.

>  We can have either ....

Why we need resource_type for ownership/withdraw? it's not required here at all.

So we need to choose some option (we have 3) and everyone wants to pick something else. Either someone will be out-voted or I'll just leaving it as it is, because we have functionality and we can change the api any time.

krhubert:
Add runner to ownership + withdraw on deletion. need to find a way to use client in e2e tests (not lcd server).

x/ownership/internal/keeper/keeper.go
```diff
@@ -77,12 +82,29 @@ func (k Keeper) Delete(ctx sdk.Context, owner sdk.AccAddress, resourceHash hash.
 		return fmt.Errorf("resource %q do not have any ownership", resourceHash)
 	}
 
+	// transfer all spendable coins from resource address to owner
+	addr := sdk.AccAddress(crypto.AddressHash(resourceHash))
+	coins := k.bankKeeper.GetCoins(ctx, addr)
+	if err := k.bankKeeper.SendCoins(ctx, addr, owner, coins); err != nil {
```
antho1404: Are we sure that this will not raise an error in case the resource (eg: runner) didn't receive any funds.
If so then we have an issue and cannot delete a runner that we just started or was never used.

krhubert: > Are we sure that this will not raise an error in case the resource (eg: runner) didn't receive any funds.

What type of error? We must be able to send coins. If you are asking if there is no coins on runner acc, then send coins will return immediately.

antho1404: Could you make sure that the following flow works:
- Start a runner for a specific service
- Don't create any executions for this runner
- Stop the runner
This should be possible with no error, I'm just not sure here this is possible because `SendCoins` will return an error and so we will not be able to stop a runner.

krhubert: Fix the bug, nice catch 😎

e2e/api_test.go
```diff
@@ -85,6 +85,8 @@ func TestAPI(t *testing.T) {
 	require.NoError(t, err)
 	httpclient, err := rpcclient.NewHTTP("http://localhost:26657", "/websocket")
 	require.NoError(t, err)
+	require.NoError(t, httpclient.OnStart())
```
NicolasMahe: We should not call `OnStart`. It will be called by `Start`. The tendermint node is using the same `service` struct. From outside, we should only call `Start` and `Stop`
```suggestion
	require.NoError(t, httpclient.Start())
```

e2e/api_test.go
```diff
@@ -73,16 +74,20 @@ func TestAPI(t *testing.T) {
...
-		panic(err)
-	}
+	require.NoError(t, err)
+
 	cosmos.CustomizeConfig(cfg)
 
 	conn, err := grpc.DialContext(context.Background(), "localhost:50052", grpc.WithInsecure())
 	require.NoError(t, err)
 
 	kb, err := cosmos.NewKeybase(filepath.Join(cfg.Path, cfg.Cosmos.RelativePath))
```
NicolasMahe: the ci error is still the same:
```
couldn't create db: Error initializing DB: stat /home/circleci/project/e2e.test/mesg/cosmos/keys/keys.db: permission denied
```
simply use another folder. eg:
```suggestion
	kb, err := cosmos.NewKeybase(filepath.Join(cfg.Path, cfg.Cosmos.RelativePath, 'e2e-keybase'))
```

x/ownership/internal/keeper/keeper.go
antho1404: Can we do a safe test here, it's not really clear that `SendCoins` will manage correctly if we send empty coin and this might change in the future.

Can we do something like
```go
if coins != 0 {
  if err := k.bankKeeper.SendCoins(ctx, addr, owner, coins); err != nil {
    return err
  }
}
```

At least we will know that we don't send if it's 0, also, in that case, we will not emit an event with empty value.

e2e/execution_test.go
```diff
@@ -254,6 +256,33 @@ func testExecution(t *testing.T) {
 			lcdGet(t, "bank/balances/"+execAddress.String(), &coins)
 			require.True(t, coins.AmountOf("atto").Equal(sdk.NewInt(0)))
 		})
+
+		// check balance of execution
```
NicolasMahe:
```suggestion
```

antho1404:
I have some issue testing it outside of the engine.

## Test

### Create a service
```
mesg-cli service:create "$(mesg-cli service:compile https://github.com/mesg-foundation/service-js-function)"
```

### Start the service
```
mesg-cli service:start js-function
```

### Create an execution: 
```
mesg-cli service:execute js-function execute --json ~/prog/MESG/services/service-js-function/test.json
```

### Get the ownership hash
```
grpcurl --proto ./protobuf/api/ownership.proto -plaintext -d "{}" localhost:50052 mesg.api.Ownership/List | jq '.ownerships' | jq '.[] | select(.resource | contains("Runner"))' | jq '.hash'
```

### Convert this hash in base58
`npm i @mesg/api` then
```
node -e "console.log(require('@mesg/api/lib/util/base58').encode(Buffer.from('aM1cXHVZDGdV8e7WnVGQta2e9hrDO9t0h11Q1/G35LE=', 'base64')))"
```

### Import account on the CLI
Use the same mnemonic than the one in the engine.
```
./bin/mesg-cli keys add -i test-account
```

### Withdraw
```
./bin/mesg-cli tx ownership withdraw-coins 8472KKYT1CvtdpEXnaVAWyLwrE77mvYk1g9xxJTMoNZE 9000atto --from mesgtest1yztk7h0la9e2wdhed3h37sd50szq8hsvrvgump --chain-id mesg-dev-chain --fees 200000atto -b block
```

## Result

```
code: 4
data: ""
rawlog: 'unauthorized: address mesgtest1yztk7h0la9e2wdhed3h37sd50szq8hsvrvgump is
  not owner of resource 8472KKYT1CvtdpEXnaVAWyLwrE77mvYk1g9xxJTMoNZE: failed to execute
  message; message index: 0'
```

Is there a bug or am I doing wrong? I'm the owner of this runner.

```
grpcurl --proto ./protobuf/api/ownership.proto -plaintext -d "{}" localhost:50052 mesg.api.Ownership/List | jq '.ownerships' | jq '.[] | select(.resource | contains("Runner"))'             

{
  "hash": "aM1cXHVZDGdV8e7WnVGQta2e9hrDO9t0h11Q1/G35LE=",
  "owner": "mesgtest1yztk7h0la9e2wdhed3h37sd50szq8hsvrvgump",
  "resourceHash": "awEFgRRDjf1lkGyZR93spEIq5n0P9QnFsV6cLXGO9as=",
  "resource": "Runner"
}
```

e2e/api_test.go
```diff
@@ -5,14 +5,18 @@ import (
 	"context"
 	"io/ioutil"
 	"net/http"
+	"os"
+	"path/filepath"
 	"testing"
 
+	"github.com/cosmos/cosmos-sdk/crypto/keys"
 	"github.com/cosmos/cosmos-sdk/types/rest"
 	"github.com/mesg-foundation/engine/app"
```
olen41, line 14: Need the payments

e2e/execution_test.go
```diff
@@ -254,6 +256,32 @@ func testExecution(t *testing.T) {
 			lcdGet(t, "bank/balances/"+execAddress.String(), &coins)
 			require.True(t, coins.AmountOf("atto").Equal(sdk.NewInt(0)))
 		})
+
+		t.Run("withdraw from service", func(t *testing.T) {
+			acc, err := cclient.GetAccount()
+			require.NoError(t, err)
+			coins := sdk.NewCoins(sdk.NewCoin("atto", sdk.NewInt(7000)))
```
olen41, line 263: Need coins

e2e/execution_test.go
```diff
@@ -254,6 +256,32 @@ func testExecution(t *testing.T) {
 			lcdGet(t, "bank/balances/"+execAddress.String(), &coins)
 			require.True(t, coins.AmountOf("atto").Equal(sdk.NewInt(0)))
 		})
+
+		t.Run("withdraw from service", func(t *testing.T) {
```
olen41, line 260: Need to put on go 2bank

olen41 (approved):

PR #1689, opened 2020-02-27
Add pagination system on http handler of executions/list

NicolasMahe:
Added a simple and not optimized pagination system on `executions/list` HTTP endpoint so the explorer can fetch executions without breaking.

The parameters `page`, `limit` and `sort=reverse` can be passed to the endpoint to return different executions:

```
curl 'localhost:1317/execution/list?page=12&limit=1&sort=reverse'
```

The sort is based on the execution's blockHeight

Related to https://github.com/mesg-foundation/engine/issues/1679

--- 

Improvements that can be done now or later:
1. Move the pagination and sort system in the module's querier but need to pass as parameters page, limit and sort.
2. Move the pagination and sort system in the module's keeper and use [`KVStorePrefixIteratorPaginated`](https://github.com/cosmos/cosmos-sdk/blob/master/store/types/iterator.go) but need to pass as parameters page, limit and find a way to sort the data without iterate on all of them (index?)

### DIFF
--- a/x/execution/client/rest/query.go
+++ b/x/execution/client/rest/query.go
@@ -73,13 +73,13 @@ func queryListHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 			return
 		}
 
-		sortedExecs := sort.Interface(execution.ByBlockHeight(execs))
+		sortFunc := sort.Interface(execution.ByBlockHeight(execs))
 		for _, tag := range tags {
 			if tag == "sort='reverse'" {
-				sortedExecs = sort.Reverse(sortedExecs)
+				sortFunc = sort.Reverse(sortFunc)
 			}
 		}
-		sort.Sort(sortedExecs)
+		sort.Sort(sortFunc)
 
 		start, end := client.Paginate(len(execs), page, limit, limit)
 		if start < 0 || end < 0 {
--- a/x/execution/client/rest/query.go
+++ b/x/execution/client/rest/query.go
@@ -53,7 +53,7 @@ func queryListHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 			return
 		}
 
-		tags, page, limit, err := rest.ParseHTTPArgs(r)
+		_, page, limit, err := rest.ParseHTTPArgs(r)
 		if err != nil {
 			rest.WriteErrorResponse(w, http.StatusBadRequest, err.Error())
 			return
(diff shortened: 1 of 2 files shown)


x/execution/client/rest/query.go
```diff
@@ -58,7 +67,28 @@ func queryListHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
...
+		for _, tag := range tags {
+			if tag == "sort='reverse'" {
+				sortFunc = sort.Reverse(sortFunc)
+			}
+		}
+		sort.Sort(sortFunc)
+
+		start, end := client.Paginate(len(execs), page, limit, limit)
+		if start < 0 || end < 0 {
+			execs = make([]*execution.Execution, 0)
```
krhubert: We should return error if the pagination boundaries are invalid (so in that case check if page/limit <= 0), otherwise returning empty list.

`s/execs = make([]*execution.Execution, 0)/execs = nil/`

x/execution/client/rest/query.go
```diff
@@ -58,7 +67,28 @@ func queryListHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
...
+			return
+		}
+
+		sortFunc := sort.Interface(execution.ByBlockHeight(execs))
+		for _, tag := range tags {
+			if tag == "sort='reverse'" {
+				sortFunc = sort.Reverse(sortFunc)
+			}
+		}
+		sort.Sort(sortFunc)
```
antho1404: I'm not sure why we need this customization. I would just force a descending order on the block height
```suggestion
		sort.Reverse(sort.Interface(execution.ByBlockHeight(execs)))
```

NicolasMahe: I was testing what the tags were and implemented the reverse sort with it.
I can revert if it's not useful

antho1404: I would KISS especially as it is not a definitive code and will certainly be replaced when we have a better pagination system.

x/execution/client/rest/query.go
NicolasMahe: the page and limit <= 0 error is managed by `rest.ParseHTTPArgs(r)`:
```go
} else if page <= 0 {
	return tags, page, limit, errors.New("page must greater than 0")
}
```
```go
} else if limit <= 0 {
	return tags, page, limit, errors.New("limit must greater than 0")
}
```
---
the problem with returning `execs = nil` is it's not a list anymore in the json response:
```json
{"height":"757","result":null}
```
so we will have issue with clients.
i would keep it as it is:
```
{"height":"773","result":[]}
```

i replaced the `make([]*execution.Execution, 0)` by simply `[]*execution.Execution{}`

NicolasMahe:
@krhubert @antho1404 modif done. please check again

PR #1691, opened 2020-02-28
Add min price for execution

antho1404:
Right now an execution can be created with 0 tokens.

I added the support of a minimum that is set for now to `10000atto`. This parameter is part of the module and present in the genesis.
I also added a config to have a default price from the clients (api/orchestrator)

The good thing of having this parameter in the genesis is that we could do some vote to change this limit and we can play with proposal/votes.

### DIFF
--- a/orchestrator/orchestrator.go
+++ b/orchestrator/orchestrator.go
@@ -18,12 +18,13 @@ import (
 )
 
 // New creates a new Process instance
-func New(mc *cosmos.ModuleClient, ep *publisher.EventPublisher) *Orchestrator {
+func New(mc *cosmos.ModuleClient, ep *publisher.EventPublisher, execPrice string) *Orchestrator {
 	return &Orchestrator{
-		mc:    mc,
-		ep:    ep,
-		ErrC:  make(chan error),
-		stopC: make(chan bool),
+		mc:        mc,
+		ep:        ep,
+		ErrC:      make(chan error),
+		stopC:     make(chan bool),
+		execPrice: execPrice,
 	}
 }
 
@@ -273,6 +274,7 @@ func (s *Orchestrator) processTask(nodeKey string, task *process.Process_Node_Ta
 		TaskKey:      task.TaskKey,
 		Inputs:       data,
 		ExecutorHash: executor.Hash,
+		Price:        s.execPrice,
 		Tags:         nil,
 	})
 	return err
--- a/x/execution/internal/keeper/params.go
+++ b/x/execution/internal/keeper/params.go
@@ -11,14 +11,10 @@ const (
 )
 
 // MinPrice - Minimum price of an execution
-func (k Keeper) MinPrice(ctx sdk.Context) sdk.Coins {
-	var coinstr string
-	k.paramstore.Get(ctx, types.KeyMinPrice, &coinstr)
-	price, err := sdk.ParseCoins(coinstr)
-	if err != nil {
-		panic(err)
-	}
-	return price
+func (k Keeper) MinPrice(ctx sdk.Context) string {
+	var coins string
+	k.paramstore.Get(ctx, types.KeyMinPrice, &coins)
+	return coins
 }
 
 // SetParams will populate all the params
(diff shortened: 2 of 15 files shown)


krhubert:
Why this should be set globally? I think that value should be set by service creator (or even voting by service creator and runners). What do you think?

orchestrator/orchestrator.go
```diff
@@ -273,6 +274,7 @@ func (s *Orchestrator) processTask(nodeKey string, task *process.Process_Node_Ta
 		TaskKey:      task.TaskKey,
 		Inputs:       data,
 		ExecutorHash: executor.Hash,
+		Price:        s.execPrice,
```
krhubert: I think orchestrator shouldn't be responsible anyhow for setting the min price. Espessialy if the executor is chosenen randomly from runners list.

antho1404:
Totally agree that the service creator/runner should set the price, this is just a min value that can later be set to 0 if necessary but the main goal here is to be able to actually see this economy. For now, everything is ready for it but we still accept free executions.

x/execution/internal/types/params.go
```diff
@@ -1,38 +1,66 @@
...
-}
+var (
+	// KeyMinPrice key for the parameter MinPrice
+	KeyMinPrice = []byte("MinPrice")
+)
 
 // Params - used for initializing default parameter for instance at genesis
-type Params struct{}
+type Params struct {
+	MinPrice string `json:"min_price" yaml:"min_price"` // min price to pay for an execution
```
NicolasMahe:
```suggestion
	MinPrice string `json:"minPrice" yaml:"minPrice"` // min price to pay for an execution
```
why snake case here?

x/execution/internal/keeper/params.go
```diff
@@ -0,0 +1,34 @@
...
+	DefaultParamspace = types.ModuleName
+)
+
+// MinPrice - Minimum price of an execution
+func (k Keeper) MinPrice(ctx sdk.Context) sdk.Coins {
+	var coinstr string
+	k.paramstore.Get(ctx, types.KeyMinPrice, &coinstr)
+	price, err := sdk.ParseCoins(coinstr)
+	if err != nil {
+		panic(err)
```
NicolasMahe: can't you return the error here?

antho1404: I delegated the parsing to the calling function

antho1404:
> what about also putting the var runnerShare = sdk.NewDecWithPrec(9, 1) in params?

Didn't want to change this on this PR but yes definitely something that we could add later.

orchestrator/orchestrator.go
antho1404: This will be the responsibility of the orchestrator because it is the one triggering the execution. After I agree now it's really simple and the price calculation should be much more complex and not only a constant. This price will be calculated based on the price defined by the service/runner.

NicolasMahe:
The error are pretty bad:
from the cli:
```
➜  engine git:(feature/default-exec-price) ✗ mesg-cli service:execute 9pifx2Hxq9sbJdhDRMKy2hevt9d7d2JbLiyJ6zsSSYEM task1 --data msg=Hello
 ›   Error: 2 UNKNOWN: transaction returned with invalid code 11
```

from the engine's logs
```
INFO[0140] Rejected bad transaction                      err="<nil>" module=mempool peerID= res="&{code:11 log:\"out of gas: out of gas in location: ReadFlat; gasWanted: 22261, gasUsed: 23058\" gas_wanted:22261 gas_used:23058 codespace:\"sdk\" }" tx=8D9DC7D6633960456FC0D71A3B131B5D68CC566574D2289D1E32F77F9A5EE4BC
ERRO[0140] finished unary call with code Unknown         error="transaction returned with invalid code 11" grpc.code=Unknown grpc.method=Create grpc.service=mesg.api.Execution grpc.start_time="2020-03-02T05:11:26Z" grpc.time_ms=2.777 module=grpc span.kind=server system=grpc
```

we really need to improve the returned error...

PR #1692, opened 2020-02-29
Custom hash serializer

NicolasMahe:
Simplification of https://github.com/mesg-foundation/engine/pull/1678
Closes https://github.com/mesg-foundation/engine/issues/1667

This PR implements a simple and very performant serializer that requires every struct to implement a `func HashSerialize() string`.

I tried different implementations and compared them in terms of performance and UX of its API.
- hashing child-structures is a huge performance loss so it simply puts the children serialized string as value
- sorting is a medium performance loss so the developer has to make sure to put data in the right order
- using multiple times `buffer.WriteString` instead of constructing string with `fmt.Sprintf` is a medium performance improvement

I also implement the same serializer in JS to make sure it's also simple to implement and check the results.

I ran some benchmarks on a simple service struct, the difference compared to the current structhash version is an x13 of performance improvement.

Here are the benchmarks i ran:
```
// previous version based on structhase
// BenchmarkSerialize-4   	   34251	     33255 ns/op	    5208 B/op	     111 allocs/op
// BenchmarkSerialize-4   	   35149	     33629 ns/op	    5208 B/op	     111 allocs/op
// BenchmarkSerialize-4   	   33976	     38833 ns/op	    5208 B/op	     111 allocs/op

// first implementation with base58
// BenchmarkSerialize-4   	   45284	     26965 ns/op	    7928 B/op	     216 allocs/op
// BenchmarkSerialize-4   	   44287	     27114 ns/op	    7928 B/op	     216 allocs/op
// BenchmarkSerialize-4   	   45910	     26025 ns/op	    7928 B/op	     216 allocs/op

// first implementation without base58
// BenchmarkSerialize-4   	   98306	     11453 ns/op	    5848 B/op	     188 allocs/op
// BenchmarkSerialize-4   	   95521	     11667 ns/op	    5848 B/op	     188 allocs/op
// BenchmarkSerialize-4   	  106831	     11040 ns/op	    5848 B/op	     188 allocs/op

// same as previous one + eliminate empty value as soon as possible
// BenchmarkSerialize-4   	  130498	      8952 ns/op	    4352 B/op	     149 allocs/op
// BenchmarkSerialize-4   	  121597	     10585 ns/op	    4352 B/op	     149 allocs/op
// BenchmarkSerialize-4   	  127125	      9357 ns/op	    4352 B/op	     149 allocs/op

// same as previous one + no sort
// BenchmarkSerialize-4   	  136657	      8135 ns/op	    4000 B/op	     138 allocs/op
// BenchmarkSerialize-4   	  133406	      8202 ns/op	    4000 B/op	     138 allocs/op
// BenchmarkSerialize-4   	  137749	      8333 ns/op	    4000 B/op	     138 allocs/op

// same as previous one + bytes
// BenchmarkSerialize-4   	  234852	      4824 ns/op	    3432 B/op	      86 allocs/op
// BenchmarkSerialize-4   	  224851	      4972 ns/op	    3432 B/op	      86 allocs/op
// BenchmarkSerialize-4   	  219336	      5380 ns/op	    3432 B/op	      86 allocs/op

// same as previous one + buffer pool
// BenchmarkSerialize-4   	  244681	      4571 ns/op	    2328 B/op	      73 allocs/op
// BenchmarkSerialize-4   	  246154	      4622 ns/op	    2328 B/op	      73 allocs/op
// BenchmarkSerialize-4   	  237271	      4530 ns/op	    2328 B/op	      73 allocs/op

// same as previous one + return early on empty pairs
// BenchmarkSerialize-4   	  263187	      4401 ns/op	    2328 B/op	      73 allocs/op
// BenchmarkSerialize-4   	  236986	      4268 ns/op	    2328 B/op	      73 allocs/op
// BenchmarkSerialize-4   	  279604	      4284 ns/op	    2328 B/op	      73 allocs/op

// same as previous one + no sort
// BenchmarkSerialize-4   	  291133	      3688 ns/op	    1976 B/op	      62 allocs/op

// same as previous one + remove intermediate array to hold the key/value, direct write to a dedicated buffer (no buffer pool)
// BenchmarkSerialize-4   	  468823	      2347 ns/op	    2000 B/op	      30 allocs/op
// BenchmarkSerialize-4   	  459656	      2329 ns/op	    2000 B/op	      30 allocs/op
// BenchmarkSerialize-4   	  458227	      2260 ns/op	    2000 B/op	      30 allocs/op
```

---

TODO in another PR:
- [ ] remove hash flags from the proto files


.golangci.yml
```diff
@@ -24,6 +24,9 @@ issues:
       linters:
         - golint
       text: "exported var Translator should have its own declaration"
+    - path: hash/structhash/structhash_test.go
+      linters:
+        - maligned
```
NicolasMahe: this can be removed now that structhash is removed

service/serialize_test.go
```diff
@@ -0,0 +1,88 @@
...
+// BenchmarkSerialize-4   	  236986	      4268 ns/op	    2328 B/op	      73 allocs/op
+// BenchmarkSerialize-4   	  279604	      4284 ns/op	    2328 B/op	      73 allocs/op
+
+// last one + no sort
+// BenchmarkSerialize-4   	  291133	      3688 ns/op	    1976 B/op	      62 allocs/op
+
+// last one + no pairs, direct write to a dedicated buffer (no buffer pool)
+// BenchmarkSerialize-4   	  468823	      2347 ns/op	    2000 B/op	      30 allocs/op
+// BenchmarkSerialize-4   	  459656	      2329 ns/op	    2000 B/op	      30 allocs/op
+// BenchmarkSerialize-4   	  458227	      2260 ns/op	    2000 B/op	      30 allocs/op
```
NicolasMahe: no need to keep the benchmark result in the test

antho1404:
Didn't review in detail but really like the direction, for sure, more code duplication but a lot more control and it is now really simple to understand. 

(Blocking):
One thing that I really don't like, this is not really a generic serialization method so having `Serialize` as function name is quite bad. I would be ok with `hashserialize` or something like that. 

(Not blocking but definitely really nice to have):
Also I hear the performance issue with hashing all the sub structs but except in the process that we might have 10+ sub structs, this is negligible and I would definitely go with a function `Hash` that does the serialization internally and returns the hash this would be really easy to understand and use.

NicolasMahe:
> (Blocking):
> One thing that I really don't like, this is not really a generic serialization method so having `Serialize` as function name is quite bad. I would be ok with `hashserialize` or something like that.

done 👌 

> (Not blocking but definitely really nice to have):
> Also I hear the performance issue with hashing all the sub structs but except in the process that we might have 10+ sub structs, this is negligible and I would definitely go with a function `Hash` that does the serialization internally and returns the hash this would be really easy to understand and use.

I would like to not use hash inside the `hashserialize` function because:
- it is not needed and it adds one more step
- it creates a different behavior for "simple" value and "complex" value (why not hashing every value?)
- I don't want to force a specific string representation of the hash (we are already discussing about using bech32 instead of base58 https://github.com/mesg-foundation/engine/issues/1700)
- It creates a dependency cycle if `hashserializer` is using `hash` package

NicolasMahe:
I updated the unit test to example test to normally generate a really nice documentation.
But because it's not released, I cannot check. @krhubert do you know how to display something like https://pkg.go.dev/mod/github.com/mesg-foundation/engine but locally?

krhubert:
@NicolasMahe I don't know if go.dev can be run locally, because it's in beta, but for sure you can run someting like godoc.org with https://godoc.org/golang.org/x/tools/cmd/godoc

antho1404:
modules are not supported with `godoc -http=":6060"` unless you are running go v1.14

NicolasMahe:
> modules are not supported with `godoc -http=":6060"` unless you are running go v1.14.
> There is a workaround for it otherwise https://stackoverflow.com/questions/54933983/how-to-serve-documentation-using-godoc-together-with-go-modules

I install go v1.14 and I was not able to access the engine's package using `godoc -http=":6060"`.
The workaround you link works! I created a `doc.sh` script to run it.

NicolasMahe:
Documentation improved!
To check it, run in the engine folder:
```
godoc -http=:6060
```
and open: http://localhost:6060/pkg/github.com/mesg-foundation/engine/hash/hashserializer

hash/hashserializer/hashserializer.go
```diff
@@ -0,0 +1,181 @@
...
+		s.buf.WriteString(prefix)
+		s.buf.WriteString(":")
+		s.buf.WriteString(value)
+		s.buf.WriteString(";")
+	}
+}
+
+// AddBool adds a boolean to the serializer.
+// If value is false, nothing is added to the HashSerializer.
+func (s *HashSerializer) AddBool(prefix string, value bool) {
```
antho1404: These `AddX` method could be chainable if they return themselves  (not blocking at all)

```go
func (s *HashSerializer) AddBool(prefix string, value bool) *HashSerializer {
  if value {
	s.AddString(prefix, "true")
  }
  return s
}
```

with this we could use it like
```go
New().
  AddBool(1, true).
  AddString(2, "ttt").
  HashSerialize()
```

process/serialize.go
```diff
@@ -0,0 +1,237 @@
...
+	return ser.HashSerialize()
+}
+
+// HashSerialize returns the hashserialized string of this type
+func (data *Process_Node_Map_Output) HashSerialize() string {
+	if data == nil {
+		return ""
+	}
+	ser := hashserializer.New()
+	ser.AddString("2", data.GetStringConst())
```
antho1404: Could we add the null just to make sure we don't forget even it will be empty? That will reduce confusion.

process/serialize.go
```diff
@@ -0,0 +1,237 @@
...
+		ser.Add(strconv.Itoa(i), value)
+	}
+	return ser.HashSerialize()
+}
+
+// HashSerialize returns the hashserialized string of this type
+func (data Process_Node_Filter_Condition) HashSerialize() string {
+	ser := hashserializer.New()
+	ser.AddString("1", data.Key)
+	ser.AddInt("2", int(data.Predicate))
```
antho1404: Value of the condition is missing here and should definitely be part of the hash

process/serialize.go
```diff
@@ -0,0 +1,237 @@
...
+	ser := hashserializer.New()
+	ser.AddString("1", data.Src)
+	ser.AddString("2", data.Dst)
+	return ser.HashSerialize()
+}
+
+type processNodes []*Process_Node
+
+// HashSerialize returns the hashserialized string of this type
+func (data processNodes) HashSerialize() string {
```
antho1404: These function `processNodes` and `processEdges` could have a sort. Sorting the data in the case of a process will not change the behavior and the process (might slightly change the performance of it). As long as the process contains the same nodes and the same edges, whatever the order of the nodes or edges it will be the same process and the hash should be similar.

protobuf/types/serialize.go
```diff
@@ -0,0 +1,72 @@
...
+	return ser.HashSerialize()
+}
+
+// HashSerialize returns the hashserialized string of this type
+func (data *Value) HashSerialize() string {
+	if data == nil {
+		return ""
+	}
+	ser := hashserializer.New()
+	ser.AddFloat("2", data.GetNumberValue())
```
antho1404: same comment here about the null value, let's add it even we know for now it will be discarded.

service/serialize.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+	"strconv"
+
+	"github.com/mesg-foundation/engine/hash/hashserializer"
+)
+
+// HashSerialize returns the hashserialized string of this type
+func (data *Service) HashSerialize() string {
+	if data == nil {
```
antho1404: Not a discussion for this PR but we might simplify the hash for the service in order to have similar services producing the same hash to only
`Name + Tasks + Events + Dependencies + Sid + Source`
Even further, the source is already a hash from IPFS so we could have only this in the hash (but I like to keep the definition too)

service/serialize.go
```diff
@@ -0,0 +1,153 @@
...
+	return ser.HashSerialize()
+}
+
+// HashSerialize returns the hashserialized string of this type
+func (data Service_Configuration) HashSerialize() string {
+	ser := hashserializer.New()
+	ser.AddStringSlice("1", data.Volumes)
+	ser.AddStringSlice("2", data.VolumesFrom)
+	ser.AddStringSlice("3", data.Ports)
+	ser.AddStringSlice("4", data.Args)
```
antho1404: Not matching the current proto but I think the current proto is wrong
https://github.com/mesg-foundation/engine/blob/6c563e098f107e1fb063252aa5a3f339ab66abd3/protobuf/types/service.proto#L122-L124

service/serialize.go
```diff
@@ -0,0 +1,153 @@
...
+	ser.AddString("5", data.Command)
```
antho1404: Same here 
https://github.com/mesg-foundation/engine/blob/6c563e098f107e1fb063252aa5a3f339ab66abd3/protobuf/types/service.proto#L127-L129

service/serialize.go
```diff
@@ -0,0 +1,153 @@
...
+	ser.AddString("8", data.Key)
+	ser.AddStringSlice("9", data.Env)
+	return ser.HashSerialize()
+}
+
+type serviceTasks []*Service_Task
+
+// HashSerialize returns the hashserialized string of this type
+func (data serviceTasks) HashSerialize() string {
+	if data == nil {
```
antho1404: Same comment as the process, these tasks could be sorted as it doesn't change the behavior of the service.

service/serialize.go
```diff
@@ -0,0 +1,153 @@
...
+		ser.Add(strconv.Itoa(i), value)
+	}
+	return ser.HashSerialize()
+}
+
+type serviceEvents []*Service_Event
+
+// HashSerialize returns the hashserialized string of this type
+func (data serviceEvents) HashSerialize() string {
+	if data == nil {
```
antho1404: Same for events, sort possible

process/serialize.go
NicolasMahe: why would you like to sort them? it's an array so the order is important.
but yes, from a process point of view, it doesn't change the behavior.
even, what would you like the array with? the value of the node key?
i think here it's up to the compiler to produce the same output.

hash/hash.go
```diff
@@ -8,51 +8,32 @@ import (
...
+	return Hash(h[:])
 }
 
-// Dump takes an interface and returns its hash representation.
-func Dump(v interface{}) Hash {
-	d := New()
-	d.WriteObject(v)
-	return d.Sum(nil)
+// Sum takes a slice of byte and returns its hash.
+func Sum(v []byte) Hash {
```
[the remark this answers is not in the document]
NicolasMahe: how can `bytes.Equal` replace a sha256 sum function?

process/serialize.go
antho1404: The idea would be to sort for example by `key`. The order of this array doesn't impact the actual structure of the graph.
It could definitely be done in the compiler too but might be interesting here as it's the final step but anyway this is not something related to this PR.

service/serialize.go
NicolasMahe: i fix the proto

process/serialize.go
NicolasMahe: could you create an issue to furthe discuss thus

hash/hash.go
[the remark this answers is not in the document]
krhubert: as I can see it is used like
```go
require.True(t, hash.Sum(structValueSort1).Equal(hash.Sum(structValueSort2))) 
```

And this could be written as 
```go
require.True(t, bytes.Equals(structValueSort1, structValueSort2))
```

Because you don't need to convert it to hash to compare the marshal output.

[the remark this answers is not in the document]
NicolasMahe: the struct are not equals, but their hash-serialized is

process/serialize.go
antho1404: https://github.com/mesg-foundation/engine/issues/1707

service/serialize.go
antho1404: https://github.com/mesg-foundation/engine/issues/1708

service/serialize.go
antho1404: https://github.com/mesg-foundation/engine/issues/1708

service/serialize.go
antho1404: https://github.com/mesg-foundation/engine/issues/1709

hash/hash.go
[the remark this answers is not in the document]
krhubert: But here you can pass and compare hash for two byte slices. We can keep it if you feel there a need for that. Just leaving the change and tests below

```
➜  types git:(feature/simple-hash) ✗ git diff
diff --git a/protobuf/types/struct_test.go b/protobuf/types/struct_test.go
index b60ed9d0..c2ed0177 100644
--- a/protobuf/types/struct_test.go
+++ b/protobuf/types/struct_test.go
@@ -1,6 +1,7 @@
 package types
 
 import (
+       bytes "bytes"
        "testing"
 
        "github.com/cosmos/cosmos-sdk/codec"
@@ -91,7 +92,7 @@ func TestStructMarshal(t *testing.T) {
                        require.NoError(t, err)
                        structValueSort2, err = cdc.MarshalBinaryLengthPrefixed(structSort2)
                        require.NoError(t, err)
-                       require.True(t, hash.Sum(structValueSort1).Equal(hash.Sum(structValueSort2)))
+                       require.True(t, bytes.Equal(structValueSort1, structValueSort2))
                })
                t.Run("Unmarshal", func(t *testing.T) {
                        require.NoError(t, cdc.UnmarshalBinaryLengthPrefixed(structValueSort1, &structUnm1))
@@ -114,7 +115,7 @@ func TestStructMarshal(t *testing.T) {
                        require.NoError(t, err)
                        structValueSort2, err = cdc.MarshalJSON(structSort2)
                        require.NoError(t, err)
-                       require.True(t, hash.Sum(structValueSort1).Equal(hash.Sum(structValueSort2)))
+                       require.True(t, bytes.Equal(structValueSort1, structValueSort2))
                })
                t.Run("Unmarshal", func(t *testing.T) {
                        require.NoError(t, cdc.UnmarshalJSON(structValueSort1, &structUnm1))
➜  types git:(feature/simple-hash) ✗ go test
PASS
ok  	github.com/mesg-foundation/engine/protobuf/types	0.042s
```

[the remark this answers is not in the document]
NicolasMahe: I changed the test but I would like to keep the function as it is useful in an upcoming PR ;)

NicolasMahe:
POC js implementation: https://github.com/mesg-foundation/js-sdk/pull/182

PR #1699, opened 2020-03-03
Get coins from execution's process if attached

krhubert:
dependant on https://github.com/mesg-foundation/engine/pull/1688

close https://github.com/mesg-foundation/engine/issues/1698


antho1404:
Let's merge it when we are sure that we can credit a process as well. 
Right now we cannot credit a process so every process will fail because their balances will be empty.
We need to make sure that this feature will not break the processes. Let's merge when we have either:
- A way on the client to calculate the address of the process and send a transfer (like https://github.com/mesg-foundation/engine/pull/1704)
- A way to deposit to the process (a new handler to deposit on the process or a balance when we create the process)

This is blocking, even this PR is working, it needs to be merged only when the client is updated otherwise all processes will be broken.

NicolasMahe:
TODO:
- [ ] take coins from signer when create a process (hardcode everything)

e2e/orchestrator_balance_withdraw_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/mesg-foundation/engine/execution"
+	"github.com/mesg-foundation/engine/hash"
+	"github.com/mesg-foundation/engine/process"
+	pb "github.com/mesg-foundation/engine/protobuf/api"
+	"github.com/mesg-foundation/engine/protobuf/types"
+	"github.com/stretchr/testify/require"
+	"github.com/tendermint/tendermint/crypto"
+)
+
+func testOrchestratorProcessBalanceWithdraw(executionStream pb.Execution_StreamClient, instanceHash hash.Hash) func(t *testing.T) {
```
antho1404: This test is not a withdraw test it more that the process actually paid correctly for the execution

PR #1703, opened 2020-03-05
Create a small proposal for a execution consensus

krhubert:
An updated version of execution consensus.

To be honest I'm not sure how to test it with our e2e tests.

I find a need to write a directly inside a keeper, so we reduce the testing surface in e2e tests.


protobuf/api/execution.proto
```diff
@@ -75,6 +75,9 @@ message CreateExecutionRequest {
   string price = 10 [
     (gogoproto.moretags) = 'validate:"coins"'
   ];
+
+    // confs is the number of conrimations the exeuction needs to have before chaingin status to passed
+  int64 confs = 11;
```
antho1404: let's call that "confirmations" and fix the typo in the comment

protobuf/types/execution.proto
```diff
@@ -119,4 +122,9 @@ message Execution {
   int64 blockHeight = 15 [
     (gogoproto.moretags) = 'hash:"-"'
   ];
+
+  // confs is the number of conrimations the exeuction needs to have before chaingin status to passed
+  int64 confs = 16 [
```
antho1404: same here, to rename and fix typo

protobuf/types/execution.proto
```diff
@@ -119,4 +122,9 @@ message Execution {
...
+    (gogoproto.moretags) = 'hash:"-"'
```
antho1404: If we add this data in the execution, then I think the confirmations should be part of the hash, 2 executions should be different if they have 2 different consensuses.

x/execution/internal/keeper/keeper.go
```diff
@@ -99,31 +99,53 @@ func (k *Keeper) Create(ctx sdk.Context, msg types.MsgCreateExecution) (*executi
...
+		execAddress := sdk.AccAddress(crypto.AddressHash(exec.Hash))
+		if err := k.bankKeeper.SendCoins(ctx, msg.Signer, execAddress, price); err != nil {
+			return nil, err
+		}
 	}
 
-	if !ctx.IsCheckTx() {
-		M.InProgress.Add(1)
+	var count int64
+	if exec.Confs > 0 && execExist {
```
antho1404: don't understand why we only add the signer if the execution already exists. What happens to the first signer who actually paid more than the other because he wrote more data. This one is not added to the list and will not be rewarded for leading this execution.

krhubert:
The point of skipping the confs on hash calculation is to disable modification of this value aftre exec creation. So the two workflow won't happen, because you can't edit confs.

-------

The execution is created by the first signer na then the confs is upadted. The exec creator is not treated as a voter because we need a confirmation from others in network. 

-------

I need to know,

who and when pays for exeuction (in case of process/runner/user) 

who and when gets reward after exeuction completed/failed (in case of process/runner/user)

So the execution can be created by process (orchestrator), inside runner, or just with cli (as I see it)

NicolasMahe:
closing in favor of https://github.com/mesg-foundation/engine/pull/1705

PR #1704, opened 2020-03-05
Use type accAddress instead of hash.Hash for resources' hash

NicolasMahe:
POC of replacing resource's `hash.Hash` by `AccAddress`.


event/serialize_test.go
```diff
@@ -0,0 +1,33 @@
...
+			"foo": {
+				Kind: &types.Value_StringValue{StringValue: "bar"},
+			},
+		},
+	},
+}
+
+func TestHashSerialize(t *testing.T) {
+	require.Equal(t, "2:cosmos1ffzdc9fkggz2srlgp6grj32uc9sg9qvzu4jym8;3:eventKey;4:1:foo:3:bar;;;;", data.HashSerialize())
+	require.Equal(t, "cosmos186j7ddz0046caf63h566ez5nxd23svd44ga94h", sdk.AccAddress(crypto.AddressHash([]byte(data.HashSerialize()))).String())
```
NicolasMahe: by default, the bech32 are using `cosmos` prefix.
we override it with:
```
cfg, err := config.New()
	if err != nil {
		panic(err)
	}
	cosmos.CustomizeConfig(cfg)
```
I would like to extract the prefix from the config and hardcode them somewhere so we could override the cosmos's prefix in an `init` function.

runner/builder/builder.go
```diff
@@ -45,20 +47,22 @@ func (b *Builder) Create(req *api.CreateRunnerRequest) (*runnerpb.Runner, error)
 	}
 
 	instanceEnv := xos.EnvMergeSlices(srv.Configuration.Env, req.Env)
-	envHash := hash.Dump(hashserializer.StringSlice(instanceEnv))
+	envHash := hash.Sum([]byte(strings.Join(instanceEnv, ",")))
```
antho1404: why not doing the same here too ;)
this way we only have one hashing function, same that events that are not needed as a cosmos address but not bad representation anyway

runner/builder/container.go
```diff
@@ -268,9 +268,9 @@ func convertVolumesFrom(s *service.Service, dVolumesFrom []string) ([]container.
 // volumeKey creates a key for service's volume based on the sid to make sure that the volume
 // will stay the same for different versions of the service.
 func volumeKey(s *service.Service, dependency, volume string) string {
-	return hash.Dump(hashserializer.StringSlice([]string{
+	return hash.Sum([]byte(strings.Join([]string{
```
antho1404: here too, cosmos address can be use

x/execution/internal/keeper/keeper.go
```diff
@@ -175,10 +172,7 @@ func (k *Keeper) Update(ctx sdk.Context, msg types.MsgUpdateExecution) (*executi
 		return nil, err
 	}
 
-	execAddress := sdk.AccAddress(crypto.AddressHash(exec.Hash))
-	runnerAddress := sdk.AccAddress(crypto.AddressHash(run.Hash))
-	serviceAddress := sdk.AccAddress(crypto.AddressHash(srv.Hash))
-	if err := k.distributePriceShares(ctx, execAddress, runnerAddress, serviceAddress, exec.Price); err != nil {
+	if err := k.distributePriceShares(ctx, exec.Hash, run.Hash, srv.Hash, exec.Price); err != nil {
```
antho1404: Love how simple it is now 😍

NicolasMahe:
closing but discussion continue here: https://github.com/mesg-foundation/engine/issues/1715

PR #1706, opened 2020-03-05
 Create a proposal for a execution emitters coins split

krhubert:
dep https://github.com/mesg-foundation/engine/pull/1703


x/execution/internal/keeper/keeper.go
```diff
@@ -19,7 +19,10 @@ import (
 )
 
 // price share for the execution runner
-var runnerShare = sdk.NewDecWithPrec(9, 1)
+var (
+	runnerShare   = sdk.NewDecWithPrec(8, 1)
+	emittersShare = sdk.NewDecWithPrec(1, 1)
```
antho1404: just for comprehension, it would be nice to also have the `serviceShare` and assert in the code if the rest of `runner + emitters != serviceShare`. Like that, we will have 100%

x/execution/internal/keeper/keeper.go
```diff
@@ -221,14 +225,30 @@ func (k *Keeper) distributePriceShares(ctx sdk.Context, execAddress, runnerAddre
...
 
 	runnerCoins, _ := sdk.NewDecCoinsFromCoins(coins...).MulDecTruncate(runnerShare).TruncateDecimal()
-	serviceCoins := coins.Sub(runnerCoins)
-	// TODO: add emitters share
+	emittersCoins, _ := sdk.NewDecCoinsFromCoins(coins...).MulDecTruncate(emittersShare).TruncateDecimal()
+	serviceCoins := coins.Sub(runnerCoins).Sub(emittersCoins)
+
+	var voters []sdk.AccAddress
+	var emitterCoins sdk.Coins
+	if store.Has(voteKey(execHash)) {
```
antho1404: We shouldn't do this test, if there is no vote on the store it means we have a serious issue before and we should either panic or at least return an error and not just ignore it I think. Or we could calculate everything and at the end assert that the result of emitters + runner + service == balance of execution.

x/execution/internal/keeper/keeper.go
```diff
@@ -221,14 +225,30 @@ func (k *Keeper) distributePriceShares(ctx sdk.Context, execAddress, runnerAddre
...
+		k.cdc.MustUnmarshalBinaryLengthPrefixed(store.Get(voteKey(execHash)), &voters)
+		// divide coins for each emitter
+		emitterCoins, _ = sdk.NewDecCoinsFromCoins(emittersCoins...).QuoDec(sdk.NewDec(int64(len(voters)))).TruncateDecimal()
+	}
+
+	// if there is no emitter get rest coins to runner
+	if emitterCoins.IsZero() {
```
antho1404: why should we do that? why would we have no emitters in the first place? and why giving to the runner and not the service or get back to the user etc... we never talked about that...

x/execution/internal/keeper/keeper.go
krhubert: Yes and no. We need to be sure that inputs == outputs and we can only have that, if we subtract it, because division can't guarantee that.

x/execution/internal/keeper/keeper.go
krhubert: This is related to a question if we should put a exec creator as an voter into the list. If yes, then this check shouldn't be here, otherwise it's should be :)

x/execution/internal/keeper/keeper.go
krhubert: > why would we have no emitters in the first place?

See above. Also, it might happen that coins/len(emitters) == 0, so we can either pay only a few random emitters or none at all.

> and why giving to the runner

Because I took the reward from runners' share.

x/execution/internal/keeper/keeper.go
antho1404: of course but we can still check that the diff matches the supposed value of the service. Anyway not blocking but it's just really confusing not to have all the repartition here, it looks like there is only 90% of the distribution and the rest maybe stays in the execution. (If we have a deep look at the code then we see it's not the case but we should not need to have a deep look at the code)

x/execution/internal/keeper/keeper.go
antho1404: > Because I took the reward from runners' share.

No, you're giving to the runner from the execution here... or I think there is a big confusion with what is the runner and all the different types of runners (emitters/executors).
The different types of runners will do a specific job and will be rewarded for that and not for the job of others.

Also, I'm seeing that in fact, this instruction is only when the coins are zero and you add zero to the existing coins. So yes we will actually not pay anyone and this whole instruction is useless isn't it?

NicolasMahe:
closing this PR in favor of https://github.com/mesg-foundation/engine/pull/1705

PR #1710, opened 2020-03-06
Replace tag hash by a comment in proto files

NicolasMahe:
Replace in proto files the hash in `moretags` by a comment starting with `@hash`:
```diff
+  // @hash: false
  bytes hash = 1 [
-    (gogoproto.moretags) = 'hash:"-"',

+ // @hash: true
  bytes instanceHash = 2 [
-    (gogoproto.moretags) = 'hash:"name:2" validate:"required,hash"',
+    (gogoproto.moretags) = 'validate:"required,hash"',
```

@mesg-foundation/mesg-core-team what do you think about the `@hash` comment? do you have a better suggestion to tell if the field should be used for hash calculation?


antho1404:
I like the idea to move it as documentation (and yes it's not needed as a tag in proto anymore).
I don't really like the format `@hash` and also this documentation is not related to the code. We could change the hash and forget to change the proto.
I would put it as documentation of each `hashserializer` functions and not directly in the proto.

NicolasMahe:
For me, the proto files must be the only source of truth about the data of the whole system. So it is the place the tell which field to use for hash.
The field's number is also used for the hash calculation. So it's simpler to have both info in the same file.
Also, I prefer that JS developers read the protofiles instead of `.go` files.

The generated doc website (docs.mesg.com) should contain a page containing the docs of protofiles and make it accessible from any developer in any language (the goal of proto).

antho1404:
This should be close because of https://github.com/mesg-foundation/engine/pull/1717

NicolasMahe:
closing because of https://github.com/mesg-foundation/engine/pull/1717

PR #1712, opened 2020-03-06
execution result consensus

krhubert:
Just a draft for result consensus.

dep on 
- https://github.com/mesg-foundation/engine/pull/1706
- https://github.com/mesg-foundation/engine/pull/1703


antho1404:
This PR was not planned and also definitely not in was we are supposed to do...
There is no consensus on the execution, only one executor will be selected (maybe multiple in case the previous doesn't respond but this is just a fallback).
We will have multiple verifications of this single result but that's not the execution part.

Also, even if this was the direction, this PR would be wrong because we have a consensus on only one selected executor in the execution.

NicolasMahe:
closing this PR as it's not the direction of current development

PR #1729, opened 2020-03-16
Wrap errors of modules' keeper, querier and msg to sent them to clients

NicolasMahe:
closes https://github.com/mesg-foundation/engine/issues/1711

Wrap errors of modules' keeper, querier and msg with the `sdkerrors.Wrap` function.
This allows the error's message to be sent to clients.
Before:
```
➜  engine git:(feature/wrap-errors) ✗ curl localhost:1317/service/get/QY8K2bPZMx3UxbU4ykVGBM6JYKf2B4ztgGP2igQG9Sb
{"error":"internal error"}%
```
After:
```
➜  engine git:(feature/wrap-errors) curl localhost:1317/service/get/QY8K2bPZMx3UxbU4ykVGBM6JYKf2B4ztgGP2igQG9Sb
{"error":"invalid request: service \"QY8K2bPZMx3UxbU4ykVGBM6JYKf2B4ztgGP2igQG9Sb\" not found"}%
```


x/execution/handler.go
```diff
@@ -30,7 +26,7 @@ func NewHandler(k Keeper) sdk.Handler {
 func handleMsgCreateExecution(ctx sdk.Context, k Keeper, msg MsgCreateExecution) (*sdk.Result, error) {
 	exec, err := k.Create(ctx, msg)
 	if err != nil {
-		return nil, sdkerrors.Wrap(errors.ErrUnknown, err.Error())
+		return nil, err
```
antho1404: why removing them here?

x/ownership/internal/types/msg.go
```diff
@@ -35,13 +35,13 @@ func (msg MsgWithdrawCoins) Type() string {
 // ValidateBasic runs stateless checks on the message.
 func (msg MsgWithdrawCoins) ValidateBasic() error {
 	if msg.Amount.IsAnyNegative() {
-		return sdkerrors.Wrap(sdkerrors.ErrInsufficientFunds, "price must be positive")
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "price must be positive")
 	}
 	if msg.Owner.Empty() {
-		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, msg.Owner.String())
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, msg.Owner.String())
```
antho1404: Why changing the invalid address error? It seems that this error give less context

x/execution/handler.go
NicolasMahe: before the error is return from a keeper function that should always already returned a wrap error.

x/ownership/internal/types/msg.go
NicolasMahe: For simplicity I changed most of the error to `ErrInvalidRequest` but maybe that was too much.
I will take a look to see which error can be used to give more context.

NicolasMahe:
TODO:
- [ ] use better error
- [ ] create error per module like `github.com/cosmos/cosmos-sdk/x/gov/types/msgs.go` or create app-wide errors?

x/execution/handler.go
antho1404: It's a bit weird to have some errors that are returned as error that don't need wrapping and some that need. Would love to find a way to fix that, maybe a specific check might be good or a specific type.

Anyway, this is not blocking and can be improved later

NicolasMahe: I checked to see if I could check if an error is already wrapped but couldn't find a nice way.

x/ownership/internal/types/msg.go
NicolasMahe: I will improve the error on msg in a next PR when I improve the msg fields ;)

NicolasMahe:
@antho1404 ready to review

x/execution/internal/keeper/keeper.go
```diff
@@ -337,9 +340,6 @@ func (k *Keeper) distributePriceShares(ctx sdk.Context, execAddress, runnerAddre
 		if err != nil {
 			return err
 		}
-		if err != nil {
```
antho1404: 🤣 good to make sure that we return the error

PR #1738, opened 2020-03-17
Add more predicates to process filter

NicolasMahe:
dependant on https://github.com/mesg-foundation/engine/pull/1737
related to https://github.com/mesg-foundation/engine/issues/1734

- Add predicates `GT`, `GTE`, `LT`, `LTE` to type `Number`
  - Check in the process validation if the value of the filter is a `Number`
- Add predicate `CONTAINS` to type `List`
  - Cannot add process validation because the task's output or event's data should be of type `List`, not the value in the filter
  - Needs reference system in the filter to implement an E2E with the current e2e service


process/process.go
```diff
@@ -33,15 +34,27 @@ func (w *Process) Validate() error {
...
 				if ref := output.GetRef(); ref != nil {
 					if _, err := w.FindNode(ref.NodeKey); err != nil {
 						return err
 					}
 				}
 			}
+		case *Process_Node_Filter_:
+			for _, condition := range n.Filter.Conditions {
+				switch condition.Predicate {
+				case Process_Node_Filter_Condition_GT,
```
antho1404: [not blocking] would be nice to also add the check for the contain

process/filter.go
```diff
@@ -9,12 +9,44 @@ func (f Process_Node_Filter) Match(data *types.Struct) bool {
...
+		case Process_Node_Filter_Condition_GT:
+			return n1.NumberValue > n2.NumberValue
+		case Process_Node_Filter_Condition_GTE:
+			return n1.NumberValue >= n2.NumberValue
+		case Process_Node_Filter_Condition_LT:
+			return n1.NumberValue < n2.NumberValue
+		case Process_Node_Filter_Condition_LTE:
+			return n1.NumberValue <= n2.NumberValue
+		}
+	case Process_Node_Filter_Condition_CONTAINS:
```
antho1404: [not blocking] contains is also often supported for strings

process/process.go
NicolasMahe: I cannot add a check for contains as I wrote in the PR's description:
> Cannot add process validation because the task's output or event's data should be of type List, not the value in the filter

Here, I don't have access to the data.

NicolasMahe: done

process/filter.go
NicolasMahe: done

PR #1739, opened 2020-03-20
Add reference system to filter

NicolasMahe:
Closes https://github.com/mesg-foundation/engine/issues/1582

- Move the reference proto message to process
- Add `dive` validation to all repeated messages in proto, otherwise, the validator doesn't go in the child elements
- Use the reference system in processes' filter


orchestrator/orchestrator.go
```diff
@@ -161,6 +161,21 @@ func (s *Orchestrator) executeNode(wf *process.Process, n *process.Process_Node,
 	return nil
 }
 
+// filterMatch returns true if the data match the current list of filters.
+func (s *Orchestrator) filterMatch(f *process.Process_Node_Filter, wf *process.Process, nodeKey string, exec *execution.Execution, data *types.Struct) bool {
+	for _, condition := range f.Conditions {
+		resolvedData, err := s.resolveRef(wf, exec, nodeKey, data, condition.Ref)
+		if err != nil {
+			return false
```
antho1404: We should return this error to the user
```suggestion
                         s.ErrC <- err
			return false
```

PR #1742, opened 2020-03-20
Simpler messages using protobuf definiton

NicolasMahe:
related to https://github.com/mesg-foundation/engine/issues/1741

- Simplifies and improve xvalidator so I can use it to validate msgs
- Simplifies cosmos message to create a service
- Add more validation to service. Move service validation functions to the service package.
- Simplifies cosmos message create and delete process
- Simplifies cosmos message create and delete runner
- Simplifies cosmos message create and update execution
- Simplifies cosmos message withdraw coins of ownership module


antho1404:
One note, we could also have the messages renamed.
For example, currently, the messages generated are for example `service/CreateService`(https://github.com/mesg-foundation/js-sdk/blob/master/packages/api/src/service-lcd.ts#L27)
I would go with something simple like just `service/Create`. I think just changing the type of the message should be enough.

Some other stuff that can/should be done just as a reminder
- messages in proto
- flatten the messages, no "request" inside and if we can have everything at the root level that's ideal

x/ownership/internal/types/msg.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+
+import "gogo/protobuf/gogoproto/gogo.proto";
+
+package mesg.ownership.types;
+option go_package = "github.com/mesg-foundation/engine/x/ownership/internal/types";
+
+// The message to withdraw coins from an ownership.
+message MsgWithdrawCoins {
```
antho1404: Should we just call it `MsgWithdraw`, we will obviously withdraw coins

PR #1757, opened 2020-03-27
Replace runner gRPC by LCD in e2e tests

NicolasMahe:
Dependent on https://github.com/mesg-foundation/engine/pull/1751
Related to https://github.com/mesg-foundation/engine/issues/1741

- fix bug runner/builder/container when current UID and GID are not 0
- Introduce a new structure `runner/builderLCD` that create and stop docker services and returns the msg to broadcast.
- Use it in e2e tests.
- Remove gRPC RunnerClient from e2e test.

### DIFF
--- a/e2e/runner_test.go
+++ b/e2e/runner_test.go
@@ -5,19 +5,51 @@ import (
 	"testing"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/mesg-foundation/engine/ext/xos"
 	"github.com/mesg-foundation/engine/hash"
 	"github.com/mesg-foundation/engine/protobuf/acknowledgement"
 	pb "github.com/mesg-foundation/engine/protobuf/api"
 	"github.com/mesg-foundation/engine/runner"
+	"github.com/mesg-foundation/engine/runner/builder"
+	runnermodule "github.com/mesg-foundation/engine/x/runner"
 	runnerrest "github.com/mesg-foundation/engine/x/runner/client/rest"
 	"github.com/stretchr/testify/require"
 )
 
-var testRunnerHash hash.Hash
-var testRunnerAddress sdk.AccAddress
+var (
+	testRunnerHash       hash.Hash
+	testInstanceEnvHash  hash.Hash
+	testRunnerAddress    sdk.AccAddress
+	testServiceImageHash string
+)
 
 func testRunner(t *testing.T) {
-	t.Run("create", func(t *testing.T) {
+	var (
+		testInstanceEnv = xos.EnvMergeSlices(testServiceStruct.Configuration.Env, []string{"BAR=3", "REQUIRED=4"})
+	)
+	t.Run("hash", func(t *testing.T) {
+		var res runnerrest.HashResponse
+		lcdPost("runner/hash", &runnerrest.HashRequest{
+			ServiceHash: testServiceHash,
+			Address:     engineAddress.String(),
+			Env:         testInstanceEnv,
+		}, &res)
+		testRunnerHash = res.RunnerHash
+		testInstanceHash = res.InstanceHash
+		testInstanceEnvHash = res.EnvHash
+	})
+
+	t.Run("build service image", func(t *testing.T) {
+		var err error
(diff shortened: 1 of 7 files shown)


e2e/runner_test.go
```diff
@@ -26,60 +28,38 @@ func testRunner(t *testing.T) {
 		require.NoError(t, err)
 		acknowledgement.WaitForStreamToBeReady(stream)
 
-		resp, err := client.RunnerClient.Create(context.Background(), &pb.CreateRunnerRequest{
-			ServiceHash: testServiceHash,
-			Env:         []string{"BAR=3", "REQUIRED=4"},
-		})
+		msg, err := runBuilder.Start(engineAddress, testServiceStruct, []string{"BAR=3", "REQUIRED=4"})
```
antho1404: I would start the service and then when started, create the message directly in the test with the result of the start (if needed) this way the start is not related to a message just to its function of starting a runner.

e2e/runner_test.go
```diff
@@ -26,60 +28,38 @@ func testRunner(t *testing.T) {
...
+		testRunnerHash = lcdBroadcastMsg(msg)
 
 		// wait for service to be ready
 		_, err = stream.Recv()
 		require.NoError(t, err)
 	})
 
 	t.Run("recreate", func(t *testing.T) {
-		_, err := client.RunnerClient.Delete(context.Background(), &pb.DeleteRunnerRequest{Hash: testRunnerHash})
+		msgDelete, err := runBuilder.Stop(engineAddress, testRunnerHash, testServiceStruct, true)
```
antho1404: same here separating the stoping function to the actual message

runner/builder/builder_lcd.go
```diff
@@ -0,0 +1,87 @@
...
+
+	inst, err := instance.New(srv.Hash, envHash)
+	if err != nil {
+		return runnermodule.MsgCreate{}, err
+	}
+	expRunner, err := runner.New(owner.String(), inst.Hash)
+	if err != nil {
+		return runnermodule.MsgCreate{}, err
+	}
+	expRunnerHash := expRunner.Hash
```
antho1404: All this could be done by calling the querier `/runner/hash`

runner/builder/builder_lcd.go
```diff
@@ -0,0 +1,87 @@
...
+	return &LCD{
+		container:    container,
+		engineName:   engineName,
+		port:         port,
+		ipfsEndpoint: ipfsEndpoint,
+	}
+}
+
+// Start start a new runner.
+func (b *LCD) Start(owner sdk.AccAddress, srv *service.Service, env []string) (runnermodule.MsgCreate, error) {
```
antho1404: This builder package is not related to the LCD so it shouldn't return data for the LCD

e2e/runner_test.go
antho1404: You could even use the querier `/runner/hash` and you would have all the information to call the functions
- `build(cont container.Container, srv *service.Service, ipfsEndpoint string)`
- `start(cont container.Container, srv *service.Service, instanceHash hash.Hash, runnerHash hash.Hash, imageHash string, env []string, engineName, port string)`

That we can put public. This way the whole LCD struct in the builder package can be removed

runner/builder/builder_lcd.go
NicolasMahe: I thought the hash endpoints were just temporary?
But yeah, if so, it will shrink this function as much as nothing: the e2e could directly call the private functions of this package (that should be converted to public ones)

runner/builder/builder_lcd.go
NicolasMahe: with the other modification done, I will remove it

e2e/runner_test.go
NicolasMahe: done

e2e/runner_test.go
NicolasMahe: done

runner/builder/builder_lcd.go
NicolasMahe: done

runner/builder/builder_lcd.go
NicolasMahe: done

PR #1758, opened 2020-03-27
Introduce a simple LCD client

NicolasMahe:
Dependent on https://github.com/mesg-foundation/engine/pull/1757

- introduce a simple LCD client
- use it in the e2e tests


cosmos/lcd.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	sequence := accR.GetSequence()
+	accR.SetSequence(accR.GetSequence() + 1)
+
+	// Create TxBuilder
+	txBuilder := authtypes.NewTxBuilder(
+		authutils.GetTxEncoder(lcd.cdc),
+		accR.GetAccountNumber(),
+		sequence,
+		flags.DefaultGasLimit*10,
```
antho1404: why `*10`?

NicolasMahe: because of the transaction that contains 10 executions. the default was not enough

NicolasMahe: it will be removed when estimate gas is implemented

antho1404: if we have this generic client we should then pass this as an argument to be able to customize it.
Can have multiple ways:
- as a parameter of this function and so all the broadcast functions
- as an `option` function on the LCD that will write it int the attributes of the client `WithGazLimit()` that will be default `flags.DefaultGasLimit`
- as a parameter of the constructor like it's done for the `minGasPrices`

I'm fine with any of these solutions but hardcoding a `*10` is a bit sketchy for a generic client

NicolasMahe:
> If we do another client in the same package `cosmos` I think we should as least rename the current client to something less generic than `Client` and maybe `ABCI`. Right now we have `Client` and `LCD`.

renamed to `RPC`

cosmos/lcd.go
NicolasMahe: I add a parameter `gasLimit` to the LCD

PR #1762, opened 2020-03-28
Fix account sequence desynchronisation when error

NicolasMahe:
Fixes https://github.com/mesg-foundation/engine/issues/1693

This PR make the lcd and rpc clients to only increase the account sequence when the transaction is ok.


cosmos/rpc.go
```diff
@@ -91,23 +91,43 @@ func (c *RPC) QueryWithData(path string, data []byte) ([]byte, int64, error) {
...
 	if err != nil {
-		c.broadcastMutex.Unlock()
 		return nil, err
 	}
+	return c.waitForTxResult(signedTx)
+}
 
-	txres, err := c.BroadcastTxSync(signedTx)
-	c.broadcastMutex.Unlock()
+func (c *RPC) buildAndBroadcastMsgNoResult(msg sdktypes.Msg) (tenderminttypes.Tx, error) {
```
NicolasMahe: I split this function into 2 to be able to use a defer on the mutex unlock instead of putting it in errors.

antho1404: you could as well explicitly unlock the mutex without a defer but I'm good with that

antho1404:
Seems valid, could you run the perf tests and check the result on the grafana to see the impact of such a modification, the before/after visualization for let's say 1000 executions would be great

NicolasMahe:
> Seems valid, could you run the perf tests and check the result on the grafana to see the impact of such a modification, the before/after visualization for let's say 1000 executions would be great

Same speed on the e2e tests with 10 executions:
Before:
```
        --- PASS: TestAPI/execution/many_executions_in_parallel (1.65s)
            --- PASS: TestAPI/execution/many_executions_in_parallel/create_executions (0.60s)
            --- PASS: TestAPI/execution/many_executions_in_parallel/check_in_progress (0.02s)
            --- PASS: TestAPI/execution/many_executions_in_parallel/check_completed (1.03s)
```
After:
```
        --- PASS: TestAPI/execution/many_executions_in_parallel (1.59s)
            --- PASS: TestAPI/execution/many_executions_in_parallel/create_executions (0.56s)
            --- PASS: TestAPI/execution/many_executions_in_parallel/check_in_progress (0.02s)
            --- PASS: TestAPI/execution/many_executions_in_parallel/check_completed (1.02s)
```

PR #1765, opened 2020-03-31
Update go-playground/validator to v10

NicolasMahe:
Update go-playground/validator to v10, so I can use new validator functions


NicolasMahe:
closing because directly integrated into https://github.com/mesg-foundation/engine/pull/1767

PR #1767, opened 2020-04-01
New Runner gRPC API

NicolasMahe:
Related to https://github.com/mesg-foundation/engine/issues/1764
Dependent on https://github.com/mesg-foundation/engine/pull/1768 for the exists endpoint.

This PR adds the new Runner gRPC API that service must use.
The big difference between this API and the previous ones is this API is secure (requires authentication) and specifically design for Runners, so it is much easier to integrate.

The API is composed of 4 endpoints:

#### Register

The service must call this API first but only when it's ready to emit events and/or execute executions.
The service has to call this endpoint with the content of the env variable `MESG_REGISTER_PAYLOAD` (injected by the CLI) to get a credential token requires for the following endpoints.

Go example:
```go
resp, _ := client.Register(context.Background(), &runner.RegisterRequest{
	Payload: os.Getenv("MESG_REGISTER_PAYLOAD"),
})
fmt.Println(resp.Token)
```

Once you got the token, the service need to pass it to the next endpoints in the gRPC metadata.
There are three ways to do this:
- You can use the gRPC helper `WithPerRPCCredentials` on the grpc `DialOption` of the client
- You can use the gRPC helper `PerRPCCredentials` on the `CallOption` of the request
- You can create the metadata object yourself and using the key `mesg_credential_token`

The first two options requires a struct that implement the interface [`credentials.PerRPCCredentials`](https://godoc.org/google.golang.org/grpc/credentials#PerRPCCredentials):

```go
// credential is a structure that manage a token.
type credential struct {
	token string
}

// GetRequestMetadata returns the metadata for the request.
func (c *credential) GetRequestMetadata(context.Context, ...string) (map[string]string, error) {
	return map[string]string{
		runner.CredentialToken: c.token,
	}, nil
}

// RequireTransportSecurity tells if the transport should be secured.
func (c *credential) RequireTransportSecurity() bool {
	return false
}
```

Credential in client:
```go
cred := &credential{
	token: resp.Token,
}
dialoptions := []grpc.DialOption{
	grpc.WithInsecure(),
	grpc.WithPerRPCCredentials(credential),
}
conn, _ := grpc.DialContext(context.Background(), os.Getenv("MESG_ENDPOINT"), dialoptions...)
client := runner.NewRunnerClient(conn)
```

Credential per request:
```go
cred := &credential{
	token: token,
}
client.XXX(context.Background(), &runner.XXX{}, grpc.PerRPCCredentials(cred))
```

#### Event
This endpoint is used to emit events.
It accepts an event's key and data.
The credential token must be passed to this request.

Go example:
```go
client.Event(context.Background(), &runner.EventRequest{
	Key: "test_service_ready",
	Data: &types.Struct{
		Fields: map[string]*types.Value{
			"msg": {
				Kind: &types.Value_StringValue{
					StringValue: "foo",
				},
			},
		},
	},
})
```

#### Execution
This endpoint is used to create a stream that will receive the execution the runner has to execute.
The credential token must be passed to this request.

Go example:
```go
stream, _ := client.Execution(context.Background(), &runner.ExecutionRequest{})
for {
	exec, _ := stream.Recv()
	log.Printf("received execution %s %s\n", exec.TaskKey, exec.Hash)
}
```

#### Result
This endpoint is used to return the execution's result once the upcoming execution from the execution stream has been executed.
It accepts the execution's hash and the result.
The credential token must be passed to this request.

Go example:
```go
client.Result(context.Background(), &runner.ResultRequest{
	ExecutionHash: exec.Hash,
	Result: &runner.ResultRequest_Outputs{
		Outputs: &types.Struct{
			Fields: map[string]*types.Value{
				"msg": {
					Kind: &types.Value_StringValue{
						StringValue: "foo",
					},
				},
			},
		},
	},
})
```

### CLI

The CLI has to create and inject the env variable `MESG_REGISTER_PAYLOAD` that the service will use to register itself against the Engine.

This variable contains the data needed for the engine to register the runner as well as a signature that the engine verifies to check the authenticity of the runner by making sure the runner is created by the CLI.

The protobuf definition of this variable is:
```protobuf
// RegisterRequest is the request of the endpoint Register.
message RegisterRequestPayload {
  // signature of the value to verify authenticity.
  bytes signature = 1 [
    (gogoproto.moretags) = 'validate:"required"'
  ];

  // value contains what the engine actually needs to register the runner.
  Value value = 2 [
    (gogoproto.moretags) = 'validate:"required"',
    (gogoproto.nullable) = false
  ];

  // Value message
  message Value {
    // Service's hash to start the runner with.
    bytes serviceHash = 1 [
      (gogoproto.moretags) = 'validate:"required,hash"',
      (gogoproto.casttype) = "github.com/mesg-foundation/engine/hash.Hash"
    ];
  
    // Hash of the environmental variables to start the runner with.
    bytes envHash = 2 [
      (gogoproto.moretags) = 'validate:"omitempty,hash"',
      (gogoproto.casttype) = "github.com/mesg-foundation/engine/hash.Hash"
    ];
  }
}
```

The message `Value` is very close to the runner module MsgCreate. It contains the minimum of data so the engine can actually create, sign, and broadcast MsgCreate to register the runner on the network.

`signature` is the signature of the message `Value` (amino-json encoded) using the engine's account.

Finally, the whole `RegisterRequestPayload` message is also amino-json encoded to be injected in the env of the runner.

Here is the implementation of this system in the e2e runner test:

```go
value := grpcrunner.RegisterRequestPayload_Value{
	ServiceHash: serviceHash,
	EnvHash:     instanceEnvHash,
}

encodedValue, _ := cdc.MarshalJSON(value)

signature, _, _ := kb.Sign(engineAccountName, engineAccountPassword, encodedValue)

payload := grpcrunner.RegisterRequestPayload{
	Signature: signature,
	Value:     value,
}
encodedPayload, _ := cdc.MarshalJSON(payload)

instanceEnv = append(instanceEnv, "MESG_REGISTER_PAYLOAD="+string(encodedPayload))
```
With `MESG_REGISTER_PAYLOAD` containing (without identation):
```json
{
	"signature": "wpSfARExAzv5c4PivU7K++XpX1DMyoLaCmzCx7bYkwMJLfBSDbZMOngaG7tMrN+sogG2UuXkPcKDQMuhIxAQZA==",
	"value": {
		"serviceHash": "BoPaG69vP2ZoexqxFXkq94ZVfn9768GB8YCiYAdNWPju",
		"envHash": "J3PJvcXq3CXKEeP8jraYGUQSAj8g7cNxMNsMQgLqfJMW"
	}
}
```


e2e/testdata/test-complex-service/main.go
```diff
@@ -10,38 +10,23 @@ import (
...
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/keepalive"
 )
 
 const (
 	// env variables for configure mesg client.
-	envMesgEndpoint     = "MESG_ENDPOINT"
-	envMesgInstanceHash = "MESG_INSTANCE_HASH"
+	envMesgEndpoint  = "MESG_ENDPOINT"
+	envMesgMsg       = "MESG_MSG"
```
NicolasMahe, line 21: what do you think to rename this envs to `MESG_REGISTER_MSG` and `MESG_REGISTER_SIGNATURE`?

e2e/testdata/test-complex-service/main.go
```diff
@@ -10,38 +10,23 @@ import (
...
+	envMesgSignature = "MESG_SIGNATURE"
```
NicolasMahe, line 22: what do you think to rename this envs to `MESG_REGISTER_MSG` and `MESG_REGISTER_SIGNATURE`?

e2e/testdata/test-complex-service/main.go
```diff
@@ -59,68 +44,89 @@ func New() (*Client, error) {
...
+// sendEvent creates a new event.
+func sendEvent(client runner.RunnerClient, token string, key string) {
 	log.Println("sending event:", key)
-	if _, err := c.EventClient.Create(context.Background(), &pb.CreateEventRequest{
-		InstanceHash: c.InstanceHash,
-		Key:          key,
-	}); err != nil {
+	if _, err := client.Event(context.Background(), &runner.EventRequest{
+		Key: key,
+	}, grpc.PerRPCCredentials(runner.NewTokenCredential(token))); err != nil {
```
NicolasMahe, line 75: this is the way to send the credential to a specific request.
There is a way to send the credential to every request of the client, but as Register doesn't need it and the token is only obtained after calling register, it will be a weird design.

server/grpc/runner/runner.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return nil, err
+	}
+	runnerHash := run.Hash
+
+	// check that runner doesn't already exist
+	runnerExist := true
+	route := fmt.Sprintf("custom/%s/%s/%s", runnermodule.QuerierRoute, runnermodule.QueryGet, runnerHash)
+	if _, _, err := s.rpc.QueryWithData(route, nil); err != nil {
+		if strings.Contains(err.Error(), fmt.Sprintf("runner %q not found", runnerHash)) {
```
NicolasMahe, line 79: i don't like to catch an error but I couldn't find a better way with the current system
maybe it will be nicer with a dedicated error instead of using the error message?

server/grpc/runner/runner.go
```diff
@@ -0,0 +1,284 @@
...
+func (c *TokenCredential) GetRequestMetadata(context.Context, ...string) (map[string]string, error) {
+	return map[string]string{
+		"token": c.token,
+	}, nil
+}
+
+// RequireTransportSecurity tells if the transport should be secured.
+func (c *TokenCredential) RequireTransportSecurity() bool {
+	return false
+	// TODO: test with true
```
NicolasMahe, line 268: to test

server/grpc/runner/runner.proto
```diff
@@ -0,0 +1,86 @@
...
+// RegisterRequest is the request of the endpoint Register.
+message RegisterRequest {
+  // msg to use to register this runner. it should be a JSON encoded runner module MsgCreate
+  string msg = 1 [
+    (gogoproto.moretags) = 'validate:"required,json"'
+  ];
+  // TODO: the following doesn't work because the msg is declared in an internal package.
+  // mesg.runner.types.MsgCreate msg = 1 [
+  //   (gogoproto.customtype) = "github.com/mesg-foundation/engine/x/runner.MsgCreate"
+  // ];
```
NicolasMahe, line 35: to do or remove

e2e/testdata/test-complex-service/main.go
NicolasMahe: the 2 envs has been replaced by `MESG_REGISTER_PAYLOAD`

server/grpc/runner/runner.go
NicolasMahe: it requires to activate TLS!

server/grpc/runner/runner.go
antho1404: or just an endpoint "exists" maybe

server/grpc/runner/runner.go
```diff
@@ -0,0 +1,283 @@
...
+	runnermodule "github.com/mesg-foundation/engine/x/runner"
+	"google.golang.org/grpc/metadata"
+)
+
+// Server is the type to aggregate all Runner APIs.
+type Server struct {
+	rpc *cosmos.RPC
+	ep  *publisher.EventPublisher
+
+	tokenToRunnerHash sync.Map
```
antho1404, line 26: Didn't know that one 👍

server/grpc/runner/runner.go
```diff
@@ -0,0 +1,283 @@
...
+	for {
+		select {
+		case execHash := <-execChan:
+			var exec *execution.Execution
+			route := fmt.Sprintf("custom/%s/%s/%s", executionmodule.QuerierRoute, executionmodule.QueryGet, execHash)
+			if err := s.rpc.QueryJSON(route, nil, &exec); err != nil {
+				return err
+			}
+			// filter execution of this runner
+			if exec.ExecutorHash.Equal(runnerHash) && exec.Status == execution.Status_InProgress {
```
antho1404, line 161: 👍

server/grpc/runner/runner.go
```diff
@@ -0,0 +1,283 @@
...
+func (c *TokenCredential) GetRequestMetadata(context.Context, ...string) (map[string]string, error) {
+	return map[string]string{
+		"token": c.token,
+	}, nil
+}
+
+// RequireTransportSecurity tells if the transport should be secured.
+func (c *TokenCredential) RequireTransportSecurity() bool {
+	return false
+}
```
antho1404, line 268: Not sure this should be here, it is not used by the engine, I would just duplicate it in the e2e test until we have a library the does it for us.

server/grpc/runner/runner.go
NicolasMahe: done in https://github.com/mesg-foundation/engine/pull/1768 and rebased current PR

server/grpc/runner/runner.go
NicolasMahe: I moved this to the services

PR #1771, opened 2020-04-03
Add Orchestrator gRPC API

NicolasMahe:
Related to https://github.com/mesg-foundation/engine/issues/1764

This PR adds the **Orchestrator** API.

This API requires authentication by verifying a base64-encoded signature of the request, set in the request's metadata using the key `mesg_request_signature`.

It defines the following RPC functions:
```protobuf
service Event {
  // Stream returns a stream of events that satisfy criteria specified.
  rpc Stream(EventStreamRequest) returns (stream types.Event) {}
}

service Execution {
  // Create an execution on the blockchain.
  rpc Create(ExecutionCreateRequest) returns (ExecutionCreateResponse) {}

  // Stream returns a stream of executions that satisfy specified filter.
  rpc Stream(ExecutionStreamRequest) returns (stream types.Execution) {}
}

service Runner {
  // Register registers a new runner to the Engine.
  // This endpoint should only be called when the runner is ready to receive execution and emit events.
  // This endpoint returns a credential token that must be use to reach the Runner APIs.
  rpc Register(RunnerRegisterRequest) returns (RunnerRegisterResponse) {}

  // Delete a runner from the blockchain.
  rpc Delete(RunnerDeleteRequest) returns (RunnerDeleteResponse) {}
}
```

Note: the runner register rpc has been moved from Runner API to this Orchestrator API.


server/grpc/server.go
```diff
@@ -92,7 +99,21 @@ func (s *Server) register() {
 	protobuf_api.RegisterOwnershipServer(s.instance, api.NewOwnershipServer(s.mc))
 	protobuf_api.RegisterRunnerServer(s.instance, api.NewRunnerServer(s.mc, s.b))
 
-	runner.RegisterRunnerServer(s.instance, runner.NewServer(s.mc.RPC, s.ep))
+	// decode public keys
+	// TODO: this is not the best place. it should be done directly in config
+	pks := make([]crypto.PubKey, 0)
+	for _, pkS := range s.authorizedPubKeys {
```
antho1404, line 105: It would be nice to have the same format that we use to sign transactions.

eg:
```js
{
      signature: Buffer.from(signature).toString('base64'),
      pub_key: {
        type: "tendermint/PubKeySecp256k1",
        value: Buffer.from(pubKeyByte).toString('base64')
      }
}
```

This way the CLI or any other will not have to deal with another prefix. What do you think?

antho1404:
I had some issues with the validation of the pub key. Here is my `config.yml` file (generated from https://github.com/mesg-foundation/js-sdk/pull/221). The key seems valid but the engines send me an error.
```
AuthorizedPubKeys[0] must be a valid bech32accpubkey
```
```yml
authorized_pubkeys:
  - mesgtestpub1pakk3s28ym3pjj443e4g3u3jnch04cuhgjdfqx
account:
  mnemonic: >-
    guard paddle nature laptop space patient one wait airport side buffalo
    ceiling prevent ten planet immune sample duty fox box option know slot bone
```

What am I missing?

NicolasMahe:
> I had some issues with the validation of the pub key. Here is my `config.yml` file (generated from [mesg-foundation/js-sdk#221](https://github.com/mesg-foundation/js-sdk/pull/221)). The key seems valid but the engines send me an error.
> 
> ```
> AuthorizedPubKeys[0] must be a valid bech32accpubkey
> ```
> 
> ```yaml
> authorized_pubkeys:
>   - mesgtestpub1pakk3s28ym3pjj443e4g3u3jnch04cuhgjdfqx
> account:
>   mnemonic: >-
>     guard paddle nature laptop space patient one wait airport side buffalo
>     ceiling prevent ten planet immune sample duty fox box option know slot bone
> ```
> 
> What am I missing?

Your pubkey seems truncated. The one I'm using is much longer: `mesgtestpub1addwnpepqwl7axwafcqnc9z92eyrl835w6j54r8806x3lnnqedydt47j6papjmn0439`

server/grpc/admin/admin.proto
```diff
@@ -0,0 +1,60 @@
...
+
+// This is the Admin API to interact with the Engine.
+service Admin {
+  // Create an execution on the blockchain.
+  // This endpoint requires the request to be signed and passed in the request's metadata using the key `mesg_request_signature`.
+  rpc CreateExecution(CreateExecutionRequest) returns (CreateExecutionResponse) {}
+
+  // Delete a runner from the blockchain.
+  // This endpoint requires the request to be signed and passed in the request's metadata using the key `mesg_request_signature`.
+  rpc DeleteRunner(DeleteRunnerRequest) returns (DeleteRunnerResponse) {}
```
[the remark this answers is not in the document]
NicolasMahe: I would like to keep the Runner API as an API that ONLY runners should use.
As opposed to this admin API where ONLY admin should use it.
Function != API

server/grpc/server.go
NicolasMahe: Cosmos already has a builtin way to encode and decoded pubkeys in a similar way than addresses: prefix (`mesgtestpub`) + bech32.

What you are showing is the amino-json-encoded of a `tendermint/PubKeySecp256k1` struct.
I think it makes sense to keep the bech32 encoding in yaml file where users should be able to edit it.

Moreover, the cosmos cli already display the pubkey in bech32. I think we should simply do the same.
```
➜  engine git:(feature/e2e-docker-cli) ./bin/mesg-cli keys list
- name: e2e
  type: local
  address: mesgtest1797drk09dqw60wd9ffvj7hag68cqxhnfex6q0x
  pubkey: mesgtestpub1addwnpepqg0ujk8vcwq86z8466rznfvfz5rk992wsy0qtczfedgsy3x90mrzk476x05
  mnemonic: ""
  threshold: 0
  pubkeys: []
- name: engine
  type: local
  address: mesgtest1sfjx3xmrnrllhc86jhfzdpqpckn4s69a5tpr4f
  pubkey: mesgtestpub1addwnpepqwl7axwafcqnc9z92eyrl835w6j54r8806x3lnnqedydt47j6papjmn0439
  mnemonic: ""
  threshold: 0
  pubkeys: []
```

e2e/api_test.go
```diff
@@ -20,36 +22,45 @@ import (
...
+	"github.com/mesg-foundation/engine/server/grpc/orchestrator"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc"
 )
 
 type apiclient struct {
-	pb.EventClient
+	EventClient pb.EventClient
+	// TODO: to switch when e2e tests doesn't create any event
+	// EventClient     orchestrator.EventClient
```
NicolasMahe, line 33: i will do in a another PR. Same for https://github.com/mesg-foundation/engine/pull/1771/files#diff-d7befae5ac5d72ce8456da5eae6f2500R117-R118

server/grpc/orchestrator/execution.go
```diff
@@ -0,0 +1,151 @@
...
+			alreadySeeHashes := make(map[string]bool)
+			for _, attr := range attrs {
+				// skip already see hash. it deduplicate same execution in multiple event.
+				if alreadySeeHashes[attr] {
+					continue
+				}
+				alreadySeeHashes[attr] = true
+				hash, err := hash.Decode(attr)
+				if err != nil {
+					return err
```
antho1404, line 98: Is it a good thing to return the error here, shouldn't we have an error channel?
Here the first error will actually stop the whole stream.
Might have the same issue on the runner API which might be more critical

server/grpc/orchestrator/execution.proto
```diff
@@ -0,0 +1,82 @@
...
+    (gogoproto.moretags) = 'validate:"required,printascii"'
+  ];
+
+  mesg.protobuf.Struct inputs = 2;
+
+  repeated string tags = 3 [
+    (gogoproto.moretags) = 'validate:"dive,printascii"'
+  ];
+
+  bytes executorHash = 8 [
```
antho1404, line 32:
```suggestion
  bytes executorHash = 4 [
```

server/grpc/orchestrator/execution.proto
```diff
@@ -0,0 +1,82 @@
...
+    (gogoproto.moretags) = 'validate:"omitempty,hash"',
+    (gogoproto.casttype) = "github.com/mesg-foundation/engine/hash.Hash"
+  ];
+
+  string price = 9 [
```
antho1404, line 37:
```suggestion
  string price = 5 [
```

server/grpc/orchestrator/runner.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	// delete any other token corresponding to runnerHash
+	s.tokenToRunnerHash.Range(func(key, value interface{}) bool {
+		savedRunnerHash := value.(hash.Hash)
+		if savedRunnerHash.Equal(runnerHash) {
+			s.tokenToRunnerHash.Delete(key)
+		}
+		return true
+	})
```
antho1404, line 95: Is it really useful if we overwrite it later?

server/grpc/server.go
```diff
@@ -92,7 +99,19 @@ func (s *Server) register() {
 	protobuf_api.RegisterOwnershipServer(s.instance, api.NewOwnershipServer(s.mc))
 	protobuf_api.RegisterRunnerServer(s.instance, api.NewRunnerServer(s.mc, s.b))
 
-	runner.RegisterRunnerServer(s.instance, runner.NewServer(s.mc.RPC, s.ep))
+	tokenToRunnerHash := &sync.Map{}
```
antho1404, line 102: would be great later to store this directly in the orchestrator so like that the orchestrator can know what are the running services fro early checks

server/grpc/orchestrator/execution.go
NicolasMahe: There is no error channel with gRPC streams. The only way to return the error to the client is to return it as it does.

server/grpc/orchestrator/runner.go
NicolasMahe: I'm not sure to understand "overwrite later".
this removes other tokens that also points to the same runnerHash.

server/grpc/server.go
NicolasMahe: this can be injected in the New function of this file from the `core/main.go` and shared with orchestrator, if we need to.

server/grpc/orchestrator/runner.go
antho1404: No my bad I was thinking we were replacing it on line 104 but no the map is reversed

server/grpc/orchestrator/authorizer.go
```diff
@@ -0,0 +1,70 @@
...
+func (a *Authorizer) IsAuthorized(ctx context.Context, payload interface{}) error {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return fmt.Errorf("signature not found in metadata, make sure to set it using the key %q", RequestSignature)
+	}
+	if len(md[RequestSignature]) == 0 {
+		return fmt.Errorf("signature not found in metadata, make sure to set it using the key %q", RequestSignature)
+	}
+	signature := md[RequestSignature][0]
+	signatureBytes, err := base64.RawStdEncoding.DecodeString(signature)
```
antho1404, line 51: What's the reason for this extra thing? The `StdEncoding` works fine when we send the signature in plain text as a parameter but without it doesn't work. I cannot make it work with `RawStdEncoding`.

```suggestion
	signatureBytes, err := base64.StdEncoding.DecodeString(signature)
```

NicolasMahe: No reason. that's what VSC suggested to use. I didn't see it have a simple `StdEncoding`.
I replaced the `RawStdEncoding` by the `StdEncoding` across the codebase.

PR #1773, opened 2020-04-07
Use new event API in e2e tests

NicolasMahe:
Dependent on https://github.com/mesg-foundation/engine/pull/1771

This PR updates the e2e tests to use the new event API.
As there is no `CreateEvent` endpoint anymore, I changed the e2e services and processes to get around this.

Diff:
https://github.com/mesg-foundation/engine/pull/1773/files/2d0ef86bce473b75b61b526d47f49c7a1140ae8b..fdecdda4d5beb336980e2145ba050872236a0682

### DIFF
--- a/e2e/event_test.go
+++ b/e2e/event_test.go
@@ -1,112 +1,106 @@
 package main
 
 import (
-	"context"
 	"testing"
-	"time"
-
-	"github.com/mesg-foundation/engine/hash"
-	"github.com/mesg-foundation/engine/protobuf/acknowledgement"
-	pb "github.com/mesg-foundation/engine/protobuf/api"
-	"github.com/mesg-foundation/engine/protobuf/types"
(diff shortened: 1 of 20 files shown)


e2e/event_test.go
```diff
@@ -1,112 +1,106 @@
...
 )
 
 func testEvent(t *testing.T) {
-	stream, err := client.EventClient.Stream(context.Background(), &pb.StreamEventRequest{
-		Filter: &pb.StreamEventRequest_Filter{},
-	})
-	require.NoError(t, err)
-	acknowledgement.WaitForStreamToBeReady(stream)
+	t.SkipNow()
+	// TODO: cannot create an event so this test is not functional
```
NicolasMahe, line 9: Should I simply delete this test?
We are not going to put back the create event endpoint, and the rest of the functions are tested in other tests.

PR #1775, opened 2020-04-07
Simplify container

NicolasMahe:
Dependent on https://github.com/mesg-foundation/engine/pull/1774

This PR merges the `runner/builder` package into `container` package and simplifying it a lot (no more intermediary struct between the service and docker).

Diff of only this PR against the previous:
https://github.com/mesg-foundation/engine/pull/1775/files/46a3994dd48786ba405e88b5bceac5b42824ef2f..d5f0eb8be5a00ed953e0572bf3f35e7e880a214d


container/container.go
```diff
@@ -47,151 +64,409 @@ var statuses = []struct {
...
+				Networks: []swarm.NetworkAttachmentConfig{
+					{
+						Target:  networkID,
+						Aliases: []string{dep.Key},
+					},
+				},
+				RestartPolicy: &swarm.RestartPolicy{
+					Condition:   swarm.RestartPolicyConditionOnFailure,
+					MaxAttempts: &maxAttempts,
+				},
```
NicolasMahe, line 209: I added this policy that only restard the container 3 times. By default it's unlimited.

container/container.go
```diff
@@ -47,151 +64,409 @@ var statuses = []struct {
...
+	}
+
+	return c.deleteNetwork(name)
+}
+
+// deleteData deletes the data volumes of instance and its dependencies.
+// TODO: right now deleteData() is not compatible to work with multiple instances of same
+// service since extractVolumes() accepts service, not instance. same applies in the start
+// api as well. make it work with multiple instances.
+// func deleteData(cont container.Container, s *service.Service) error {
```
NicolasMahe, line 315: This function is not used anymore so I had to comment it because of the lint.
I would like to keep it if we need it in the future.

container/container.go
```diff
@@ -47,151 +64,409 @@ var statuses = []struct {
...
 	}
 	defer tr.Close()
 
-	resp, err := c.client.ImageBuild(context.Background(), tr, types.ImageBuildOptions{
+	if _, err := c.client.ImageBuild(context.Background(), tr, types.ImageBuildOptions{
 		Remove:         true,
 		ForceRemove:    true,
 		SuppressOutput: true,
-	})
+		Tags:           []string{imageTag + srv.Hash.String()},
```
NicolasMahe, line 143: build with a tag 🙆‍♂

antho1404: You removed the `defer resp.Body.Close()`

NicolasMahe: yes because we don't read it anymore and it doesn't seem to cause any issue.
should I put it back?

antho1404: Good point, if we don't read it I think it's fine

antho1404:
We should now be able to safely remove the bind of the docker socket when we start the engine :)
Maybe you could update the dev script too in this PR

PR #1796, opened 2020-04-26
Remove ipfs endpoint from config

NicolasMahe:
- Remove IPFS endpoint from config
- Split container build function in download and build
- update e2e test to reflect the previous changes: build the test services directly from their paths
- Remove e2e service tests mesg.yml as it's not needed anymore.

This PR also simplifies the update of the e2e test services as the compilation step is no needed anymore.


container/container.go
```diff
@@ -90,38 +89,42 @@ func New(ipfsEndpoint, engineName, engineAddress, engineNetwork string) (*Contai
...
 		engineName:     engineName,
 		engineNetwork:  engineNetwork,
 	}, nil
 }
 
-// Build the imge of the container
-func (c *Container) Build(srv *service.Service) error {
+// Download downloads the tarball of the source of a service from a HTTP url.
+// Don't forget to remove the downloaded service.
+func (c *Container) Download(url string) (string, error) {
```
NicolasMahe, line 100: This function is not used anymore but I would like to keep

PR #1806, opened 2020-05-05
Create pre-configured dev docker image

NicolasMahe:
dependent on https://github.com/mesg-foundation/engine/pull/1801 and https://github.com/mesg-foundation/engine/pull/1808

related to https://github.com/mesg-foundation/engine/issues/1815

- Added a dev chain in the folder `dev-chain`.
- Added a dockerfile that use configure the "default" image to a dev image using the generated dev chain
  - This image is starting internally 3 processes by using a custom script that to the following:
    - if one is dead: it terminates all other processes and exit with an error code
    - if the script is terminated (by docker for example) then its terminates processes and exit without error
- Update e2e tests to use it


.circleci/config.yml
```diff
@@ -70,7 +70,7 @@ jobs:
       - run: docker swarm init
       - <<: *restore_go_cache
       - <<: *restore_go_path
-      - run: make e2e
+      - run: make e2e version=`echo $CIRCLE_SHA1 | cut -c1-7`
```
NicolasMahe, line 73: this is reverted in the next PR (i set a default value to version)

scripts/run-e2e.sh
```diff
@@ -2,23 +2,33 @@
...
+  --name engine \
+  -p 1317:1317 \
+  -p 50052:50052 \
+  -p 26657:26657 \
+  --network engine \
+  --label com.docker.stack.namespace=engine \
+  mesg/engine:cli-dev
+
+echo "waiting to give some time to the container to start and run"
+sleep 10 &
```
antho1404, line 30: FYI, I handle the api start with https://github.com/liteflow-labs/liteflow-js/blob/33422349a2260ced8f3ca9584ddd7e9bce5bcf40/packages/api/setup-test-environment#L16-L23

Dockerfile.cli.dev
```diff
@@ -0,0 +1,8 @@
+ARG from
+FROM $from
+
+COPY ./dev-chain/cli /root/.mesg-cli
```
antho1404, line 4: can't we just generate all the commands in the docker file instead of copy the files. This way we don't have to store and maintain the files and we can still have a deterministic chain with mnemonic as args of the docker.

I'm afraid that we will have a lot of issues if we do some migration of the genesis

scripts/run-e2e.sh
```diff
@@ -2,23 +2,33 @@
...
-./scripts/dev.sh -q
-
+if [[ -z $(docker network list -f name="engine" -q) ]]; then
+  docker network create --driver overlay engine
+fi
+docker service create \
+  --name engine \
+  -p 1317:1317 \
+  -p 50052:50052 \
+  -p 26657:26657 \
```
antho1404, line 24: is this port necessary? we shouldn't use it in e2e

scripts/run-e2e.sh
```diff
@@ -2,23 +2,33 @@
...
+  --network engine \
+  --label com.docker.stack.namespace=engine \
+  mesg/engine:cli-dev
+
+echo "waiting to give some time to the container to start and run"
+sleep 10 &
+wait $!
```
antho1404, line 31: why not simply `sleep 10` why do you have to add the `wait $!`

scripts/dev-cli.sh
```diff
@@ -0,0 +1,50 @@
...
+# turn on bash's job control
+set -m
+
+# start daemon
+mesg-daemon start &
+daemon=$!
+
+# wait 5 sec for the daemon to start rpc server
+sleep 5 &
+wait $!
```
antho1404, line 12: Isn't this sufficient?
```suggestion
sleep 5
```

scripts/dev-cli.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
antho1404, line 1: I'm not sure I understand the reason behind this script. It looks quite complicated and a simple
```bash
mesg-daemon start &
daemon=$!
mesg-cli rest-server --laddr tcp://0.0.0.0:1317 &
lcd=$!
mesg-cli orchestrator start &
orchestrator=$!
wait daemon lcd orchestrator
```
might have done the job no?

Dockerfile.cli.dev
NicolasMahe: That's could be possible but I think there will be some issue with the way the cosmos commands takes password: there is not password flag and the passwords are read directly from stdin.
I would like to do this (if possible) in another PR.

There will not be any migration of genesis as this dev version is intended to be used on temp volume that are discarded at every start/stop of the associated docker.

scripts/run-e2e.sh
NicolasMahe: Tendermint RPC 26657 is mandatory to be able to run commands outside of the container and connect to it.
Eg: I'm developing something that use the tendermint RPC client.

The same comment apply to the orchestrator port 50052. It currently not necessary to publish it but it's mandatory to develop services on the host machine.

scripts/run-e2e.sh
NicolasMahe: sleep prevent the script to be stopped by user (ctrl+c), but wait can be killed.
So the solution is to run sleep in background and wait for it.

scripts/dev-cli.sh
NicolasMahe: see answer https://github.com/mesg-foundation/engine/pull/1806#discussion_r421927943

scripts/dev-cli.sh
NicolasMahe: No it's not. The problem with the `wait` command is it will wait for ALL process to stop.

As I said in the description:
>-  if one is dead: it terminates all other processes and exit with an error code
> - if the script is terminated (by docker for example) then its terminates processes and exit without error

Dockerfile.cli.dev
antho1404: The migration I'm talking about is if we add a new parameter and the structure of the genesis changes for some reasons. It would be better to re-generate it all the time or at least have tests to ensure that breaking changes on the genesis will be changed on the dev directory. But otherwise yes a script if possible would be nice in another PR but not a priority

scripts/run-e2e.sh
antho1404: makes sense 👍

scripts/run-e2e.sh
NicolasMahe: I will try to use something like this ;)

Dockerfile.cli.dev
NicolasMahe: In this current version, if the genesis is not valid, then the e2e tests will breaks.

PR #1808, opened 2020-05-05
Add cli dockerfile

NicolasMahe:
This PR adds a new Dockerfile that compiles and exposes the 2 clis.
The default dockerfile cmd is using only on cli, but the otherone can still be usable by overriding the default cmd when starting the container.

### DIFF
--- a/Dockerfile.cli
+++ b/Dockerfile.cli
@@ -11,7 +11,7 @@ ARG version
 RUN go build -mod=readonly -o ./bin/mesg-cli -ldflags="-s -w -X 'github.com/cosmos/cosmos-sdk/version.Name=mesg' -X 'github.com/cosmos/cosmos-sdk/version.ServerName=mesg-daemon' -X 'github.com/cosmos/cosmos-sdk/version.ClientName=mesg-cli' -X 'github.com/cosmos/cosmos-sdk/version.Version=$version'" ./cmd/mesg-cli/
 RUN go build -mod=readonly -o ./bin/mesg-daemon -ldflags="-s -w -X 'github.com/cosmos/cosmos-sdk/version.Name=mesg' -X 'github.com/cosmos/cosmos-sdk/version.ServerName=mesg-daemon' -X 'github.com/cosmos/cosmos-sdk/version.ClientName=mesg-cli' -X 'github.com/cosmos/cosmos-sdk/version.Version=$version'" ./cmd/mesg-daemon/
 
-# ubuntu image wit binaries for distribution
+# ubuntu image with binaries for distribution
 FROM ubuntu:18.04
 RUN apt-get update && \
   apt-get install -y --no-install-recommends ca-certificates=20180409 && \
--- a/Dockerfile.cli
+++ b/Dockerfile.cli
@@ -1,0 +1,27 @@
+# base Go image version for building the binaries
+FROM golang:1.13.10 AS build
+WORKDIR /app
+
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . .
+ARG version
+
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -61,6 +61,9 @@ build-cmd: dep
 	go build -mod=readonly -o ./bin/mesg-cli ./cmd/mesg-cli/
 	go build -mod=readonly -o ./bin/mesg-daemon ./cmd/mesg-daemon/
 
+build-docker-cli: check-version
+	docker build -t mesg/engine:cli -f ./Dockerfile.cli --build-arg version=$(version) .
```
NicolasMahe, line 65: `mesg/engine:cli` is a temp image tag. it will be replaced in following PR

Dockerfile.tools
```diff
@@ -1,5 +1,5 @@
 # base Go image version.
-FROM golang:1.13.0-stretch
+FROM golang:1.13.10
```
NicolasMahe, line 2: updated this in order to use the same base image in all future dockerfile

Dockerfile.cli
```diff
@@ -0,0 +1,27 @@
...
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY . .
+ARG version
+
+RUN go build -mod=readonly -o ./bin/mesg-cli -ldflags="-s -w -X 'github.com/cosmos/cosmos-sdk/version.Name=mesg' -X 'github.com/cosmos/cosmos-sdk/version.ServerName=mesg-daemon' -X 'github.com/cosmos/cosmos-sdk/version.ClientName=mesg-cli' -X 'github.com/cosmos/cosmos-sdk/version.Version=$version'" ./cmd/mesg-cli/
+RUN go build -mod=readonly -o ./bin/mesg-daemon -ldflags="-s -w -X 'github.com/cosmos/cosmos-sdk/version.Name=mesg' -X 'github.com/cosmos/cosmos-sdk/version.ServerName=mesg-daemon' -X 'github.com/cosmos/cosmos-sdk/version.ClientName=mesg-cli' -X 'github.com/cosmos/cosmos-sdk/version.Version=$version'" ./cmd/mesg-daemon/
+
+# ubuntu image wit binaries for distribution
```
NicolasMahe, line 14:
```suggestion
# ubuntu image with binaries for distribution
```

Dockerfile.cli
```diff
@@ -0,0 +1,27 @@
...
+COPY . .
+ARG version
+
+RUN go build -mod=readonly -o ./bin/mesg-cli -ldflags="-s -w -X 'github.com/cosmos/cosmos-sdk/version.Name=mesg' -X 'github.com/cosmos/cosmos-sdk/version.ServerName=mesg-daemon' -X 'github.com/cosmos/cosmos-sdk/version.ClientName=mesg-cli' -X 'github.com/cosmos/cosmos-sdk/version.Version=$version'" ./cmd/mesg-cli/
+RUN go build -mod=readonly -o ./bin/mesg-daemon -ldflags="-s -w -X 'github.com/cosmos/cosmos-sdk/version.Name=mesg' -X 'github.com/cosmos/cosmos-sdk/version.ServerName=mesg-daemon' -X 'github.com/cosmos/cosmos-sdk/version.ClientName=mesg-cli' -X 'github.com/cosmos/cosmos-sdk/version.Version=$version'" ./cmd/mesg-daemon/
+
+# ubuntu image with binaries for distribution
+FROM ubuntu:18.04
+RUN apt-get update && \
+  apt-get install -y --no-install-recommends ca-certificates=20180409 && \
```
antho1404, line 17: do we still need this install? If I remember correctly it was related to the fetch from git that we had but we don't have this anymore.

NicolasMahe: I think it's good to install the `ca certificates` to verify SSL connection.
User could setup a SSL remote connection to another peer or to a RPC endpoint?

PR #1809, opened 2020-05-05
fix runner e2e that block the tests on error

NicolasMahe:
fix runner e2e that block the tests on error

to remember: when a `require.` function "fail" it will not run other `t.Run` functions but will run the remaining code in the parent function.

### DIFF
--- a/e2e/runner_test.go
+++ b/e2e/runner_test.go
@@ -69,9 +69,10 @@ func testRunner(t *testing.T) {
 			require.NoError(t, cont.Start(testServiceStruct, testInstanceHash, testRunnerHash, testInstanceEnvHash, testInstanceEnv, registerPayloadSignature))
 		})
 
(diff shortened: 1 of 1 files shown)


antho1404:
Not sure I understand this fix? I don't see any problem merging that but not sure what it actually fix

NicolasMahe:
> Not sure I understand this fix? I don't see any problem merging that but not sure what it actually fix

behavior in the dev version:
if the previous test fails (https://github.com/mesg-foundation/engine/pull/1809/files#diff-b45f199c25c42d4525f0000d48498a06R68-R70) then the `stream.Recv()` blocks forever and the e2e tests never fail properly.

behavior in this branch version:
if the previous test fails, then the test following test (that should block with `stream.Recv()`) is simply never skipped and the e2e tests fails properly.

PR #1817, opened 2020-05-08
Use docker container in e2e tests

NicolasMahe:
Update e2e test to use docker container instead of docker service

### DIFF
--- a/container/container.go
+++ b/container/container.go
@@ -4,24 +4,22 @@ import (
 	"context"
 	"encoding/base64"
 	"errors"
-	"fmt"
 	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
 	"strconv"
-	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


container/container.go
```diff
@@ -174,115 +154,115 @@ func (c *Container) Start(srv *service.Service, instanceHash, runnerHash, instan
...
 			},
-			EndpointSpec: &swarm.EndpointSpec{
-				Ports: convertPorts(dep.Ports),
+			Env:          dep.Env,
+			ExposedPorts: exposedPort,
+		}
+		hostConfig := &containertypes.HostConfig{
+			PortBindings: portBindings,
+			Mounts:       append(volumes, volumesFrom...),
+			RestartPolicy: containertypes.RestartPolicy{
```
antho1404, line 173: Not critical for this PR but I would be strict on the restart policy here as it is used only for e2e and not allow any restart

PR #1818, opened 2020-05-09
Add orchestrator logs and publish them on a new gRPC stream - advanced version

NicolasMahe:
Based on PR https://github.com/mesg-foundation/engine/pull/1812 but can be merged instead.

The difference with the other PR is the introduction of specific types of logs that hold specific data.

The concept is to not use a classic log system, but a system that describes exactly what happens in the orchestrator so we can be GUIs that are easy for users to understand.

The previous PR (https://github.com/mesg-foundation/engine/pull/1812) is basically a classic log system exposed on a gRPC Stream. It is overkill as it's already easy to access logs in the stdout of any container with docker (even more, docker has multiple log driver that , and not general-purpose as it is compatible only with the orchestrator.

This new PR transforms the previous log system to a system closer to state machine transition logs or a tracer tool.

New Orchestrator log protobuf:
```proto
message OrchestratorLog {

  message TypeErrorOccurred {
    string error = 1;
  }

  message TypeExecutionCreated {
    // Hash of the execution created by this process. Can be empty.
    bytes executionHash = 6 [
      (gogoproto.casttype) = "github.com/mesg-foundation/engine/hash.Hash"
    ];
  }

  message TypeMapExecuted {}

  message TypeFilterMatched {}

  message TypeFilterDidNotMatched {}

  message TypeTriggeredByResult {
    bytes executionHash = 1 [
      (gogoproto.casttype) = "github.com/mesg-foundation/engine/hash.Hash"
    ];
  }

  message TypeTriggeredByEvent {
    bytes eventHash = 1 [
      (gogoproto.casttype) = "github.com/mesg-foundation/engine/hash.Hash"
    ];
  }

  // Hash of the process
  bytes processHash = 1 [
    (gogoproto.casttype) = "github.com/mesg-foundation/engine/hash.Hash"
  ];

  // Key of the node being executed.
  string nodeKey = 2;
  
  // Type of the node being executed.
  string nodeType = 3;

  // Time used in nanoseconds by the orchestrator to execute this action.
  int64 timeConsumed = 4;

  oneof type {
    TypeExecutionCreated    executionCreated = 5;
    TypeMapExecuted         mapExecuted = 6;
    TypeFilterMatched       filterMatched = 7;
    TypeFilterDidNotMatched filterDidNotMatched = 8;
    TypeTriggeredByResult   triggeredByResult = 9;
    TypeTriggeredByEvent    triggeredByEvent = 10;
    TypeErrorOccurred       errorOccurred = 11;
  }
}
```


antho1404:
I don't really like this approach, I find it way too complicated just to get the "type" of logs that it is. 
Also, the time consumed is confusing, what is it really in case of an event or an error for example. I would prefer to add more logs and see the delay between these logs than a duration that we don't really know when it starts/ends.

I think the previous PR is more what we are looking for. For the part that it might be overkill to have it in an API, that might be right (we had the errors without any need for an API). 

For me, the goal is more to have differents transport for the logs:
- standard output like currently
- WebSocket/stream
- file
- systems like sentry for error tracking
...

Basically one log system with multiple "outputs" that's why I like the simplicity of a simple "log level", "message", "data" and I find this solution overkill and limiting but maybe I don't see the full picture of what you have in mind.

The part to connect to a UI for visual debugging, for example, is interesting but again I would rather filter specific logs and transport them through a specific API that can have specific types.

NicolasMahe:
closing for a simpler version

PR #1825, opened 2020-05-18
Add filter on execution list

antho1404:
Add filter for execution list on the keeper.

Filter available:
- parentHash: base58 hash
- eventHash: base58 hash
- instanceHash: base58 hash
- processHash: base58 hash
- status string (Proposed, InProgress, Completed, Failed)
- stepKey: string

These filters are available as query parameter in the `/execution/list` API 
eg:
```
curl http://localhost:1317/execution/list\?instanceHash\=7dTayprXsBTzc7Z6iqCTjfJzh3Sbtasdgh9wdPzT4gUL\&processHash\=BACectqn5sHeDfK7N6dQdCKKY9ZfrCt5Rh428RYERYXU
```

PR #1826, opened 2020-05-19
Add pagination meta data in the header

antho1404:
dependency: #1825 

Add pagination detail on the header in order to have the client able to calculate the total number of pages.

I use the `X-` that was a recommendation for custom headers. It is not recommended anymore and it is recommended to use HTTP Link headers but those only support prev and next

### DIFF
--- a/x/execution/client/rest/query.go
+++ b/x/execution/client/rest/query.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"sort"
+	"strconv"
 
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/context"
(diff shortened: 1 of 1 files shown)


x/execution/client/rest/query.go
```diff
@@ -132,6 +132,9 @@ func queryListHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 		sort.Sort(sort.Reverse(execution.ByBlockHeight(execs)))
 
 		start, end := client.Paginate(len(execs), page, limit, limit)
+		w.Header().Set("X-Page", fmt.Sprintf("%d", page))
+		w.Header().Set("X-Limit", fmt.Sprintf("%d", limit))
+		w.Header().Set("X-Total-Count", fmt.Sprintf("%d", len(execs)))
```
NicolasMahe, line 137: don't use `fmt.Sprintf` for type conversion:
```suggestion
		w.Header().Set("X-Page", strconv.Itoa(page))
		w.Header().Set("X-Limit", strconv.Itoa(limit))
		w.Header().Set("X-Total-Count", strconv.Itoa(len(execs)))
```

NicolasMahe: don't forget to import `strconv`

PR #1827, opened 2020-05-19
update dev script to use docker container instead of docker service

NicolasMahe:
- update dev script to use container instead of service
- fix make clean

### DIFF
--- a/scripts/dev.sh
+++ b/scripts/dev.sh
@@ -17,13 +17,10 @@ if [[ "$2" == "monitoring" ]]; then
 fi
 
 function onexit {
-  docker service rm $ENGINE_NAME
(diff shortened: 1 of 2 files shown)


scripts/dev.sh
```diff
@@ -32,35 +29,41 @@ function onexit {
...
+docker run \
+  -d \
+  --rm \
   --name $ENGINE_NAME \
   -p 1317:1317 \
   -p 50052:50052 \
   -p 26657:26657 \
   --network $NETWORK_NAME \
-  --label com.docker.stack.namespace=$ENGINE_NAME \
+  --volume engine:/root/ \
```
NicolasMahe, line 66: I set up a volume so the data are persisted

PR #1839, opened 2020-05-23
Pay processes based on the owner's wallet

antho1404:
Instead of taking the coins from the process address, it uses a specific field of the process. This field can be (as default) the process address or any other addresses.

I added the feature that by default when a process is created, the owner pays for the executions. 

With this, we don't need to have to maintain balances for each process and we can also later extend to update the process "payment address" to any wallet that the owner of the process has access (a shared wallet, a service wallet...)


antho1404:
merged into https://github.com/mesg-foundation/engine/pull/1842

PR #1840, opened 2020-05-23
Add optional task fee

antho1404:
We can now set a `fee` to a task that will be added to the default execution min price while executing it. This allows us to have different prices for different tasks. 
An error is thrown if the user/orchestrator doesn't provide the minimal execution price + task fee.

I also changed the way the orchestrator is initialized. the `minExecPrice` was given in string but and only throwing an error if the format is invalid on the first execution created by the orchestrator (which can take a while). Now the error is thrown directly in the main so if the formatting is wrong, the orchestrator will not start.


antho1404:
merged into https://github.com/mesg-foundation/engine/pull/1842

PR #1841, opened 2020-05-26
Fix dev-starter when user stop script during the first sleep of 5 sec

NicolasMahe:
Fix dev-starter when user stop script during the first sleep of 5 sec

### DIFF
--- a/scripts/dev-starter.sh
+++ b/scripts/dev-starter.sh
@@ -7,15 +7,29 @@ set -m
 mesg-daemon start &
 daemon=$!
 
+# function that stop child processes
+function stop {
+  echo "stopping daemon"
+  kill $daemon
+  wait $daemon
+  exit 0
+}
+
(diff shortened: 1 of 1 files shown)


scripts/dev-starter.sh
```diff
@@ -7,15 +7,29 @@ set -m
...
 sleep 5 &
 wait $!
 
 # start lcd
+echo "starting lcd"
 mesg-cli rest-server --laddr tcp://0.0.0.0:1317 &
 lcd=$!
 
 # start orchestrator
+echo "starting orchestrator"
```
NicolasMahe, line 32: @antho1404  Are those log disturbing for the reading of the orchestrator log by CLI?

antho1404: I didn't test but it shouldn't as we start to log only when the API is ready and with a tail of 0. Will need confirmation of that.

PR #1842, opened 2020-05-26
Execution price based on service task price and new credit module

antho1404:
## New credit module

A credit module that stores a balance for each address (default 0) that can be negative. This is quite similar to the default bank module but with the possibility to go bellow 0 and have negative credits.

## Process address payment

The process now has a new attribute `paymentAddress` that delegates the payment of executions of this process to a specific address. The default is the owner of the process. The process doesn't have any balance anymore and each executions transfer from the `paymentAddress` of the process and not the process itself. We can still have the a process that has credits and use the address of the process as `paymentAddress`.

## Execution price

The execution price is not set by the user based on a `minExecPrice` anymore but calculated based on 3 variables defined by service's tasks:
- `perCall`: The cost in credit of an invocation
- `perSec`: The cost in credit of a second of invocation (rounded to the higher second, 1.2 sec will cost 2sec)
- `perKB`: The cost in credit of the sum of the inputs and outputs of the execution (rounded to the highest kB, 0.4 kB will cost the price of 1 kB)
The current calculation is 
```
perCall + duration x perSec + size(input) x perKB + size(input) x perKB
```

All these payment are done with the credit modules and not the native token anymore. This allows to have negative credits for executions that can be paid later.

### TODO in other PR / discuss:
- [ ] Export/Import for the genesis
- [ ] Limit the balance as parameter of the credit module (safe limit) (maybe directly in the data to have control per user)
- [ ] Handler to change payment address in the process


x/execution/internal/keeper/keeper.go
```diff
@@ -469,3 +406,53 @@ func (k *Keeper) fetchEmitters(ctx sdk.Context, proc *process.Process, nodeKey s
...
+	}
+	srv, err := k.serviceKeeper.Get(ctx, inst.ServiceHash)
+	if err != nil {
+		return sdk.Int{}, err
+	}
+	task, err := srv.GetTask(exec.TaskKey)
+	if err != nil {
+		return sdk.Int{}, err
+	}
+	perCall := sdk.NewInt(200)
```
antho1404, line 423: these constants will move into module parameter in another PR

x/execution/internal/keeper/keeper.go
```diff
@@ -469,3 +406,53 @@ func (k *Keeper) fetchEmitters(ctx sdk.Context, proc *process.Process, nodeKey s
...
+	perKB := sdk.NewInt(100)
+	if task.Price != nil && task.Price.PerKB != "" {
+		var ok bool
+		perKB, ok = sdk.NewIntFromString(task.Price.PerKB)
+		if !ok {
+			return sdk.Int{}, fmt.Errorf("invalid price per kb %q", task.Price.PerKB)
+		}
+	}
+	duration := sdk.NewInt(exec.GetDuration())
+	inputs, err := k.cdc.MarshalBinaryLengthPrefixed(exec.Inputs)
```
antho1404, line 448: Size of inputs/outputs based on what we actually store in db. It might not be absolutely exact as we save the whole execution so maybe there are some kind of gzip or anything that could make this data even small.
We could as well calculate the size on the whole execution but this would require that all meta data that the user doesn't control has to be paid for which is an idea that I don't really like.

execution/execution.go
```diff
@@ -70,3 +71,10 @@ func (execution *Execution) Fail(err error) error {
 	execution.Status = Status_Failed
 	return nil
 }
+
+// GetDuration returns the duration of the execution in seconds
+func (execution *Execution) GetDuration() int64 {
+	delta := execution.Stop - execution.Start
+	deltaInSecond := float64(delta) / 1e9
+	return int64(math.Ceil(deltaInSecond))
```
antho1404, line 79: For now the calculation is in second and not per 100ms in order to make the calculation simpler. We can change it in 100ms if it's critical but I don't think it is right now (at least in this PR)

e2e/api_test.go
```diff
@@ -50,8 +50,7 @@ const (
 )
 
 var (
-	executionPrice        = sdk.NewCoins(sdk.NewInt64Coin("atto", 10000)) // /x/execution/internal/types/params.go#DefaultMinPrice
-	processInitialBalance = sdk.NewCoins(sdk.NewInt64Coin("atto", 10000000))
+	executionPrice = sdk.NewCoins(sdk.NewInt64Coin("atto", 10000)) // /x/execution/internal/types/params.go#DefaultMinPrice
```
NicolasMahe, line 53: this should be deleted, no?

protobuf/types/service.proto
```diff
@@ -36,6 +36,17 @@ message Service {
...
+    message Price {
+      string perCall = 1 [
+        (gogoproto.moretags) = 'hash:"name:1" validate:"bigint"'
+      ];
+      string perSec = 2 [
+        (gogoproto.moretags) = 'hash:"name:2" validate:"bigint"'
+      ];
+      string perKB = 3 [
+        (gogoproto.moretags) = 'hash:"name:3" validate:"bigint"'
+      ];
```
NicolasMahe, line 48: I'm not against the string type in proto so it's easier to implement in other languages,
But in go it should be `Int`. Did you try `gogoproto.customtype` or `gogoproto.casttype`?

protobuf/types/service.proto
```diff
@@ -60,6 +71,11 @@ message Service {
     repeated Parameter outputs = 7 [
       (gogoproto.moretags) = 'hash:"name:7" validate:"dive,required"'
     ];
+
+    // price for the task.
+    Price price = 9 [
+      (gogoproto.moretags) = 'hash:"name:9"'
```
NicolasMahe, line 77: I think the `dive` is mandatory to tell the validator to check the struct.
And I guess this this a required parameter?
```suggestion
      (gogoproto.moretags) = 'hash:"name:9" validate:"dive,required"'
```

protobuf/types/service.proto
NicolasMahe: Is `bigint` validation accept empty?
I suggest to explicitly said if it requires or not:
```suggestion
    message Price {
      string perCall = 1 [
        (gogoproto.moretags) = 'hash:"name:1" validate:"required,bigint"'
      ];
      string perSec = 2 [
        (gogoproto.moretags) = 'hash:"name:2" validate:"required,bigint"'
      ];
      string perKB = 3 [
        (gogoproto.moretags) = 'hash:"name:3" validate:"required,bigint"'
      ];
```
or if optional:
```suggestion
    message Price {
      string perCall = 1 [
        (gogoproto.moretags) = 'hash:"name:1" validate:"omitempty,bigint"'
      ];
      string perSec = 2 [
        (gogoproto.moretags) = 'hash:"name:2" validate:"omitempty,bigint"'
      ];
      string perKB = 3 [
        (gogoproto.moretags) = 'hash:"name:3" validate:"omitempty,bigint"'
      ];
```

server/grpc/runner/runner.go
```diff
@@ -34,6 +36,7 @@ func NewServer(rpc *cosmos.RPC, eventPublisher *publisher.EventPublisher, tokenT
 		rpc:               rpc,
 		eventPublisher:    eventPublisher,
 		tokenToRunnerHash: tokenToRunnerHash,
+		inProgress:        &sync.Map{},
```
NicolasMahe, line 39: More context please
```suggestion
		execInProgress:        &sync.Map{},
```

server/grpc/runner/runner.go
```diff
@@ -139,9 +143,15 @@ func (s *Server) Result(ctx context.Context, req *ResultRequest) (*ResultRespons
 	if err != nil {
 		return nil, err
 	}
+	start, ok := s.inProgress.Load(req.ExecutionHash.String())
+	if !ok {
+		panic("execution should be in memory")
```
NicolasMahe, line 148: return error please

x/credit/internal/keeper/keeper.go
```diff
@@ -0,0 +1,97 @@
...
+
+// Sub a number of credits to an address
+func (k Keeper) Sub(ctx sdk.Context, address sdk.AccAddress, amount sdk.Int) (sdk.Int, error) {
+	value, err := k.Get(ctx, address)
+	if err != nil {
+		return sdk.Int{}, err
+	}
+	res := value.Sub(amount)
+	return k.Set(ctx, address, res)
+}
```
NicolasMahe, line 51: Don't you want to also emit events with the amount added or subtracted and not only the updated balance?

x/credit/internal/keeper/keeper.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	store.Set(address.Bytes(), encoded)
+
+	// emit event
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventType,
+			sdk.NewAttribute(sdk.AttributeKeyAction, types.AttributeActionUpdated),
+			sdk.NewAttribute(types.AttributeKeyAddress, address.String()),
+			sdk.NewAttribute(types.AttributeKeyValue, amount.String()),
```
NicolasMahe, line 79: Let's be more explicit. This is the balance:
```suggestion
			sdk.NewAttribute(types.AttributeKeyBalance, amount.String()),
```

x/credit/internal/keeper/keeper.go
```diff
@@ -0,0 +1,97 @@
...
+		),
+	)
+	return amount, nil
+}
+
+// Get the amount of a specific address
```
NicolasMahe, line 85:
```suggestion
// Get the balance of a specific address
```

x/credit/internal/keeper/keeper.go
```diff
@@ -0,0 +1,97 @@
...
+		return err
+	}
+	if _, err := k.Add(ctx, to, amount); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Set a number of credit to a specific address
+func (k Keeper) Set(ctx sdk.Context, address sdk.AccAddress, amount sdk.Int) (sdk.Int, error) {
```
NicolasMahe, line 65: Let's rename amount to balance to be more explicit about what is it.
```suggestion
func (k Keeper) Set(ctx sdk.Context, address sdk.AccAddress, balance sdk.Int) (sdk.Int, error) {
```

x/execution/internal/keeper/keeper.go
```diff
@@ -37,11 +28,21 @@ type Keeper struct {
...
+	creditKeeper   types.CreditKeeper
 	paramstore     params.Subspace
 }
 
 // NewKeeper creates a execution keeper
-func NewKeeper(cdc *codec.Codec, key sdk.StoreKey, bankKeeper types.BankKeeper, serviceKeeper types.ServiceKeeper, instanceKeeper types.InstanceKeeper, runnerKeeper types.RunnerKeeper, processKeeper types.ProcessKeeper, paramstore params.Subspace) Keeper {
+func NewKeeper(
+	cdc *codec.Codec,
+	key sdk.StoreKey,
+	bankKeeper types.BankKeeper,
```
NicolasMahe, line 39: bankKeeper is not used anymore. to delete.

server/grpc/runner/runner.go
```diff
@@ -139,9 +143,15 @@ func (s *Server) Result(ctx context.Context, req *ResultRequest) (*ResultRespons
...
+	}
 	msg := executionmodule.MsgUpdate{
 		Executor: acc.GetAddress(),
 		Hash:     req.ExecutionHash,
+		Start:    start.(uint64),
+		Stop:     uint64(time.Now().UnixNano()),
```
NicolasMahe, line 154: why not just keeping the elapsed time?

x/execution/internal/keeper/keeper.go
```diff
@@ -470,6 +407,62 @@ func (k *Keeper) fetchEmitters(ctx sdk.Context, proc *process.Process, nodeKey s
...
+	if err != nil {
+		return sdk.Int{}, err
+	}
+	srv, err := k.serviceKeeper.Get(ctx, inst.ServiceHash)
+	if err != nil {
+		return sdk.Int{}, err
+	}
+	task, err := srv.GetTask(exec.TaskKey)
+	if err != nil {
+		return sdk.Int{}, err
```
NicolasMahe, line 421: All error in keeper should be wrapped

x/execution/internal/keeper/keeper.go
```diff
@@ -470,6 +407,62 @@ func (k *Keeper) fetchEmitters(ctx sdk.Context, proc *process.Process, nodeKey s
...
+	}
+	perCall := sdk.NewInt(200)
+	if task.Price != nil && task.Price.PerCall != "" {
+		price, ok := sdk.NewIntFromString(task.Price.PerCall)
+		if !ok {
+			return sdk.Int{}, fmt.Errorf("invalid price per call %q", task.Price.PerCall)
```
NicolasMahe, line 427: All errors in keeper should be wrapped

x/execution/internal/keeper/keeper.go
NicolasMahe: What about a simple `sizeOf`?
https://dlintw.github.io/gobyexample/public/memory-and-sizeof.html

> We could as well calculate the size on the whole execution but this would require that all meta data that the user doesn't control has to be paid for which is an idea that I don't really like.

So why forcing to encode with `MarshalBinaryLengthPrefixed` that add a lot of metadata?

execution/execution_test.go
```diff
@@ -71,3 +72,27 @@ func TestExecutionHash(t *testing.T) {
 
 	require.NoError(t, quick.Check(f, nil))
 }
+
+func TestGetDuration(t *testing.T) {
```
NicolasMahe, line 76: All time manipulation should use the time package.
We should never do stuff like `1000*1000*1000`, this is error prone.

ext/xvalidator/validator.go
```diff
@@ -133,6 +133,14 @@ func New(prefix string) (*validator.Validate, ut.Translator) {
 		return t
 	})
 
+	validate.RegisterValidation("bigint", IsBigInt)
+	validate.RegisterTranslation("bigint", trans, func(ut ut.Translator) error {
+		return ut.Add("bigint", "{0} must be a valid credit", false)
```
NicolasMahe, line 138: There is not about credit here.
```suggestion
		return ut.Add("bigint", "{0} must be a valid big int", false)
```

process/process.go
```diff
@@ -19,6 +19,7 @@ func New(name string, nodes []*Process_Node, edges []*Process_Edge) (*Process, e
 	}
 	p.Hash = hash.Dump(p)
 	p.Address = sdk.AccAddress(crypto.AddressHash(p.Hash))
+	p.PaymentAddress = p.Address
```
NicolasMahe, line 22: are you sure about this? shouldn't it be set to an address passed by parameter?

x/execution/internal/keeper/keeper.go
```diff
@@ -291,6 +269,14 @@ func (k *Keeper) Update(ctx sdk.Context, msg types.MsgUpdate) (*executionpb.Exec
 		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "no execution result supplied")
 	}
 
+	exec.Start = msg.Start
+	exec.Stop = msg.Stop
```
NicolasMahe, line 273: I definitely suggest using elapsed time rather than saving the beginning and the end timestamp.

x/execution/internal/keeper/keeper.go
```diff
@@ -470,6 +407,62 @@ func (k *Keeper) fetchEmitters(ctx sdk.Context, proc *process.Process, nodeKey s
...
+	}
+	perKB := sdk.NewInt(100)
+	if task.Price != nil && task.Price.PerKB != "" {
+		price, ok := sdk.NewIntFromString(task.Price.PerKB)
+		if !ok {
+			return sdk.Int{}, fmt.Errorf("invalid price per kb %q", task.Price.PerKB)
+		}
+		if price.GT(perKB) {
+			perKB = price
+		}
```
NicolasMahe, line 451: didn't we say we could put whatever value we want? like 0 credit per KB but a lot per call?

x/execution/internal/types/msg.proto
```diff
@@ -76,4 +71,14 @@ message MsgUpdate {
...
+
+  // Start time
+  uint64 start = 5 [
+    (gogoproto.moretags) = 'validate:"required"'
+  ];
+  
+  // Stop time
+  uint64 stop = 6 [
+    (gogoproto.moretags) = 'validate:"required,gtfield=Start"'
+  ];
```
NicolasMahe, line 83: Let's use a better type here or at least say what is it in the comment.
You can use `gogoproto.casttype` to cast it to `time.Duration` or something a bit better than `uint64`.

NicolasMahe: Is it even a good idea to use `uint64` type in the proto?
You used a string for the task' prices.
I would recommend using a safe type in proto, and a specific one in Go using `gogoproto.casttype` or `gogoproto.customtype`

x/process/internal/keeper/keeper.go
```diff
@@ -72,13 +68,7 @@ func (k Keeper) Create(ctx sdk.Context, msg *types.MsgCreate) (*processpb.Proces
...
 	}
 
-	procInitBal, err := sdk.ParseCoins(processCreateInitialBalance)
-	if err != nil {
-		return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidCoins, err.Error())
-	}
-	if err := k.bankKeeper.SendCoins(ctx, msg.Owner, p.Address, procInitBal); err != nil {
-		return nil, err
-	}
+	p.PaymentAddress = msg.Owner
```
NicolasMahe, line 71: this should be set in the function `process.New`.
Once a resource is created, we should avoid as much as possible to modify it.

NicolasMahe:
TODO:
- [ ] update the ownership transfer function:
https://github.com/mesg-foundation/engine/blob/dev/x/ownership/internal/keeper/keeper.go#L134-L152

cmd/mesg-cli/orchestrator.go
```diff
@@ -82,8 +82,7 @@ func startOrchestratorCmd(cdc *codec.Codec) *cobra.Command {
 			ep := publisher.New(rpc)
 
 			// orchestrator
-			logger.Info("Starting orchestrator")
```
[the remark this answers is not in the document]
antho1404: deleted by mistake or maybe a wrong merge

e2e/api_test.go
antho1404: done

server/grpc/runner/runner.go
antho1404: done

x/execution/internal/keeper/keeper.go
antho1404: I don't think `sizeOf` is a good idea. First, it will be much easier to correlate the price to the actual data that we store so if we want to verify we can just see what's in the store. That's also the way gas is calculated, based on the message to store. Also, I'm not sure `sizeOf` behave the same way on all machines.

server/grpc/runner/runner.go
antho1404: The error that is returned is the error for the client (here the service). In this case, the error is an orchestrator error and should be treated by the orchestrator otherwise we might skip these kinds of errors.
I agree that the panic is a bit too much but I don't think to return the error is safe here. Maybe we should have an error channel in the orchestrator or something like that.

server/grpc/runner/runner.go
antho1404: I don't really like to keep only the elapsed time because we lose some information. Having the duration is a thing but knowing when it started (or stopped) is good too if we want to debug what happened.
We could have the duration but in this case, we also need to have the `startTime` to make sure that we can calculate everything again.
In general, I prefer to have raw data here that can be used to calculate and not just have the calculation result that for some reasons may be wrong.

process/process.go
antho1404: I wanted to keep the default behavior which is using the `process` address as payment address but that might not be necessary as this is a breaking change anyway.

x/execution/internal/keeper/keeper.go
antho1404: see comment here https://github.com/mesg-foundation/engine/pull/1842#discussion_r432919324

x/execution/internal/keeper/keeper.go
antho1404: If we don't check for minimum value then this is just a default and any user can go bellow this. This is a minimal value and not a default value but nothing stop us to put a minimal value equals to 0 and then have the behavior that you describe.

Another way is to set this `price` field require and have the check on the `service.create` only but this is a massive breaking change as all services need to be redeployed.

protobuf/types/service.proto
antho1404: see https://github.com/mesg-foundation/engine/pull/1842#discussion_r432923254

x/credit/internal/keeper/keeper.go
antho1404: Not sure this is really useful for now, I suggest to add it when we need it and for now to keep it simple with only one event.

server/grpc/runner/runner.go
NicolasMahe: let's put the error in the logger (the logger should be passed to this struct, the package `server/grpc` has it)

protobuf/types/service.proto
antho1404: After discussion, we set the prices to required and services will need to be redeployed

x/execution/internal/keeper/keeper.go
antho1404: After discussion, we will not put any limitation on the price, no default, and no minimum limit. Price is now required for all tasks so we can simplify this code.

server/grpc/runner/runner.go
antho1404: As discussed, we agreed on keeping start and stop as we will always need 2 out of the 3 parameters (start, stop, duration) in order to calculate anything.

x/credit/internal/keeper/keeper.go
antho1404: Will replace the event in the set function by events in the `add` and `sub` functions to focus on the action itself.

x/execution/internal/keeper/keeper.go
antho1404: as agreed, we keep the current implementation

server/grpc/runner/runner.go
antho1404: done and set a default start time to the current time (that should be at least 1nano sec that will be converted to 1s of execution)

x/execution/internal/types/msg.proto
antho1404: Duration is different than time but I did replace the time from `uint64` to `int64` to avoid many conversions.

protobuf/types/service.proto
antho1404: done

x/credit/internal/keeper/keeper.go
antho1404: events added "added" and "subtracted"

x/execution/internal/keeper/keeper.go
```diff
@@ -294,8 +291,8 @@ func (k *Keeper) Update(ctx sdk.Context, msg types.MsgUpdate) (*executionpb.Exec
 		}
 		from = proc.PaymentAddress
 	}
-
-	if err := k.creditKeeper.Transfer(ctx, from, runExecutor.Address, price); err != nil {
+	_, err = k.creditKeeper.Sub(ctx, from, price)
+	if err != nil {
```
NicolasMahe:
```suggestion
	if _, err := k.creditKeeper.Sub(ctx, from, price); err != nil {
```

x/process/internal/types/msg.go
```diff
@@ -22,7 +22,7 @@ func (msg MsgCreate) ValidateBasic() error {
 	if err := xvalidator.Struct(msg); err != nil {
 		return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, err.Error())
 	}
-	if _, err := processpb.New(msg.Name, msg.Nodes, msg.Edges); err != nil {
+	if _, err := processpb.New(msg.Name, msg.Nodes, msg.Edges, sdk.AccAddress{}); err != nil {
```
NicolasMahe, line 25:
```suggestion
	if _, err := processpb.New(msg.Name, msg.Nodes, msg.Edges, msg.Owner); err != nil {
```

process/process.go
```diff
@@ -11,15 +11,20 @@ import (
...
 	}
 	p.Hash = hash.Dump(p)
 	p.Address = sdk.AccAddress(crypto.AddressHash(p.Hash))
-	p.PaymentAddress = p.Address
+	if !paymentAddress.Empty() {
+		p.PaymentAddress = paymentAddress
+	} else {
+		p.PaymentAddress = p.Address
+	}
+
```
NicolasMahe: PaymentAddress should be mandatory

x/execution/internal/keeper/keeper.go
NicolasMahe: moving this to todo

process/process.go
antho1404: done

x/execution/internal/keeper/keeper.go
```diff
@@ -300,16 +283,15 @@ func (k *Keeper) Update(ctx sdk.Context, msg types.MsgUpdate) (*executionpb.Exec
...
 	}
 
-	inst, err := k.instanceKeeper.Get(ctx, exec.InstanceHash)
-	if err != nil {
-		return nil, err
-	}
-	srv, err := k.serviceKeeper.Get(ctx, inst.ServiceHash)
-	if err != nil {
-		return nil, err
+	from := msg.Executor
```
NicolasMahe, line 286: This is very weird:
Why the runner should be debited for the execution it just executed?

We don't know who created the execution, so we cannot debit it.
But, in the `create` function, there is this test:
```go
if proc == nil && run.Owner != msg.Signer.String() {
	return nil, sdkerrors.Wrap(sdkerrors.ErrUnauthorized, "signer is not the execution's executor")
}
```
Only the owner of the runner can create an execution is not attached to a process.
Thus, if the execution doesn't have a process, we could make it free. What do you think?

execution/execution.go
```diff
@@ -75,6 +76,6 @@ func (execution *Execution) Fail(err error) error {
 // GetDuration returns the duration of the execution in seconds
 func (execution *Execution) GetDuration() int64 {
 	delta := execution.Stop - execution.Start
-	deltaInSecond := float64(delta) / 1e9
+	deltaInSecond := float64(delta) / float64(time.Second.Nanoseconds())
```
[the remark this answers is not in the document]
antho1404: For information on the proto I tried:

###  cast type:
`(gogoproto.casttype) = "time"` doesn't work because proto/gogo doesn't integrate default packages.

### google proto 
https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp 
This proto doesn't implement the equal function that we need 

Also, the time is hard to set from convert from int64, the only way is to extract the number of seconds and the left of nanoseconds...

x/execution/internal/keeper/keeper.go
antho1404: Let's see that on another PR and totally remove the ability to create a standalone execution

PR #1844, opened 2020-05-26
Add a check in LCD client if the account exists or not

NicolasMahe:
add a check in LCD client if the account exists or not

### DIFF
--- a/cosmos/lcd.go
+++ b/cosmos/lcd.go
@@ -187,12 +187,11 @@ func (lcd *LCD) getAccount() (*auth.BaseAccount, error) {
 	if accR.Address.Empty() {
 		return nil, fmt.Errorf("account %q doesn't exist", lcd.acc.GetAddress().String())
 	}
-	localSeq := lcd.acc.GetSequence()
-	lcd.acc = accR
 	// replace seq if sup
-	if localSeq > lcd.acc.GetSequence() {
-		lcd.acc.SetSequence(localSeq)
+	if lcd.acc.GetSequence() > accR.GetSequence() {
+		accR.SetSequence(lcd.acc.GetSequence())
 	}
+	lcd.acc = accR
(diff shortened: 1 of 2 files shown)


cosmos/lcd.go
```diff
@@ -180,10 +180,15 @@ func (lcd *LCD) getAccount() (*auth.BaseAccount, error) {
...
-	localSeq := lcd.acc.GetSequence()
-	if err := lcd.Get("auth/accounts/"+lcd.acc.GetAddress().String(), &lcd.acc); err != nil {
+	var accR *auth.BaseAccount
+	if err := lcd.Get("auth/accounts/"+lcd.acc.GetAddress().String(), &accR); err != nil {
 		return nil, err
 	}
+	if accR.Address.Empty() {
+		return nil, fmt.Errorf("account %q doesn't exist", lcd.acc.GetAddress().String())
+	}
+	localSeq := lcd.acc.GetSequence()
```
antho1404, line 190: Why move this one? the goal was to have the local sequence and compare it to the sequence on the chain.
Here the local sequence becomes totally useless and so the condition 2 lines bellow `localSeq > lcd.acc.GetSequence()`.
Either this needs to be put back before the fetch or the whole local sequence need to be deleted

NicolasMahe: Look carefully, the get store the fetched account in variable `accR`.
If the sequence system was broken, the e2e test that executes 10 executions in the same block will not work.

antho1404: Hum yes you're right I missed the part that it was replaced... The code is a bit weird. Maybe something like that would be less confusing.
```go

	// replace seq if sup
	if localSeq > accR.GetSequence() {
		accR.SetSequence(localSeq)
	}
	lcd.acc = accR
```

PR #1848, opened 2020-05-27
Implement import and export functions in modules

NicolasMahe:
Closes https://github.com/mesg-foundation/engine/issues/1847

This PR implements the import and export functions in modules.

To export, use the command:
```
mesg-daemon export
```
> Make sure the first line returned by the command doesn't contain "WARNING: State is not initialized. Returning genesis file.". If it's the case, try to start the node and stop it `mesg-daemon start`. I think I have this issue because I'm running the node in docker and the cli on mac. So when I start the node on the Mac, it may resolve the issue.

To import, copy the output of previous command in the genesis.json file of a new node that doesn't have any database, start the node, and see the data are present.

### DIFF
--- a/x/execution/internal/types/genesis.go
+++ b/x/execution/internal/types/genesis.go
@@ -1,23 +1,38 @@
 package types
 
+import (
+	fmt "fmt"
+
+	executionpb "github.com/mesg-foundation/engine/execution"
+	"github.com/mesg-foundation/engine/ext/xvalidator"
+)
+
 // GenesisState - all instance state that must be provided at genesis
 type GenesisState struct {
-	Params Params `json:"params" yaml:"params"`
+	Params     Params                   `json:"params" yaml:"params" validate:"dive"`
+	Executions []*executionpb.Execution `json:"executions" yaml:"executions" validate:"dive"`
 }
 
 // NewGenesisState creates a new GenesisState object
-func NewGenesisState(params Params) GenesisState {
-	return GenesisState{Params: params}
+func NewGenesisState(params Params, execs []*executionpb.Execution) GenesisState {
+	return GenesisState{
+		Params:     params,
+		Executions: execs,
+	}
 }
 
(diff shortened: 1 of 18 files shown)


antho1404:
I think the export/import should stay as close as possible to the database we use. Here we are losing the key on all export. We usually use the hash but we might change to the address or anything. We shouldn't loose this information. I would recommend to switch to a map instead of array or an array of key/value

NicolasMahe:
> I think the export/import should stay as close as possible to the database we use. Here we are losing the key on all export. We usually use the hash but we might change to the address or anything. We shouldn't loose this information. I would recommend to switch to a map instead of array or an array of key/value

As you say, the keys are the hash. If the data didn't contain keys, then a map can be implemented, but it's not the case.
Each module can have its own export and import function, so there is no need to implement the same way everywhere.

x/execution/internal/types/genesis.go
```diff
@@ -1,23 +1,38 @@
...
+		Executions: execs,
+	}
 }
 
 // DefaultGenesisState - default GenesisState used by Cosmos Hub
 func DefaultGenesisState() GenesisState {
 	return GenesisState{
-		Params: DefaultParams(),
+		Params:     DefaultParams(),
+		Executions: nil,
```
antho1404, line 28: Shouldn't it be an empty array to make sure we iterate over something. Otherwise `InitGenesis` will import with a nil value. Same comment for all the other modules

NicolasMahe: It will not cause any issue within Go.
But it could cause some issue in other language or in JSON.
I will update to an empty slice to be sure it doesn't cause any issue in other languages.
Source: https://www.pixelstech.net/article/1539870875-Empty-slice-vs-nil-slice-in-GoLang

NicolasMahe:
Now the genesis contains empty array instead of null.
Before:
```
{"instances":null}
```
After:
```
{"instances":[]}
```

PR #1855, opened 2020-05-27
Add modules gov, upgrade, evidence and crisis

NicolasMahe:
Closes https://github.com/mesg-foundation/engine/issues/1854

Add modules gov, upgrade, evidence and crisis


app/app.go
```diff
@@ -230,6 +251,48 @@ func NewInitApp(
 		app.subspaces[slashing.ModuleName],
 	)
 
+	app.crisisKeeper = crisis.NewKeeper(
```
antho1404, line 254: This is useless if there is no route associated. Can be added later anyway, when/if we need to use the module.

app/app.go
```diff
@@ -179,6 +197,9 @@ func NewInitApp(
 	app.subspaces[staking.ModuleName] = app.paramsKeeper.Subspace(staking.DefaultParamspace)
 	app.subspaces[distr.ModuleName] = app.paramsKeeper.Subspace(distr.DefaultParamspace)
 	app.subspaces[slashing.ModuleName] = app.paramsKeeper.Subspace(slashing.DefaultParamspace)
+	app.subspaces[gov.ModuleName] = app.paramsKeeper.Subspace(gov.DefaultParamspace).WithKeyTable(gov.ParamKeyTable())
```
antho1404, line 200: I would suggest storing the subspace and not the paramspace like that it's consistent with the rest

app/app.go
```diff
@@ -274,13 +339,15 @@ func NewInitApp(
...
+		upgrade.NewAppModule(app.upgradeKeeper),
+		evidence.NewAppModule(app.evidenceKeeper),
 	)
 	// During begin block slashing happens after distr.BeginBlocker so that
 	// there is nothing left over in the validator fee pool, so as to keep the
 	// CanWithdrawInvariant invariant.
 
-	app.mm.SetOrderBeginBlockers(distr.ModuleName, slashing.ModuleName)
-	app.mm.SetOrderEndBlockers(staking.ModuleName)
+	app.mm.SetOrderBeginBlockers(upgrade.ModuleName, distr.ModuleName, slashing.ModuleName)
```
antho1404, line 349: Evidence should be here as it validates the double sign at each begin block.

antho1404: Wait for confirmation on an issue with cosmos team https://github.com/cosmos/scaffold/issues/80

